PR #295, opened 2021-01-14
Feature/alerts

hachiojidev:
Added alert system to redux and a listener to display alerts.

```
 const { alertInfo, alertWarning, AlertDanger, AlertSuccess, push, remove, clear } = useAlerts()

alertInfo(title: string, description?: string)
 ```

### DIFF
--- a/src/state/hooks.ts
+++ b/src/state/hooks.ts
@@ -1,5 +1,6 @@
+import { useEffect, useMemo } from 'react'
 import BigNumber from 'bignumber.js'
-import { useEffect } from 'react'
+import { kebabCase } from 'lodash'
 import { useSelector, useDispatch } from 'react-redux'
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: src/state/hooks.ts [hunk 1]
```diff
@@ -91,3 +95,32 @@ export const usePriceCakeBusd = (): BigNumber => {
   const farm = useFarmFromPid(pid)
   return farm.tokenPriceVsQuote ? bnbPriceUSD.times(farm.tokenPriceVsQuote) : ZERO
 }
+
+// Alerts
+
+export const useAlert = () => {
+  const dispatch = useDispatch()
+  const helpers = useMemo(() => {
```
hachiojidev, line 103: Saw this used like this in the `use-wallet` repo. I think it is a good idea because the alerts will likely be triggered a lot in effects.

PR #301, opened 2021-01-15
perf: Route-based code splitting

RabbitDoge:
Split the JS bundle based on react-router routes


RabbitDoge:
Waiting for the designer input about the loading component

src/App.tsx
```diff
@@ -1,21 +1,25 @@
...
 import NftGlobalNotification from './views/Nft/components/NftGlobalNotification'
 
+// Route-based code splitting
+const Home = lazy(() => import('./views/Home'))
+const Farms = lazy(() => import('./views/Farms'))
+const Farm = lazy(() => import('./views/Farm'))
+const Lottery = lazy(() => import('./views/Lottery'))
+const Pools = lazy(() => import('./views/Pools'))
+const Ifos = lazy(() => import('./views/Ifos'))
+const NotFound = lazy(() => import('./views/NotFound'))
```
hachiojidev, line 19: Probably not needed for this but we should consider keeping critical routes in the bundle.

hachiojidev (changes requested):
Just one issue I see is that the loader.

- It is too big and not responsive
- It should be centered vertically

hachiojidev (approved):

PR #327, opened 2021-01-20
feat(profile): Redux structure

hachiojidev:
Sets up basic redux structure based on the data we have

### DIFF
--- a/src/state/profile/index.tsx
+++ b/src/state/profile/index.tsx
@@ -1,0 +1,30 @@
+/* eslint-disable no-param-reassign */
+import { createSlice, PayloadAction } from '@reduxjs/toolkit'
+import { Profile, ProfileState } from 'state/types'
+
+const initialState: ProfileState = {
+  isInitialized: false,
+  isLoading: true,
+  data: null,
+}
+
+export const profileSlice = createSlice({
+  name: 'profile',
+  initialState,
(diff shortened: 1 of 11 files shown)


src/state/types.ts
```diff
@@ -28,6 +28,25 @@ export interface Pool extends PoolConfig {
...
+  id: number
+  name: string
+  address: string
+  points: number
+  nftAddress: string
+  tokenId: string
+  isActive: boolean
+  team?: Team
+}
+
```
hachiojidev, line 49: This structure will be changing I reckon.

src/state/profile/index.tsx
```diff
@@ -0,0 +1,42 @@
+/* eslint-disable no-param-reassign */
+import { createSlice, PayloadAction } from '@reduxjs/toolkit'
+import { Profile, ProfileState } from 'state/types'
+import getProfile from './getProfile'
+
+const initialState: ProfileState = {
+  isInitialized: false,
+  isLoading: true,
+  profile: null,
```
RabbitDoge, line 9: I think the pattern we used on other states is `data`. Otherwise you'll have this structure `profile.profile`

RabbitDoge (changes requested):

RabbitDoge (approved):

PR #337, opened 2021-01-25
chore: add 6 farms

Chef-Chungus:
added stuff into contracts and types also


Chef-Chungus:
Scheduled to be live @ 17:00 JST

RabbitDoge (approved):

PR #348, opened 2021-01-27
feat(profile): Mint starter

hachiojidev:
- Add minting functionality to profile creation
- Remove deprecated NFT code (partial)


src/hooks/useApproveConfirmTransaction.ts
```diff
@@ -0,0 +1,114 @@
+import { useReducer } from 'react'
```
hachiojidev, line 1: I am not in love with this hook. It needs a lot of work.

RabbitDoge (approved):

PR #349, opened 2021-01-27
feat: Add username selection and profile creation

RabbitDoge:
Username selection and profile validation


src/views/Profile/ProfileCreation/contexts/types.ts
```diff
@@ -0,0 +1,21 @@
+export type Actions =
+  | { type: 'set_step'; step: number }
+  | { type: 'set_team'; teamId: number | null }
+  | { type: 'set_bunny'; bunnyId: number | null }
+  | { type: 'set_username'; userName: string | null }
+  | { type: 'initialize'; step: number }
```
hachiojidev, line 6: We should use the constants you created here.

hachiojidev (approved):

PR #374, opened 2021-02-02
feat (farm/pool): Add HGET Farm and pool

Chef-Chungus:


RabbitDoge (approved):
I couldn't verify the farm and pool contracts, because they are still empty

PR #376, opened 2021-02-03
feat: Add hardhart fork

RabbitDoge:
This is an attempt of using the fork feature of hardhat in order to make a snapshot of the main net and use it as a development blockchain with hardhat.
To make it works, we still need to find a way to use the main net contract addresses when using the hardhart network.

This is still WIP

### DIFF
--- a/hardhat.config.js
+++ b/hardhat.config.js
@@ -1,0 +1,14 @@
+/**
+ * @type import('hardhat/config').HardhatUserConfig
+ */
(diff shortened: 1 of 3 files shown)


matbrgz:
Why this was closed without merge?

RabbitDoge:
> Why this was closed without merge?

Not working, needed more investigation, and we unfortunately don't have the time to spend on this

PR #377, opened 2021-02-03
feat(profile): Save profile data to local storage

hachiojidev:


hachiojidev:
Tested with https://github.com/pancakeswap/pancake-swap-interface/pull/158. I get the CORS issue on localhost but it should work when it is on the same domain.

RabbitDoge (approved):

PR #403, opened 2021-02-05
chore: Use address helper to pull all the contract addresses

RabbitDoge:
This pull request is required to make #376 works.


src/utils/addressHelpers.ts
```diff
@@ -1,34 +1,36 @@
 import addresses from 'config/constants/contracts'
+import { Address } from 'config/constants/types'
 
-const chainId = process.env.REACT_APP_CHAIN_ID
+export const getAddress = (address: Address): string => {
+  const mainNetChainId = 56
```
hachiojidev, line 5: Good idea.

hachiojidev (approved):

PR #418, opened 2021-02-09
Bugfix/translation

hachiojidev:
Adds a 3rd data parameter to the `TranslateString` method so it replaces the Crowdin translations properly.


RabbitDoge (approved):

src/hooks/useI18n.ts
```diff
@@ -1,17 +1,42 @@
...
 
-  return (translationId: number, fallback: string) => {
-    if (translations[0] === 'error') {
-      return fallback
-    }
-    if (translations.length > 0) {
-      return getTranslation(translations, translationId, fallback)
+  return (translationId: number, fallback: string, data: ContextData = {}) => {
+    const foundTranslation = translations.find((translation) => {
+      return translation.data.stringId === translationId
```
karelianpie, line 14: `translation.data` is `undefined` when I run the project locally.

It seems that if you've never selected a language then you'll just get the data to be the array with the error string `['error']`. Once you've selected any language and get back to EN then it seems to be working fine.

PR #427, opened 2021-02-10
chore: Update order to descending chronological order & fix typo

Chef-Chungus:


RabbitDoge (approved):

hachiojidev:
@Chef-Chungus Just a note on this. While it is certainly going to work it doesn't guarantee the order. There is a chance that the code will manipulate the list, changing the order in the future. This was the original intent on using `sortOrder`

Issue #265, opened 2021-01-07
Unable to view lottery tickets from past draws


Chef-Chungus:
![image](https://user-images.githubusercontent.com/71833681/103838841-710d0e00-50d1-11eb-8c95-6f924197e198.png)

When on the past draws section of the lottery and you open a lottery which you previously participated in, your tickets aren't available to view.

PR #432, opened 2021-02-10
fix: Fix 265 - Unable to view lottery tickets from past draws

karelianpie:
Fixes https://github.com/pancakeswap/pancake-frontend/issues/265

The issue was with this assertion `if (ticketIssue.toString() === issueIdex) {}` in `src/utils/lotteryUtils.js`. One is a string and the other a number, so when we use the triple equals for strict equality they are not equal.

I've fixed it with double equals so that it performs type coercion, I'm happy to convert the `issueIdex` to string as well if that's more preferable. I think it looks cleaner using loose equality though, but had to disable a eslint rule.

### DIFF
--- a/src/utils/lotteryUtils.js
+++ b/src/utils/lotteryUtils.js
@@ -62,8 +62,7 @@ export const getTickets = async (lotteryContract, ticketsContract, account, cust
 
   const finalTokenids = []
   ticketIssues.forEach(async (ticketIssue, i) => {
-    // eslint-disable-next-line eqeqeq
-    if (ticketIssue.toString() == issueIdex) {
+    if (parseInt(ticketIssue, 10) === issueIdex) {
       finalTokenids.push(tokenIds[i])
     }
   })
--- a/src/utils/lotteryUtils.js
+++ b/src/utils/lotteryUtils.js
@@ -44,7 +44,7 @@ export const multiBuy = async (lotteryContract, price, numbersList, account) => 
 }
 
 export const getTickets = async (lotteryContract, ticketsContract, account, customLotteryNum) => {
-  const issueIdex = customLotteryNum || (await lotteryContract.methods.issueIndex().call())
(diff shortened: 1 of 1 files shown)


RabbitDoge (changes requested):
Indeed, let's try to avoid disabled eslint rules as much as possible.
It's probably better to perform a conversion. Since both variables are supposed to be numbers, a conversion to number (parseInt, dont forget to add the base as second param), or BigNumber (seems overkill) makes more sense than a conversion to string.
While you are on it, can you fix the typo on `issueIdex` plz ?

This file will need to be migrated to TS at some point.

karelianpie:
> Indeed, let's try to avoid disabled eslint rules as much as possible.
> It's probably better to perform a conversion. Since both variables are supposed to be numbers, a conversion to number (parseInt, dont forget to add the base as second param), or BigNumber (seems overkill) makes more sense than a conversion to string.

Sounds good, used `parseInt()` https://github.com/pancakeswap/pancake-frontend/pull/432/commits/ea1ede1db0b344d32498e1bcf241009f877a6c7d

> While you are on it, can you fix the typo on `issueIdex` plz ?

No worries, fixed with https://github.com/pancakeswap/pancake-frontend/pull/432/commits/5d8a47ef2f87613b39787f79357af1a207edbea2

> This file will need to be migrated to TS at some point.

I could work on that if you think it's important, let me know

hachiojidev (approved):
Nice job. Next time please prepend your branch name with `feature/` so a preview build is made and we can verify the fix.

hachiojidev (approved):

RabbitDoge (approved):

RabbitDoge (changes requested):
Just some conflicts to resolve

karelianpie:
> Just some conflicts to resolve

Merge conflicts resolved @RabbitDoge

RabbitDoge (approved):

PR #436, opened 2021-02-12
Bounty: Easteregg

luisantoniocrag:
This PR adds an easteregg when konamiCode is typed: [up, up, down, down, left, right, left, right, b, a]

At Slack we agreed to call this effect `walletRouting`

This use: PR [180](https://github.com/pancakeswap/pancake-uikit/pull/180) in uikit repository
Example: 
easteregg


src/components/Menu/index.tsx
```diff
@@ -6,17 +6,19 @@ import { LanguageContext } from 'contexts/Localisation/languageContext'
...
   const { selectedLanguage, setSelectedLanguage } = useContext(LanguageContext)
   const { isDark, toggleTheme } = useTheme()
   const cakePriceUsd = usePriceCakeBusd()
   const { profile } = useProfile()
+  useWalletRouting()
 
   return (
     <UikitMenu
-      account={account}
+      account={account}z
```
karelianpie, line 21: Accidental `z`

RabbitDoge (changes requested):
See the slack dev group for the feedbacks

hachiojidev:
@luisantoniocrag Moving forward please follow our commit message guidelines.
https://docs.pancakeswap.finance/code/contributing#committing

Chef-Chungus:
Feedback:

**What 100% need to be changed:**

- Follow the team’s coding rules (linting and prettify)
- Move the konami code hook inside the uikit and make it generic
- Remove the conditional hook call inside the Konami hook

**What may need to be changed:**

- The hack with setTimeout + setInterval is probably not needed
- Maybe using a component that will render the animation for X seconds will be cleaner than modify the dom directly. You can render/not render this component based on what happen in the konami code hook

**Suggestions**
The Konami hook should take a callback as a param, and call this function when the konami code is triggered
You can create a WalletRouting component that will replace the useWalletRouting hook. This component will ruse the konamiCode hook , and render or not the animation by using a state props, and pass the setState as the callback of the konamiCode hook

As a rule of thumb, it’s better to not mutate the dom directly when using React

PR #453, opened 2021-02-14
fix: improve error handling when loading translations

Rovak:
When first starting up the project i got the following error:

Screenshot from 2021-02-14 14-02-47

Change the value when an error is encountered while loading crowdin translations to an empty array to prevent errors in code that doesn't handle the 'error' string value


hachiojidev (approved):
Thanks!

PR #464, opened 2021-02-16
Feature/ifo points

hachiojidev:
Collect IFO points, see achievements


package.json
```diff
@@ -64,6 +64,7 @@
     "@types/jest": "^26.0.19",
     "eslint": "^7.20.0",
     "husky": "4.3.5",
+    "@types/react-redux": "^7.1.16",
```
hachiojidev, line 67: Couldn't figure out why you auto importing `useDispatch` wasn't working - this was the reason.

src/config/constants/campaigns.ts
```diff
@@ -0,0 +1,55 @@
...
+ * badge: Achievement avatar
+ */
+
+const campaigns: Campaign[] = [
+  {
+    id: '511050000',
+    type: 'ifo',
+    title: 'Soteria',
+    badge: 'ifo-wsote.svg',
+  },
```
hachiojidev, line 18: The idea here is keep campaign meta separate because they are eventually going to be many different types.

src/state/achievements/helpers.ts
```diff
@@ -0,0 +1,52 @@
+import { getProfileContract } from 'utils/contractHelpers'
+import { campaignMap } from 'config/constants/campaigns'
+import { Achievement } from 'state/types'
+import { getAchievementTitle, getAchievementDescription } from 'utils/achievements'
+
+/**
+ * Gets all user point increase events on the profile filtered by wallet address
+ */
+export const getUserPointIncreaseEvents = async (account: string) => {
```
hachiojidev, line 9: This function returns all event logs where the user has collected points.

src/state/achievements/helpers.ts
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Gets all user point increase events and adds achievement meta
+ */
+export const getAchievements = async (account: string): Promise<Achievement[]> => {
+  const pointIncreaseEvents = await getUserPointIncreaseEvents(account)
+
+  return pointIncreaseEvents.reduce((accum, event) => {
+    if (!campaignMap.has(event.returnValues.campaignId)) {
+      return accum
+    }
```
hachiojidev, line 35: Look here if we are missing achievements. Likely we did not connect it via the campaign ID properly.

src/utils/achievements.ts
```diff
@@ -0,0 +1,96 @@
...
+        id: 999,
+        fallback: `Committed more than $5 worth of LP in the ${campaign.title} IFO`,
+        data: {
+          name: campaign.title as string,
+        },
+      }
+    default:
+      return campaign.description
+  }
+}
```
hachiojidev, line 48: The idea here is these two functions will handle returning the correct translation for the achievement type.

RabbitDoge (approved):

PR #468, opened 2021-02-16
Bugfix/profile creation mint

hachiojidev:
There was a bug when checking a wallet's minimum CAKE balance. We were passing in `4` instead of the full decimal value (18) of CAKE. This resulted in the message always disappearing for people with the tiniest amount of CAKE in their wallet.

### DIFF
--- a/src/views/Profile/ProfileCreation/Mint.tsx
+++ b/src/views/Profile/ProfileCreation/Mint.tsx
@@ -14,16 +14,18 @@ import useProfileCreation from './contexts/hook'
 
 const starterBunnyIds = [5, 6, 7, 8, 9]
 const nfts = nftList.filter((nft) => starterBunnyIds.includes(nft.bunnyId))
-const minimumCakeBalance = 4
+const cakeCostToMint = 4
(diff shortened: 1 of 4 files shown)


src/views/Profile/ProfileCreation/Mint.tsx
```diff
@@ -14,16 +14,18 @@ import useProfileCreation from './contexts/hook'
 
 const starterBunnyIds = [5, 6, 7, 8, 9]
 const nfts = nftList.filter((nft) => starterBunnyIds.includes(nft.bunnyId))
-const minimumCakeBalance = 4
+const cakeCostToMint = 4
+const minimumCakeBalanceToMint = new BigNumber(cakeCostToMint).multipliedBy(new BigNumber(10).pow(18))
```
hachiojidev, line 18: I plan on checking if we can get these values from the contract so we don't have to explicitly set them here and in Username.

RabbitDoge (approved):

bread-cn:
I want to change the pancakeswap logo image on the homepage, please specify the directory to change thanks

PR #469, opened 2021-02-16
feat: add price to document title

Rovak:
Adds the current CAKE price to the document title

Screenshot from 2021-02-16 18-17-23

### DIFF
--- a/src/DocumentTitle.tsx
+++ b/src/DocumentTitle.tsx
@@ -1,0 +1,21 @@
+import React from 'react'
+import useTitle from 'react-use/lib/useTitle'
+import { usePriceCakeBusd } from './state/hooks'
+
+const DocumentTitle = ({ title, showPrice = true }) => {
+  const documentTitle = []
+
+  // Append price if enabled
+  const cakePriceUsd = usePriceCakeBusd()
+  if (showPrice) {
+    documentTitle.push(`$${cakePriceUsd.toFixed(3)}`)
+  }
+
+  // Append title
+  documentTitle.push(title)
+
+  useTitle(documentTitle.join(' - '))
+  return null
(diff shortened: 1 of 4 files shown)


src/DocumentTitle.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+import useTitle from 'react-use/lib/useTitle'
+import { usePriceCakeBusd } from './state/hooks'
+
+const DocumentTitle = ({ title, showPrice = true }) => {
```
hachiojidev, line 5: Missing types

src/DocumentTitle.tsx
```diff
@@ -0,0 +1,21 @@
...
+  const documentTitle = []
+
+  // Append price if enabled
```
hachiojidev, line 8: Try and use comments only if the code needs supplemental explanation. I think this case it is pretty obvious what is happening.

src/DocumentTitle.tsx
```diff
@@ -0,0 +1,21 @@
...
+  const cakePriceUsd = usePriceCakeBusd()
+  if (showPrice) {
+    documentTitle.push(`$${cakePriceUsd.toFixed(3)}`)
+  }
+
+  // Append title
```
hachiojidev, line 14: Same here

hachiojidev (changes requested):
Thank you for submitting your PR. In this case I don't think bringing on an extra dependency to change the document title is necessary. If you look at the hook's source you'll see it is pretty simple.

https://github.com/streamich/react-use/blob/master/src/useTitle.ts

Chef-Chungus:
Closed in favor of #494

PR #480, opened 2021-02-17
feat: make Burned Balance value user friendly

memoyil:
Burned balance in cake stats shows value with decimals. We should show as same type that total supply does


RabbitDoge:
Sorry @memoyil but your PR doesn't do anything explained in your description. It's just moving some code around, for exactly the same outcome.

Chef-Chungus:
Your PR doesn't achieve what your description describes. Closing.

Issue #463, opened 2021-02-15
Total CAKE Burned shows 0 instead of correct value


memoyil:
**Bug Description**
Total CAKE Burned in the cake stats section in the landing page shows 0 instead of correct value

**Steps to Reproduce**

1. Go to home page
2. Check cake stats

**Expected Behavior**
It should show the amount of cakes burned so far

PR #486, opened 2021-02-18
fix: show the burned balance with or without a wallet

hyeres:
This fixes issue #463

This works by using the hook `useWeb3` instead of `useWallet`. The `useWeb3` hook attempts to use the wallet provider first, then if that does not work, it will use the http provider as a fallback.

Then `web3.currentProvider` passes the correct provider to `getTokenBalance` to find the amount of CAKE tokens in address `0x000000000000000000000000000000000000dEaD`.


capitaldev:
🙏🏻

RabbitDoge (approved):

PR #494, opened 2021-02-20
feat: Add price to document title 2

diegochui:
Attempt 2 to get the price on the document title without extra dependency


hachiojidev (changes requested):
A couple things:

You got the right idea but creating component `DocumentTitle` only to run logic and return null doesn't make much sense. Lets put it into a hook that is called in `App.tsx`

Also, for the future please just create one PR. Please close one.

diegochui:
> A couple things:
> 
> You got the right idea but creating component `DocumentTitle` only to run logic and return null doesn't make much sense. Lets put it into a hook that is called in `App.tsx`
> 
> Also, for the future please just create one PR. Please close one.

Thank you for the feedback, I just made the changes requested. I can't close the other PR as it was not created by me. I saw that it has been a while since the changes were requested on the original PR,  so I decided to make a new one since this is a feature that will be very useful to me personally.

Chef-Chungus:
Thanks @diegochui - we will close the other PR in favour of this one. 

@RabbitDoge will review this shortly

src/App.tsx
```diff
@@ -45,6 +46,10 @@ const App: React.FC = () => {
     }
   }, [account, connect])
 
+  useEffect(() => {
+    document.title = ["PancakeSwap", cakePriceUsd.toFixed(3)].join(' - ')
+  }, [cakePriceUsd])
+
```
hachiojidev, line 52: Perfect. Just 4 more minor changes :)

1. Put the hook into it's own file (in the hooks folder)
2. Please use [toLocaleString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString)
3. Add a `$` sign
4. Use a template literal instead of the join

```
`PancakeSwap - $${cakePriceUsd}`
```

hachiojidev (changes requested):

RabbitDoge (changes requested):
See Hachioji comment

src/App.tsx
diegochui: Thank you for the feedback, I just committed the changes requested. Please let me know if this is what you had in mind.

hachiojidev (changes requested):
Perfect, now just 2 more things :)

1. Please rename the hook to be something more relevant to what the hook does. Don't worry about being too verbose. Better to have a long name that other developers will pick up immediately then a truncated and/or confusing one.

2. Looks like you are not formatting your code with Prettier. Please do.

src/hooks/usePriceUsd.ts
```diff
@@ -0,0 +1,13 @@
+
+import {useEffect} from 'react'
+import {usePriceCakeBusd } from 'state/hooks'
+
+const usePriceUsd = () => {
+    const cakePriceUsd = usePriceCakeBusd()
+    useEffect(() => {
+    document.title = `PancakeSwap - $${Number(cakePriceUsd).toLocaleString(undefined, {minimumFractionDigits: 3, maximumFractionDigits: 3})}`
```
hachiojidev, line 8: This line does not look formatted.

src/hooks/usePriceUsd.ts
```diff
@@ -0,0 +1,13 @@
...
+    }, [cakePriceUsd])
+    return null
```
hachiojidev, line 10: Does eslint throw an error on this? If it _doesn't_ please remove.

diegochui:
> Perfect, now just 2 more things :)
> 
> 1. Please rename the hook to be something more relevant to what the hook does. Don't worry about being too verbose. Better to have a long name that other developers will pick up on immediately then a truncated and/or confusing one.
> 2. Looks like you are not formatting your code with Prettier. Please do.

Thank you

src/hooks/usePriceUsd.ts
diegochui: It doesn't I will remove that

src/hooks/usePriceUsd.ts
diegochui: Noted

hachiojidev:
@diegochui Yes, please update your commit messages and fix conflicts.

diegochui:
> @diegochui Yes, please update your commit messages and fix conflicts.

It got a bit too messy trying to rename the older commits. I just made a new PR with the latest changes requested, we can close this PR.

PR #512, opened 2021-02-23
Feature/web3 react(web3.js)

aliangm:
[x] Before opening a pull request, please read the [contributing guidelines](https://github.com/pancakeswap/pancake-frontend/blob/master/CONTRIBUTING.md) first
[x] If your PR is work in progress, open it as `draft`
[x] Before requesting a review, all the checks need to pass
[x] Explain what your PR does

- Upgrade @binance-chain/bsc-use-wallet to @web3-react package


.env.development
```diff
@@ -2,7 +2,7 @@ REACT_APP_CHAIN_ID = "56"
 REACT_APP_GTAG = "GTM-PXLD3XW"
 
 # 10+ nodes balanced, US/EU
-REACT_APP_NODE_1 = "https://bsc-dataseed1.ninicoin.io"
+REACT_APP_NODE_1 = "https://bsc-dataseed1.binance.io"
```
Chef-Chungus, line 5: Why did you change the node?

package.json
```diff
@@ -31,24 +31,30 @@
...
     "canvas-confetti": "^1.3.3",
     "chart.js": "^2.9.4",
     "date-fns": "^2.17.0",
+    "ethers": "^5.0.31",
     "lodash": "^4.17.20",
     "node-fetch": "^2.6.1",
     "react": "^17.0.1",
     "react-chartjs-2": "^2.11.1",
     "react-countup": "^4.3.3",
+    "react-device-detect": "^1.15.0",
```
hachiojidev, line 57: Why is this needed?

src/connectors/index.ts
```diff
@@ -0,0 +1,37 @@
...
+export const network = new NetworkConnector({
+  urls: { [NETWORK_CHAIN_ID]: NETWORK_URL }
+})
+
+let networkLibrary: Web3Provider | undefined
+export async function getNetworkLibrary(): Promise<Web3Provider> {
+  const provider = await network.getProvider()
+  networkLibrary = networkLibrary ?? new Web3Provider(provider as any)
+  return networkLibrary
+}
```
hachiojidev, line 25: Can you explain a bit why `networkLibrary` is necessary outside the function?

src/components/Web3ReactManager/index.tsx
```diff
@@ -0,0 +1,59 @@
+import React, { useState, useEffect } from 'react'
+import { useWeb3React } from '@web3-react/core'
+
+import useEagerConnect from 'hooks/useEagerConnect'
+import useInactiveListener from 'hooks/useInactiveListener'
+import PageLoader from 'components/PageLoader'
+import { network } from 'connectors'
+
+export default function Web3ReactManager({ children }: { children: JSX.Element }) {
```
hachiojidev, line 9: This does not follow our pattern with typing or functions.

src/components/Web3ReactManager/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const [showLoader, setShowLoader] = useState(false)
+  useEffect(() => {
+    const timeout = setTimeout(() => {
+      setShowLoader(true)
+    }, 600)
+
+    return () => {
+      clearTimeout(timeout)
+    }
+  }, [])
```
hachiojidev, line 35: Why are we delaying the loading state?

src/components/Web3ReactManager/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  // if the account context isn't active, and there's an error on the network context, it's an irrecoverable error
+  if (!active && error) {
+    // return (
+    //   <MessageWrapper>
+    //     <Message>{t('unknownError')}</Message>
+    //   </MessageWrapper>
+    // )
+  }
```
hachiojidev, line 49: I hope we're going to do something here.

src/components/Web3ReactManager/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+    return () => {
+      clearTimeout(timeout)
+    }
+  }, [])
+
+  // on page load, do nothing until we've tried to connect to the injected connector
+  if (!triedEager) {
+    return null
+  }
```
hachiojidev, line 40: I am not sure this is the best approach. The user will see a blank white page.

hachiojidev (commented):
When I click on Binance Smart Wallet and don't have it on my browser the site reverts to a blank page.

src/components/UnlockButton.tsx
```diff
@@ -1,12 +1,26 @@
...
 import useI18n from 'hooks/useI18n'
 
 const UnlockButton = (props) => {
   const TranslateString = useI18n()
-  const { connect, reset } = useWallet()
-  const { onPresentConnectModal } = useWalletModal(connect, reset)
+  const { account, activate, deactivate, connector } = useWeb3React()
+  const handleLogin = (connectorId: ConnectorId) => {
+    if (connectorId === 'walletconnect') {
+      if (connector instanceof WalletConnectConnector && connector.walletConnectProvider?.wc?.uri) {
```
RabbitDoge, line 14: What this if block is for ?

src/App.tsx
```diff
@@ -41,9 +42,9 @@ const App: React.FC = () => {
 
   useEffect(() => {
```
RabbitDoge, line 43: isn't it redundant with `useEagerConnect` ?

src/components/Web3ReactManager/index.tsx
```diff
@@ -0,0 +1,59 @@
+import React, { useState, useEffect } from 'react'
```
RabbitDoge, line 1: I don't know what you guys think (@aliangm @hachiojidev ) but for me , this component and the `useEagerConnect` should be merged together. The logic in this component seems to be a continuation of the `useEagerConnect` hook, and I don't see a use case where we only want to use `useEagerConnect`.

Also, it seems that this component render a loaders while trying to connect, we probably want to render the website even if the user is not connected....

src/connectors/index.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export const injected = new InjectedConnector({
+  supportedChainIds: [56, 97]
+})
+
+// mainnet only
+export const walletconnect = new WalletConnectConnector({
+  rpc: { 1: NETWORK_URL },
```
RabbitDoge, line 33: Wild guess, but isn't `1` the network id ? It's not going to work.

src/connectors/index.ts
```diff
@@ -0,0 +1,37 @@
...
+
+let networkLibrary: Web3Provider | undefined
+export async function getNetworkLibrary(): Promise<Web3Provider> {
+  const provider = await network.getProvider()
+  networkLibrary = networkLibrary ?? new Web3Provider(provider as any)
+  return networkLibrary
+}
+
+export const injected = new InjectedConnector({
+  supportedChainIds: [56, 97]
```
RabbitDoge, line 28: Supported chain ids should be taken from the env files, in case we support different setup with local blockchains later

.env.development
aliangm: I changed it because I'm unable to get response from ninicoin. I will revert it back

package.json
aliangm: Removed

src/connectors/index.ts
aliangm: No need for now. I removed it

src/utils/getLibrary.ts
```diff
@@ -0,0 +1,5 @@
+import { Web3Provider } from '@ethersproject/providers'
```
RabbitDoge, line 1: This function should instantiate a web3 instance according to the doc https://github.com/NoahZinsmeister/web3-react/tree/v6/docs#props , not just returning the provider. 

If we do so, the `useWeb3` hook become obsolete, because we can retrieve the web3 instance from the `useWeb3React` hook

src/components/Web3ReactManager/index.tsx
aliangm: Updated

src/components/Web3ReactManager/index.tsx
aliangm: As it's Web3ReactManager is wrapper, it will be loaded before children. That's why I added delay

src/components/Web3ReactManager/index.tsx
aliangm: I was going to add toast

src/components/Web3ReactManager/index.tsx
aliangm: This case could be removed as we always try with injected provider when users unable to connect wallet

src/components/UnlockButton.tsx
aliangm: It's for walletconnect provider init issue

aliangm: When users sign in again using walletconnect, it produces bug

src/App.tsx
aliangm: Correct

src/connectors/index.ts
aliangm: Fixed

src/connectors/index.ts
aliangm: Fixed

src/utils/getLibrary.ts
aliangm: I updated this part because if we change, we should update the part used `useWeb3` hook

src/components/Web3ReactManager/index.tsx
hachiojidev: What happens without the delay?

src/connectors/NetworkConnector.ts
```diff
@@ -0,0 +1,194 @@
...
+        if ('error' in result) {
+          reject(new RequestError(result?.error?.message, result?.error?.code, result?.error?.data))
+        } else if ('result' in result) {
+          resolve(result.result)
+        } else {
+          reject(new RequestError(`Received unexpected JSON-RPC response to ${method} request.`, -32000, result))
+        }
+      }
+    }
+  }
```
hachiojidev, line 112: Is there a reason for this to be outside the `try` block? If not I think we should put it inside.

src/hooks/useInactiveListener.ts
```diff
@@ -0,0 +1,40 @@
+import { useWeb3React as useWeb3ReactCore } from '@web3-react/core'
+import { useEffect } from 'react'
+import { injected } from '../connectors'
+
+export default function useInactiveListener(suppress = false) {
+  const { active, error, activate } = useWeb3ReactCore() // specifically using useWeb3React because of what this hook does
+
+  useEffect(() => {
+    const { ethereum } = window
```
hachiojidev, line 9: Is working on the global `ethereum` object absolutely necessary? We want to avoid globals if we can.

src/react-app-env.d.ts
```diff
@@ -1 +1,9 @@
 /// <reference types="react-scripts" />
+interface Window {
+  ethereum?: {
+    isMetaMask?: true
+    on?: (...args: any[]) => void
+    removeListener?: (...args: any[]) => void
+    autoRefreshOnNetworkChange?: boolean
+  }
+}
```
hachiojidev, line 9: I am not sure but does this extend the `window` interface or overwrite it? It is the latter then we should make sure this extends.

src/connectors/NetworkConnector.ts
self-coding-crab: I used NetworkConnector from uniswap-interface coz they made web3.js compatible connector.
It seems we can put inside

src/hooks/useInactiveListener.ts
self-coding-crab: It's necessary for only inactive listener

src/react-app-env.d.ts
self-coding-crab: It does extend the interface I think but let me confirm

RabbitDoge:
#528

PR #519, opened 2021-02-25
Feature/ifo fixes

hachiojidev:
- Refactor IFO including new designs
- Fix state not updated when contributing


src/hooks/useApprove.ts
```diff
@@ -67,14 +67,8 @@ export const useLotteryApprove = () => {
...
-    try {
-      const tx = await tokenContract.methods
-        .approve(spenderAddress, ethers.constants.MaxUint256)
-        .send({ from: account })
-      return tx
-    } catch {
-      return false
-    }
+    const tx = await tokenContract.methods.approve(spenderAddress, ethers.constants.MaxUint256).send({ from: account })
+    return tx
```
hachiojidev, line 71: Removed try/catch so the error would not be swallowed

src/hooks/useAllowance.ts
```diff
@@ -29,17 +29,17 @@ export const useLotteryAllowance = () => {
...
+  const [allowance, setAllowance] = useState(new BigNumber(0))
 
   useEffect(() => {
     const fetch = async () => {
       try {
         const res = await tokenContract.methods.allowance(account, spenderAddress).call()
         setAllowance(new BigNumber(res))
       } catch (e) {
-        setAllowance(null)
+        console.error(e)
```
RabbitDoge, line 42: Don't you want to display a ui message here ?

src/hooks/useApprove.ts
RabbitDoge: The problem was the lack of ui feedback, imo we should keep the try catch and display a notification saying the approval has failed

hachiojidev: This solves that exact problem. It was swallowing the error not allowing the UI to catch it.

hachiojidev: Perhaps we should re-throw the error instead.

RabbitDoge (approved):

PR #528, opened 2021-02-26
feat: Replace use-wallet by web3-react

### LINKED ISSUE #317: Add a message when using the wrong network
Display an error message when a user tries to connect from a different network than BSC.

### LINKED ISSUE #372: QA: Web3 React integration
Regarding: https://github.com/pancakeswap/pancake-frontend/pull/346

## MetaMask
**FARMS**
- [ ] Modal doesn't close when funds are deposited despite transaction being confirmed, stays in "Pending confirmation" state.
- [ ] Modal doesn't close when funds are withdrawn despite transaction being confirmed, stays in "Pending confirmation" state.
- [x] Harvest OK
- [x] Deposit OK
- [x] Withdraw OK
- [x] Approve OK

image

**POOLS**
- [ ] Your stake - NaN
- [ ] Cake earned - NaN
- [ ] NaN cake Available when unstaking CAKE
- [ ] Modal doesn't close when funds are unstaked despite transaction being confirmed, stays in "Pending confirmation" state.
- [ ] Modal doesn't close when funds are staked despite transaction being confirmed, stays in "Pending confirmation" state.
- [x] Staking OK
- [x] Unstaking OK
- [ ] Can't harvest due to NaN Cake Earned
image
image

RabbitDoge:
Web3-react migration
Fix #317 
Fix #372

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     ]
   },
   "dependencies": {
-    "@binance-chain/bsc-use-wallet": "^0.8.1",
+    "@binance-chain/bsc-connector": "^1.0.0",
     "@crowdin/crowdin-api-client": "^1.10.4",
     "@ethersproject/abi": "^5.0.7",
     "@pancakeswap-libs/uikit": "^0.11.0",
@@ -40,10 +40,14 @@
     "@types/react-dom": "^17.0.1",
     "@types/react-router-dom": "^5.1.5",
     "@types/styled-components": "^5.1.7",
+    "@web3-react/core": "^6.1.9",
+    "@web3-react/injected-connector": "^6.0.7",
+    "@web3-react/walletconnect-connector": "^6.1.9",
     "bignumber.js": "^9.0.0",
     "canvas-confetti": "^1.3.3",
     "chart.js": "^2.9.4",
     "date-fns": "^2.17.0",
+    "ethers": "^5.0.31",
     "js-cookie": "^2.2.1",
     "lodash": "^4.17.20",
     "node-fetch": "^2.6.1",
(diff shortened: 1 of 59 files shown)


package.json
```diff
@@ -40,10 +40,14 @@
...
     "@types/react-router-dom": "^5.1.5",
     "@types/styled-components": "^5.1.7",
+    "@web3-react/core": "^6.1.9",
+    "@web3-react/injected-connector": "^6.0.7",
+    "@web3-react/walletconnect-connector": "^6.1.9",
     "bignumber.js": "^9.0.0",
     "canvas-confetti": "^1.3.3",
     "chart.js": "^2.9.4",
     "date-fns": "^2.17.0",
+    "ethers": "^5.0.31",
```
RabbitDoge, line 50: The package is already used here https://github.com/pancakeswap/pancake-frontend/blob/develop/src/hooks/useApprove.ts#L4 but wasn't in the deps list.
We should remove it but it's probably out of the scope of this pr

src/hooks/useEagerConnect.ts
```diff
@@ -0,0 +1,20 @@
...
+import { connectorsByName } from 'utils/web3React'
+
+const { injected } = connectorsByName
+
+const useEagerConnect = () => {
+  const { activate } = useWeb3React()
+
+  useEffect(() => {
+    injected.isAuthorized().then((isAuthorized) => {
+      const hasSignedIn = window.localStorage.getItem('accountStatus')
```
hachiojidev, line 12: We should consider using cookies if we plan on keeping the exchange on a different domain and we want to share the login logic.

hachiojidev (approved):

PR #534, opened 2021-02-26
Feature/web3 rework

RabbitDoge:
Following the work done on #528 

Each contract instance can be retrieve with a helper function, taking an optional web3 instance as a param.
If the web3 instance is provided, the contract will have access to the default account (if the user is connected)

Hooks inside `useContract` are helpers to use the getContracts function with an instance of web3.


src/utils/contractHelpers.ts
```diff
@@ -1,29 +1,83 @@
...
+import bep20Abi from 'config/abi/erc20.json'
+import cakeAbi from 'config/abi/cake.json'
+import ifoAbi from 'config/abi/ifo.json'
+import pointCenterIfo from 'config/abi/pointCenterIfo.json'
+import lotteryAbi from 'config/abi/lottery.json'
+import lotteryTicketAbi from 'config/abi/lotteryNft.json'
+import masterChef from 'config/abi/masterchef.json'
+import sousChef from 'config/abi/sousChef.json'
+import sousChefBnb from 'config/abi/sousChefBnb.json'
+import claimRefundAbi from 'config/abi/claimRefund.json'
```
hachiojidev, line 36: Just a thought but as the app grows maybe we want to import these dynamically.

src/utils/web3.ts
```diff
@@ -1,22 +1,14 @@
...
- */
-const getWeb3 = () => {
-  const web3 = new Web3(httpProvider)
-  return web3
-}
-const getContract = (abi: any, address: string, contractOptions?: ContractOptions) => {
-  const web3 = getWeb3()
-  return new web3.eth.Contract((abi as unknown) as AbiItem, address, contractOptions)
+const getWeb3NoAccount = () => {
+  return web3NoAccount
```
hachiojidev, line 10: Is using the same instance of `Web3` intentional?

hachiojidev (approved):

src/utils/web3.ts
RabbitDoge: yes, the function is just there for compatibility. this is to avoid re creating the web3 instance each time (which is the reason why we get spammed by the deprecation warnings). When the user is not connected, this instance doesn't need to change.

PR #551, opened 2021-03-02
feat: Integrate profile subgraph

hachiojidev:


src/state/achievements/helpers.ts
```diff
@@ -1,25 +1,38 @@
...
+            points {
+              id
+              campaignId
+              points
+            }
+          }
+        }
+      `,
+    )
+    return data.user.points
```
hachiojidev, line 33: Deliberately kept it simple so we can decide later how we want to structure graphql integration.

RabbitDoge (approved):

PR #574, opened 2021-03-04
Fixes on Farm table

cbunnydev:
Used Image component from uikit to show coins
Updated small titles to fit the figma design
Removed forwardRef from Table component as it was not used
Updated table row layout when its the middle-sized viewport

### DIFF
--- a/src/views/Farms/components/FarmTable/Row.tsx
+++ b/src/views/Farms/components/FarmTable/Row.tsx
@@ -102,7 +102,7 @@ const Row: React.FunctionComponent<RowProps> = (props) => {
                 return (
                   <td key={key}>
                     <CellInner>
-                      <CellLayout label={TranslateString(999, 'APR')}>
+                      <CellLayout label={TranslateString(736, 'APR')}>
                         <Apr {...props.apr} hideButton={isMobile} />
                       </CellLayout>
                     </CellInner>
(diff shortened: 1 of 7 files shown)


src/views/Farms/components/FarmTable/Row.tsx
```diff
@@ -109,7 +102,7 @@ const Row: React.FunctionComponent<RowProps> = (props) => {
                 return (
                   <td key={key}>
                     <CellInner>
-                      <CellLayout label={TranslateString(999, 'Apr')}>
+                      <CellLayout label={TranslateString(999, 'APR')}>
```
hachiojidev, line 105: This has a translation.
```
{TranslateString(736, 'APR')}
```

hachiojidev: I also see the APR string in `ActionPanel`. That should be translated as well.

src/views/Farms/components/FarmTable/Row.tsx
```diff
@@ -146,7 +141,7 @@ const Row: React.FunctionComponent<RowProps> = (props) => {
               </CellLayout>
             </EarnedMobileCell>
             <AprMobileCell>
-              <CellLayout label={TranslateString(999, 'Apr')}>
+              <CellLayout label={TranslateString(999, 'APR')}>
```
hachiojidev, line 144: Same here

src/views/Farms/components/types.ts
```diff
@@ -29,89 +28,79 @@ export type TableDataTypes = {
...
     translationId: 999,
     sortable: true,
+    label: 'Earned',
   },
   {
     id: 3,
-    bold: '',
-    normal: 'APR',
     name: 'apr',
     translationId: 999,
```
hachiojidev, line 49: Same here...

hachiojidev (changes requested):
Please check the project to see if there is an existing translation.

Issue #584, opened 2021-03-05
[Bounty] Add link to add token to Metamask in Pool card details


Chef-Chungus:
Reward: $150 USD in CAKE

**Description:**
Add "Add {Token name} to MetaMask" to the details section of the Pool card.
We may need to add the earn token contract address to `src/config/constants/pools.ts` in order to make this work. 

**Important notes:**
MUST BE THE EARNING TOKEN, NOT THE STAKING TOKEN.

**Resources:**
Metamask Docs: https://docs.metamask.io/guide/registering-your-token.html#code-free-example
![image](https://user-images.githubusercontent.com/71833681/110076863-85674180-7dc8-11eb-9c56-2ad324c984d1.png)

rickmort:
hey! I'm working on this issue =)

Tomcat88:
Working on it!

Questions:

- According to the MetaMask doc an url must be provided for the token icon, it is optional (a placeholder will be showed), is it ok if i leave it optional in the PoolConfig type?

Tomcat88:
Ah it seems that @rickmort was faster than me :)

josemtm:
im working on it but it will some changes in the pool config interface like you said chef

Chef-Chungus:
approved - @rickmort's PR #593. Bounty to go to RickMort

PR #593, opened 2021-03-05
Feature/Add link to add token to Metamask in Pool card details

rickmort:
closes #584

This PR adds a link functionality to register a token to Metamask from Pool Card.

This uses https://github.com/pancakeswap/pancake-uikit/pull/212

METAMASK-CAKE

PR #595, opened 2021-03-05
feat: Add metamask link and add needed token props

Tomcat88:
This PR adds the MetaMask link requested in issue #584 

In order to do so the following changes has been made:

- Add MetaMask fox svg icon download directly from their site ()
- Add needed properties on type PoolConfig
- Add needed value only for the CAKE pool
- Add metamask utils file to invoke MetaMask token watch
- Add new component in CardFooter showing a link only if needed values are present (and metamask is found)
- Changed PoolCard in order to pass needed props to CardFooter

If I'm missing something please let me know.


RabbitDoge:
It seems to be redundant with #593

Chef-Chungus:
@Tomcat88 - Thanks for sending this through, we will close this in favor of #593 as that was submitted first

PR #600, opened 2021-03-07
fix: APR loading on farms page

cbunnydev:


src/views/Farms/Farms.tsx
```diff
@@ -243,9 +243,7 @@ const Farms: React.FC = () => {
 
     const row: RowProps = {
       apr: {
-        value: farm.apy
-          ? Number(`${farm.apy.times(new BigNumber(100)).toNumber().toLocaleString('en-US').slice(0, -1)}`)
-          : null,
+        value: farm.apy && farm.apy.times(new BigNumber(100)).toNumber().toLocaleString('en-US').slice(0, -1),
```
hachiojidev, line 246: What is going on here? What format are you trying to achieve?

cbunnydev: I used the same APR formatting from the FarmCard and it seems that it was wrong.

I think he needed to use maximumFractionDigits option.

Just updated it in FarmCard component

hachiojidev (approved):

PR #604, opened 2021-03-08
chore: Move block to redux state

RabbitDoge:
Move `block` from context to redux state


src/state/block/index.ts
```diff
@@ -0,0 +1,17 @@
+import { createSlice } from '@reduxjs/toolkit'
+import { Block } from '../types'
+
+const initialState: Block = { blockNumber: 0 }
```
hachiojidev, line 4: Just had a thought. I wonder if the UI would see improvement if we saved the last block number in local storage.

hachiojidev (approved):

PR #651, opened 2021-03-12
fix: Sort by APR not sorting high APR properly

### LINKED ISSUE #650: Sort by APR not sorting properly
**Bug Description**
Sort by APR in the Farms table does not work properly if farm has APR > 1000%

**Steps to Reproduce**

1. Go to Farms
2. Choose `APR` from "SORT BY"

**Expected Behavior**
Expect to see farms with 1000 or more APR on top of the table above farms with 3-digit APRs

**Additional Context**
On screenshot you can see that `BUX-BNB` farm has 1186% APR and is below other farms that have less APR.
Screenshot 2021-03-12 at 11 43 17

ArtemKolichenkov:
In Farms table sort by APR does not sort farms with more than 1000 APR properly.
Currently in master it gets shoved below farms with 200% APR due to the way how lodash's orderBy works with BigNumber.js numbers.
You can see contrived example in sandbox here - https://codesandbox.io/s/c1kgz
This PR fixes ordering so farms with >1000% APR come on top.

Fixes #650 

**Current master (note BUX-BNB)**
Screenshot 2021-03-12 at 11 43 17

**This PR**
Screenshot 2021-03-12 at 11 43 40

### DIFF
--- a/src/views/Farms/Farms.tsx
+++ b/src/views/Farms/Farms.tsx
@@ -142,7 +142,7 @@ const Farms: React.FC = () => {
   const sortFarms = (farms: FarmWithStakedValue[]): FarmWithStakedValue[] => {
     switch (sortOption) {
       case 'apr':
-        return orderBy(farms, (farm: FarmWithStakedValue) => Number(farm.apy), 'desc')
+        return orderBy(farms, (farm: FarmWithStakedValue) => farm.apy.toNumber(), 'desc')
       case 'multiplier':
         return orderBy(farms, (farm: FarmWithStakedValue) => Number(farm.multiplier.slice(0, -1)), 'desc')
       case 'earned':
--- a/src/views/Farms/Farms.tsx
+++ b/src/views/Farms/Farms.tsx
@@ -142,7 +142,7 @@ const Farms: React.FC = () => {
   const sortFarms = (farms: FarmWithStakedValue[]): FarmWithStakedValue[] => {
     switch (sortOption) {
       case 'apr':
-        return orderBy(farms, 'apy', 'desc')
+        return orderBy(farms, (farm: FarmWithStakedValue) => Number(farm.apy), 'desc')
       case 'multiplier':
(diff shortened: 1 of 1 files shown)


memoyil:
Sometimes apr stucks at Loading... for a while. Can you check casting that string value to number will cause a error?

ArtemKolichenkov:
> Sometimes apr stucks at Loading... for a while. Can you check casting that string value to number will cause a error?

The "apr stuck at Loading" issue was fixed in b906ef589b6240589958d07b0f76f528aa0b6b76 and I believe it was in submitted directly to master originally and recently synced to develop. Are you sure you're not somehow running it on top of outdated develop branch? 🤔  Please check `Farms.tsx` on line 246 if it still has old casting to Number. Not sure how it can be unsynced since I rebased on latest develop before making PR, but idk... 

If above is not the case then I'm not sure why you see Loading, can't reproduce it 🤔

hachiojidev:
@ArtemKolichenkov Thanks for submitting this fix. Currently @cbunnydev is actively developing the table component. @cbunnydev can you review this PR and determine if it is appropriate for the work you're doing?

src/views/Farms/Farms.tsx
```diff
@@ -142,7 +142,7 @@ const Farms: React.FC = () => {
   const sortFarms = (farms: FarmWithStakedValue[]): FarmWithStakedValue[] => {
     switch (sortOption) {
       case 'apr':
-        return orderBy(farms, 'apy', 'desc')
+        return orderBy(farms, (farm: FarmWithStakedValue) => Number(farm.apy), 'desc')
```
hachiojidev, line 145: I think it would better to ensure that `apy` is the correct type to begin with. @cbunnydev what do you think?

RabbitDoge: farm.apy.toNumber() instead of Number(farm.apy)

RabbitDoge (changes requested):

RabbitDoge (approved):

PR #653, opened 2021-03-12
fix: Prevent EasterEgg bunnies reposition on rerenders

z3s:
Currently FallingBunnies use random position for bunnies. so RefreshContextProvider make the EasterEgg component rerender then it changes bunnies position in the middle of animation and it's not nice.


Chef-Chungus:
Thanks for submitting this! It's been bothering me, but it wasn't a priority for us to fix ahah. We'll review this next week. @z3s

hachiojidev (approved):

PR #657, opened 2021-03-13
feat: Change past lottery draws input to dropdown

memoyil:
image

### DIFF
--- a/src/views/Lottery/components/Select/RoundSelect.tsx
+++ b/src/views/Lottery/components/Select/RoundSelect.tsx
@@ -1,0 +1,149 @@
+import React, { useState, useRef, useEffect } from 'react'
+import styled, { css } from 'styled-components'
+import { ArrowDropDownIcon, Text } from '@pancakeswap-libs/uikit'
+
+const DropDownHeader = styled.div`
+  width: 100%;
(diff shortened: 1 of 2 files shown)


RabbitDoge:
@memoyil  there are some conflicts

src/views/Lottery/components/Select/RoundSelect.tsx
```diff
@@ -0,0 +1,149 @@
...
+  options: RoundOptionProps[]
+  onChange?: (option: RoundOptionProps) => void
+}
+
+export interface RoundOptionProps {
+  label: string
+  value: any
+}
+
+const RoundSelect: React.FunctionComponent<RoundSelectProps> = ({ options, onChange }) => {
```
ArtemKolichenkov, line 103: I think it is worth to do a bit of refactoring and move `views/Farms/components/Select` under common `src/components` instead of copying the whole thing again and again to every view that needs it. The only differences between Farm's `Select` and this `RoundSelect` are:
- Allowing `overflow-y`, which could be just some optional prop to the `Select`, for example `allowOverflow={true}`.
- Farm's `Select` filters out selected option. Similarly could just be ` excludeSelected={true}`.

PR #669, opened 2021-03-16
fix: Fix earned text color to disabled (which is still active for the farms that you earn token) after wallet disconnect

### LINKED ISSUE #668: Farm earned textcolor after wallet disconnect is in active color
**Bug Description**
Farm earned textcolor after wallet disconnect is in active color

**Steps to Reproduce**

1. Go to farms
2. Connect your wallet
3. Check your earnings
3. Disconnect from your wallet
4. The farms that you earned look like this with active '?' color

Screenshot 2021-03-16 at 09 58 29

**Expected Behavior**
Should be in disabled text color

memoyil:
Fixes #668

### DIFF
--- a/src/views/Farms/components/FarmTable/Earned.tsx
+++ b/src/views/Farms/components/FarmTable/Earned.tsx
@@ -15,9 +15,13 @@ const Amount = styled.span<{ earned: number }>`
 
 const Earned: React.FunctionComponent<EarnedProps> = ({ earnings }) => {
   const { account } = useWeb3React()
-  const displayBalance = earnings !== null && account ? earnings.toLocaleString() : '?'
+  let amountEarned = earnings
+  if (account == null) {
+    amountEarned = null
+  }
 
-  return <Amount earned={earnings}>{displayBalance}</Amount>
+  const displayBalance = amountEarned !== null ? amountEarned.toLocaleString() : '?'
+  return <Amount earned={amountEarned}>{displayBalance}</Amount>
 }
 
 export default Earned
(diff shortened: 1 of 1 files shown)


src/views/Farms/components/FarmTable/Earned.tsx
```diff
@@ -15,9 +15,13 @@ const Amount = styled.span<{ earned: number }>`
 
 const Earned: React.FunctionComponent<EarnedProps> = ({ earnings }) => {
   const { account } = useWeb3React()
-  const displayBalance = earnings !== null && account ? earnings.toLocaleString() : '?'
+  let amountEarned = earnings
+  if (account == null) {
```
hachiojidev, line 19: `!account` is preferable

hachiojidev (changes requested):

src/views/Farms/components/FarmTable/Earned.tsx
RabbitDoge: I'm surprised eslint doesn't throw an error for `==`

memoyil: done

RabbitDoge (changes requested):
the commit description doesn't match the pr content

src/views/Farms/components/FarmTable/Earned.tsx
RabbitDoge: You can use a ternary with a const. As a rule of thumb, avoiding let is always a good idea.

memoyil: Fixed

memoyil:
@RabbitDoge can you explain me more? What is not matching?

memoyil (commented):
Change to ternary condition

hachiojidev (approved):

RabbitDoge (approved):

PR #675, opened 2021-03-17
refactor: Rework pools config

RabbitDoge:
Use a token list


src/config/constants/types.ts
```diff
@@ -47,6 +47,13 @@ export interface Address {
   56: string
 }
 
+export interface Token {
+  symbol: string
+  address?: Address
+  decimals?: number
```
Chef-Chungus, line 53: decimals shouldn't be optional, right?

Chef-Chungus (approved):

src/config/constants/types.ts
RabbitDoge: When we use bnb, it doesn't make sense to have a decimal

PR #682, opened 2021-03-18
fix: Black icon and texts in dark theme

### LINKED ISSUE #681: [Bug] some icons and font on some devices is black
**Bug Description**
While using trust wallet, there are some icons and font which are displaying as black on my device.

**Steps to Reproduce**
Unsure if reproducible
Visit https://pancakeswap.finance/pools on trust wallet ios 

**Expected Behavior**
The font and icon colour should match what displays on the desktop

**Additional Context**

memoyil:
Fixes #681


src/views/Pools/components/CardFooter.tsx
```diff
@@ -105,12 +106,12 @@ const CardFooter: React.FC<Props> = ({
...
 
+  const { isDark } = useTheme()
+
   return (
     <StyledFooter isFinished={isFinished}>
       <Row>
-        <FlexFull>
-          <Tag />
-        </FlexFull>
+        <FlexFull>{Tag({ outline: !isDark })}</FlexFull>
```
hachiojidev, line 114: Why change to a function?

hachiojidev (changes requested):

src/views/Pools/components/CardFooter.tsx
memoyil: FIxed

memoyil (commented):
Fixed in next commit

memoyil:
@hachiojidev ready for you to review again

src/components/Balance.tsx
```diff
@@ -16,7 +16,15 @@ interface BalanceProps extends TextProps {
 }
 
 const StyledText = styled(Text)<TextProps>`
-  color: ${({ isDisabled, color, theme }) => (isDisabled ? theme.colors.textDisabled : color)};
+  color: ${({ isDisabled, color, theme }) => {
+    if (isDisabled) {
+      return theme.colors.textDisabled
+    }
+    if (!color) {
```
RabbitDoge, line 23: 🤔 
```
   if (!color) {
      return color
    }
```

RabbitDoge: this change doesn't seem useful because the color already default to `text`

src/components/Tags.tsx
```diff
@@ -1,28 +1,41 @@
...
-  <Tag variant="secondary" outline startIcon={<VerifiedIcon />} {...props}>
-    Core
-  </Tag>
-)
+const CoreTag = (props) => {
+  const { isDark } = useTheme()
+  return (
+    <Tag variant="secondary" outline={!isDark} startIcon={<VerifiedIcon />} {...props}>
+      Core
+    </Tag>
```
hachiojidev, line 10: I think these are suppose to remain solid. If this is just to fix the icon color you can apply the color to the icon directly.

```
<Tag variant="secondary" outline startIcon={<VerifiedIcon color="secondary" />} {...props}>
```

memoyil: Done

src/components/Balance.tsx
memoyil: You are right, don't remember why i put it there :D. I changed now, i am checking against default color value if there is no value coming from component props (which in this case) I fallback to theme text color .

memoyil:
@hachiojidev @RabbitDoge You can review it again

src/components/Balance.tsx
RabbitDoge: For me, this change does exactly the same thing than the code before, but with more code. I'm not sure I understand what you're trying to do

memoyil: Issue is when you switch to dark theme, it uses the same color value which is ‘text’ instead of theme one  unless it is passed explicitly. Thats why I added a theme.color.text. It is reproducible in iOS.

RabbitDoge: `text` and `theme.color.text` are the same

src/components/Balance.tsx
```diff
@@ -16,7 +16,15 @@ interface BalanceProps extends TextProps {
...
-  color: ${({ isDisabled, color, theme }) => (isDisabled ? theme.colors.textDisabled : color)};
+  color: ${({ isDisabled, color, theme }) => {
+    if (isDisabled) {
+      return theme.colors.textDisabled
+    }
+    if (color !== Balance.defaultProps.color) {
+      return color
+    }
+    return theme.colors.text
+  }};
```
dobpp, line 27: color: ${({ isDisabled, color, theme }) => isDisabled ? theme.colors.textDisabled : color !== Balance.defaultProps.color ? color : theme.color.text}};

memoyil: I think nested ternary conditions is not readable simply

src/components/Balance.tsx
memoyil: Good news, I think i found the issue, when we use styledtext with css attributes, it overrides color attribute of actual dom object and interpreted as color: 'text' or color: 'textdisabled' instead of actual html color values. Browsers failed and fallback to other color and in some browsers (metamask, safari) it is different than we expect. Thats why I removed styledText and put condition in Text props.

src/components/Balance.tsx
RabbitDoge: Nested ternaries raise an error with our eslint config

RabbitDoge (approved):

hachiojidev (approved):

PR #684, opened 2021-03-18
refactor: Farm config

RabbitDoge:
Use the token list in the farm config


src/config/constants/tokens.ts
```diff
@@ -597,6 +597,222 @@ const tokens = {
     decimals: 18,
     projectLink: 'https://dusk.network/',
   },
+  busd: {
+    symbol: 'BUSD',
```
hachiojidev, line 601: I think we should use the `QuoteToken` values here because we check against frequently.
```
farm.quoteToken.symbol === QuoteToken.BNB
```

Chef-Chungus (approved):

hachiojidev (changes requested):

src/config/constants/tokens.ts
RabbitDoge: I wanted to remove this `QuoteToken` struct, because it doesn't make too much sens to have it, basically any token can be a quote token. Furthermore, we check against a lot in the farm for the apy, but like the pools, I'll change the apy algo to be more generic, so less check against the token name

hachiojidev (approved):

PR #686, opened 2021-03-18
fix: Past pools with BNB token

RabbitDoge:
BNB doesn't have an address , which break some pools after the token list rework


ChefHutch (approved):
Looks good 👍 

Should we wrap this in a check? https://github.com/pancakeswap/pancake-frontend/blob/cc5e002b57b37464919099584238d9996dd7aa28/src/views/Pools/components/CardFooter.tsx#L148

```
            { tokenAddress && <TokenLink ... >
              ...
            </TokenLink> }
```

I don't have the domain knowledge yet to really comment on the impact of passing `' '` to `useERC20`

RabbitDoge:
Good catch, I added the check. useERC20 without an address will just create a contract with no address set, but in that case the contrat won't be used so it doesn't matter.

Chef-Chungus (approved):

PR #689, opened 2021-03-18
fix: Farm card and row liquidity and details translations

memoyil:


src/views/Farms/components/FarmCard/CardActionsContainer.tsx
```diff
@@ -68,15 +68,17 @@ const CardActions: React.FC<FarmCardActionsProps> = ({ farm, account, addLiquidi
     )
   }
 
+  const tokenEarnedString = TranslateString(330, 'CAKE Earned', { asset: 'CAKE' }).split(' ')
```
RabbitDoge, line 71: You don't need to use a translation here , furthermore, splitting like doesn't guarantee to return an acceptable result. What takes 2 words in English, can take 1 or 3 in other langs.

After the tokens reworks, you can get the token symbol with `farm.token` and `farm.quoteToken`. Have a look at the `/state/types` file

RabbitDoge (changes requested):

src/views/Farms/components/FarmCard/CardActionsContainer.tsx
memoyil: 👍🏻 sure

memoyil: We don't have earned token info in farm object therefore I reverted that code stub

memoyil (commented):
@RabbitDoge Can you review it again

src/components/ExpandableSectionButton/ExpandableSectionButton.tsx
```diff
@@ -19,10 +20,12 @@ const Wrapper = styled.div`
 `
 
 const ExpandableSectionButton: React.FC<ExpandableSectionButtonProps> = ({ onClick, expanded }) => {
+  const TranslateString = useI18n()
+
   return (
     <Wrapper aria-label="Hide or show expandable content" role="button" onClick={() => onClick()}>
       <Text color="primary" bold>
-        {expanded ? 'Hide' : 'Details'}
+        {expanded ? 'Hide' : TranslateString(658, 'Details')}
```
hachiojidev, line 28: Please wrap `Hide` in the translate function as well. If we don't have an ID already just use 999

src/views/Farms/components/FarmTable/Actions/HarvestAction.tsx
```diff
@@ -42,11 +42,13 @@ const HarvestAction: React.FunctionComponent<FarmWithStakedValue> = ({ pid, user
...
 
+  const tokenEarnedString = TranslateString(330, 'CAKE Earned', { asset: 'CAKE' }).split(' ')
+
   return (
     <ActionContainer>
       <ActionTitles>
-        <Title>CAKE </Title>
-        <Subtle>EARNED</Subtle>
+        <Title>{tokenEarnedString[0]} </Title>
+        <Subtle>{tokenEarnedString[1]}</Subtle>
```
hachiojidev, line 51: In this case it is okay to leave `CAKE` as is and just wrap `Earned` in the translate string
```
{TranslateString(999, 'Earned')}
```

hachiojidev (changes requested):

src/views/Farms/components/FarmTable/Actions/HarvestAction.tsx
```diff
@@ -42,11 +42,13 @@ const HarvestAction: React.FunctionComponent<FarmWithStakedValue> = ({ pid, user
     updateValue.current(earningsBusd)
   }, [earningsBusd, updateValue])
 
+  const tokenEarnedString = TranslateString(330, 'CAKE Earned', { asset: 'CAKE' }).split(' ')
```
RabbitDoge, line 45: cake doesn't need to be translated, and this split can lead to unexpected behaviour. Please remove and just translate `EARNED`

RabbitDoge (changes requested):

src/views/Farms/components/FarmTable/Actions/HarvestAction.tsx
memoyil: Done

src/views/Farms/components/FarmTable/Actions/HarvestAction.tsx
memoyil: Done

src/components/ExpandableSectionButton/ExpandableSectionButton.tsx
memoyil: Done

hachiojidev (approved):

RabbitDoge (approved):

PR #693, opened 2021-03-19
refactor: Farms APY

RabbitDoge:
Make farm apys generic

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -5,10 +5,9 @@ BigNumber.config({
   DECIMAL_PLACES: 80,
 })
 
-export const CAKE_PER_BLOCK = new BigNumber(40)
-export const BLOCKS_PER_YEAR = new BigNumber(10512000)
 export const BSC_BLOCK_TIME = 3
-export const CAKE_POOL_PID = 1
+export const CAKE_PER_BLOCK = new BigNumber(40)
+export const BLOCKS_PER_YEAR = new BigNumber((60 / BSC_BLOCK_TIME) * 60 * 24 * 365) // 10512000
 export const BASE_URL = 'https://pancakeswap.finance'
 export const BASE_EXCHANGE_URL = 'https://exchange.pancakeswap.finance'
 export const BASE_ADD_LIQUIDITY_URL = `${BASE_EXCHANGE_URL}/#/add`
--- a/src/views/Farms/Farms.tsx
+++ b/src/views/Farms/Farms.tsx
@@ -5,16 +5,15 @@ import BigNumber from 'bignumber.js'
 import { useWeb3React } from '@web3-react/core'
 import { Image, Heading, RowType, Toggle, Text } from '@pancakeswap-libs/uikit'
 import styled from 'styled-components'
-import { BLOCKS_PER_YEAR, CAKE_PER_BLOCK, CAKE_POOL_PID } from 'config'
 import FlexLayout from 'components/layout/Flex'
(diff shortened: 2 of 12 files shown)


Chef-Chungus:
approved from my side (functional test)

Chef-Chungus (approved):
Approved (Functional test)

src/config/index.ts
```diff
@@ -5,10 +5,9 @@ BigNumber.config({
   DECIMAL_PLACES: 80,
 })
 
-export const CAKE_PER_BLOCK = new BigNumber(40)
-export const BLOCKS_PER_YEAR = new BigNumber(10512000)
 export const BSC_BLOCK_TIME = 3
-export const CAKE_POOL_PID = 1
+export const CAKE_PER_BLOCK = new BigNumber(40)
```
hachiojidev, line 9: Didn't you mention this is actually 25?

src/views/Farms/Farms.tsx
```diff
@@ -159,58 +157,18 @@ const Farms: React.FC = () => {
...
-        if (!farm.lpTotalInQuoteToken) {
-          liquidity = null
-        }
-        if (farm.quoteToken.symbol === tokens.wbnb.symbol) {
-          liquidity = bnbPrice.times(farm.lpTotalInQuoteToken)
-        }
-        if (farm.quoteToken.symbol === tokens.cake.symbol) {
-          liquidity = cakePrice.times(farm.lpTotalInQuoteToken)
-        }
+        const quoteTokenPriceUsd = prices[farm.quoteToken.symbol.toLowerCase()]
```
hachiojidev, line 167: Should we add a check in case the API does not have the farm symbol? Or a fallback number.

src/views/Home/components/EarnAPYCard.tsx
```diff
@@ -24,55 +24,24 @@ const CardMidContent = styled(Heading).attrs({ size: 'xl' })`
...
-            farm.tokenPriceVsQuote &&
-            new BigNumber(farm.tokenPriceVsQuote)
-              .times(farm.dual.rewardPerBlock)
-              .times(BLOCKS_PER_YEAR)
-              .div(farm.lpTotalInQuoteToken)
-
-          apy = cakeApy && dualApy && cakeApy.plus(dualApy)
+  const highestApy = useMemo(() => {
+    const apys = farmsLP
+      .filter((farm) => farm.pid !== 0 && farm.multiplier !== '0X')
```
hachiojidev, line 32: Is this necessary really if we're getting the highest apy?

src/config/index.ts
RabbitDoge: Still 40, 15 go to a burn pool

src/views/Farms/Farms.tsx
RabbitDoge: In case we don't have a number, `quoteTokenPriceUsd` will be null, then the apy will be `NaN` and this case is handled by the function used to get the apy, it will just return null. The FE will keep showing the loader.

RabbitDoge: BTW, my next step in this rework is to remove the api, and compute the price from the farms data, so we won't have any missing data

src/views/Home/components/EarnAPYCard.tsx
RabbitDoge: probably not

hachiojidev (approved):

PR #713, opened 2021-03-22
Update config objects accessing tokens.blk & tokens.btc 

ChefHutch:
- Following some recent changes to the `btc` & `blk` token config, we were getting compilation errors as our farms & pools configs were still accessing the old values
- This updates them to the new values - `btcb` & `blink`


Chef-Chungus:
resolved in #710

PR #714, opened 2021-03-22
Make profile avatar responsive to sizes set in parent container

ChefHutch:
**Code changes**
- Currently `ProfileAvatar` has hard coded width/height values for its images.
- This is problematic as other components (i.e. in the competition page) make use of the `ProfileAvatar` but need it to be different sizes.
- This PR makes `ProfileAvatar`'s images: `AvatarInactive` & `AvatarWrapper` & `TeamAvatar` make use of width/height %s (set by parent divs) rather than hard coding those values.
- Added wrapping divs in each place that the component is used with explicit width/height values

**Testing**
- Side note - @hachiojidev  I think this is a reasonable argument for the devil's work AKA snapshot tests! They'd show any instances where the `<ProfileAvatar>` styles have changed but I've missed wrapping it in a div setting an explicit height/width.

[ ] Explain what your PR does


src/config/constants/farms.ts
```diff
@@ -649,7 +649,7 @@ const farms: FarmConfig[] = [
       97: '',
       56: '0xC743Dc05F03D25E1aF8eC5F8228f4BD25513c8d0',
     },
-    token: tokens.blk,
+    token: tokens.blink,
```
ChefHutch, line 652: Branched from https://github.com/pancakeswap/pancake-frontend/pull/713 to fix the compile errors. This diff will disappear once that's merged

src/views/Profile/components/ProfileAvatar.tsx
```diff
@@ -11,16 +11,14 @@ const TeamAvatar = styled.img`
   border: 1px solid ${({ theme }) => theme.card.background};
   border-radius: 50%;
   bottom: 0px;
-  height: 24px;
   position: absolute;
   right: 0px;
-  width: 24px;
+  width: 37.5%;
+  height: 37.5%;
```
hachiojidev, line 17: Why this value? Will this scale properly?

ChefHutch: Yea so I just checked the existing ratios between the smaller team image and the larger profile image, which are 

below mediaQueries.sm: `24px > 64px` - 37.5%
above mediaQueries.sm: `48px > 128px` - 37.5%

I'll also put a min-width of 20px on the team image?

hachiojidev (approved):

PR #720, opened 2021-03-23
chore: Set default sort order to 999 if none is provided

Chef-Chungus:


src/views/Collectibles/components/NftList.tsx
```diff
@@ -59,7 +59,7 @@ const NftList = () => {
 
   return (
     <NftGrid>
-      {orderBy(nfts, 'sortOrder').map((nft) => {
+      {orderBy(nfts, 'sortOrder' || 999).map((nft) => {
```
hachiojidev, line 62: 1. Have you confirmed this works? I didn't know you could pass in a key and/or a primitive value.
2. If we are going to go this route lets put the sort order in `config/index.ts` as `DEFAULT_SORT_ORDER`

hachiojidev (changes requested):

src/views/Collectibles/components/NftList.tsx
Chef-Chungus: I havent cause I'm a pleb

PR #747, opened 2021-03-25
fix: Align farm details between card and tableview

memoyil:
Things done :

1. Make get token lp and view contract names are the same for card and tableview

Screenshot 2021-03-29 at 11 46 39
Screenshot 2021-03-29 at 11 47 30

2. Not show get token lp link when farm is finished for card and tableview

image
Screenshot 2021-03-29 at 11 39 47

3. Add total liquidity even when farm is finished (it can be visible when you select tableview so i thought we should put it here too)

4. Add pairinfo link for card

Screenshot 2021-03-29 at 11 39 47


RabbitDoge (commented):
Add a screen of the changes in order to get them validated by Bugs plz

memoyil:
@RabbitDoge done

hachiojidev (approved):

PR #753, opened 2021-03-27
fix: Apr check when number is 0 and simplify apr component

memoyil:
Fixed When apr is 0 (when farm is finished) it fails in apr value check and returns itself which ends up in the runtime row apr value type is number instead of string which might cause problem in the future

Fixed Apr component uses originalValue when 0 instead of value as well.

### DIFF
--- a/src/views/Farms/Farms.tsx
+++ b/src/views/Farms/Farms.tsx
@@ -208,7 +208,7 @@ const Farms: React.FC = () => {
 
     const row: RowProps = {
       apr: {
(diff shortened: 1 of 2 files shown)


RabbitDoge:
A check is already done on the APY value, I don't think this change is relevant.

https://github.com/pancakeswap/pancake-frontend/blob/develop/src/utils/apy.ts#L34

memoyil:
@RabbitDoge It is related to finished farms 

https://github.com/pancakeswap/pancake-frontend/blob/a0031cc8d51078ada32e00d88163ca5548648539/src/views/Farms/Farms.tsx#L174

PR #754, opened 2021-03-28
fix: Expand margin and padding shorthands and remove duplicate padding

memoyil:


RabbitDoge (changes requested):
3-values shorthands just make the code harder to read and to understand

memoyil:
@RabbitDoge You are right, I expanded shorthands

hachiojidev:
@memoyil @RabbitDoge I don't think using shorthand is a problem if your intention is to set the values. For example if you need a margin left and margin right you should write it like this. (too many times I see it written `0 8px` when only the left and right should be set).

```
margin-left: 8px;
margin-right: 8px;
```

If your intention is to explicitly set the top and bottom margin to 0 it should be written like this (in my opinion)

```
margin: 0 8px;
```

I don't think this PR is necessary but. Will leave it in your hands @RabbitDoge

RabbitDoge:
We won't merge this because it's more a change related to your personal preferences rather than a real improvement

PR #756, opened 2021-03-29
chore: Update # of cake per block

Chef-Chungus:
Deploy @ 21:00 GMT+9

### DIFF
--- a/src/views/Home/components/CakeStats.tsx
+++ b/src/views/Home/components/CakeStats.tsx
@@ -42,7 +42,7 @@ const CakeStats = () => {
         </Row>
         <Row>
           <Text fontSize="14px">{TranslateString(540, 'New CAKE/block')}</Text>
-          <CardValue fontSize="14px" decimals={0} value={25} />
+          <CardValue fontSize="14px" decimals={0} value={22} />
         </Row>
       </CardBody>
     </StyledCakeStats>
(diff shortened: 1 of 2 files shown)


src/views/Home/components/CakeStats.tsx
```diff
@@ -42,7 +42,7 @@ const CakeStats = () => {
         </Row>
         <Row>
           <Text fontSize="14px">{TranslateString(540, 'New CAKE/block')}</Text>
-          <CardValue fontSize="14px" decimals={0} value={25} />
+          <CardValue fontSize="14px" decimals={0} value={22} />
```
hachiojidev, line 45: I feel we at least need a comment in here why it is 22, and not 25, or even 40.

hachiojidev (changes requested):

src/views/Home/components/CakeStats.tsx
```diff
@@ -26,6 +26,12 @@ const CakeStats = () => {
   const burnedBalance = getBalanceNumber(useBurnedBalance(getCakeAddress()))
   const cakeSupply = totalSupply ? getBalanceNumber(totalSupply) - burnedBalance : 0
 
+  // CAKE/Block details
+  // 40 CAKE per block is minted per block
```
hachiojidev, line 30: A little redundant.
> 40 CAKE per block is minted

hachiojidev (changes requested):
>Although that is probably best put somewhere higher up since it is so useful. Can we put it in config/index.ts above CAKE_PER_BLOCK

hachiojidev (approved):

PR #757, opened 2021-03-29
chore(api): TVL endpoint

ChefKai:
Use a tailored endpoint for frontend to return only used values such as to display the Total Value Locked, and avoid loading all pairs state.


hachiojidev (approved):

pnghai:
> Use a tailored endpoint for frontend to return only used values such as to display the Total Value Locked, and avoid loading all pairs state.

Hi, I guess the api call 'tvl' came from repository locating at https://github.com/pancakeswap/go-stat, yet I can't found the tvl api in its source code. Which api server source are you using?

PR #762, opened 2021-03-30
feat: Amends on farm table

cbunnydev:
Fixed issues below.

Use the loading component for anything that is loading, not a ?
When everything is loaded, and there is no earned amount, we should show 0, not ?
If we can automatically expand  farms that you have a staked amount in, it would be better for the user experience.
There should be a transition animation when opening the details. (see attached gif)


memoyil:
@cbunnydev Can you review this too ? https://github.com/pancakeswap/pancake-frontend/pull/747

RabbitDoge:
Why there are duplicated commits on this pr?

PR #763, opened 2021-03-30
fix: Translation is not visible when language is selected before when refresh or enter the webpage

memoyil:
Race condition in use effects creates multiple promises which overrides the selected translation and show english sometimes when enter the website or refresh it.

image


RabbitDoge (changes requested):
I see the problem you are describing, but I don't think you picked the best solution.
Imo, we should either set the lang at null first (this is why there is this check in the second use effect) or either group the 2 use effect in order to be sure the fetch is execute after the lang has been set.

@hachiojidev has worked a lot on the translation system, I'm curious to have his opinion

memoyil:
@RabbitDoge I changed to null. There is one more issue which I think this fix can solve that; for the last 2-3 days I have been getting '429 Too Many Requests' error for English translations. Since this fix will reduce the requests if the user selected other languages, we can also benefit from that.

RabbitDoge (approved):

PR #769, opened 2021-04-01
feat: Add usePrevious hook

hachiojidev:
https://reactjs.org/docs/hooks-faq.html#how-to-get-the-previous-props-or-state


RabbitDoge (approved):
usePreviousValue would be more appropriate. With the name, we don't know if it's related to the router.

PR #773, opened 2021-04-01
feat(timer): Timer component

ChefHutch:
This is a generic timer component that will render a D, H, M countdown and a tooltip.

UI Example with tooltip while hovered
Screenshot 2021-04-01 at 15 15 12

UI example without tooltip
Screenshot 2021-04-01 at 15 15 32

UI example without using custom text components
Screenshot 2021-04-01 at 15 16 45

Code usage:
```js
  <Timer
    timerStage={'string that preceeds the timer'}
    minutes={'minutes value as number'}
    hours={'hours value as number'}
    days={'days value as number'}
    blockNumber={'number used by the tooltip to show the specific block being counted to'}
    showTooltip={'boolean - render or don't render tooltip'}
    HeadingTextComponent={'component - allows you to pass a custom header text comp, with sensible defaults'}
    BodyTextComponent={'component - allows you to pass a custom body text comp, with sensible defaults'}
  />
```

Code usage example:

```js
const TimerHeadingComponent = ({ children }) => (
  <StyledHeading background="linear-gradient(180deg, #FFD800 0%, #EB8C00 100%)" fill>
    {children}
  </StyledHeading>
)

const TimerBodyComponent = ({ children }) => (
  <Text color="#ffff" fontWeight="600" fontSize="16px" mr="16px">
    {children}
  </Text>
)

const App = ({minutes, hours, days, targetBlockNumber}) => {
...
...
...
  return (
    <Timer
      timerStage={
        competitionState.state === LIVE ? `${TranslateString(410, 'End')}:` : `${TranslateString(1212, 'Start')}:`
      }
      minutes={minutes}
      hours={hours}
      days={days}
      blockNumber={targetBlockNumber}
      showTooltip
      HeadingTextComponent={TimerHeadingComponent}
      BodyTextComponent={TimerBodyComponent}
    />
  )
}
```

### DIFF
--- a/src/components/Timer/index.tsx
+++ b/src/components/Timer/index.tsx
@@ -51,7 +51,7 @@ const Wrapper: React.FC<TimerProps> = ({
   hours,
   days,
   blockNumber,
-  showTooltip = true,
+  showTooltip = false,
   HeadingTextComponent = DefaultHeadingTextComponent,
   BodyTextComponent = DefaultBodyTextComponent,
 }) => {
--- a/src/components/Timer/index.tsx
+++ b/src/components/Timer/index.tsx
@@ -10,8 +10,8 @@ export interface TimerProps {
   days?: number
   showTooltip?: boolean
   blockNumber?: number
-  HeadingTextComponent?: any
-  BodyTextComponent?: any
+  HeadingTextComponent?: React.ElementType
+  BodyTextComponent?: React.ElementType
 }
 
 const StyledTimerFlex = styled(Flex)<{ showTooltip?: boolean }>`
@@ -50,8 +50,8 @@ const Wrapper: React.FC<TimerProps> = ({
   minutes,
   hours,
   days,
-  showTooltip,
   blockNumber,
+  showTooltip = true,
   HeadingTextComponent = DefaultHeadingTextComponent,
   BodyTextComponent = DefaultBodyTextComponent,
 }) => {
--- a/src/components/Timer/index.tsx
+++ b/src/components/Timer/index.tsx
@@ -1,0 +1,86 @@
+import React from 'react'
+import styled from 'styled-components'
+import { Flex, Heading, Text } from '@pancakeswap-libs/uikit'
+import TimerTooltip from './TimerTooltip'
+
+export interface TimerProps {
+  timerStage?: string
+  minutes?: number
+  hours?: number
+  days?: number
+  showTooltip?: boolean
+  blockNumber?: number
+  HeadingTextComponent?: any
+  BodyTextComponent?: any
+}
+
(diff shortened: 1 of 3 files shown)


src/views/Farms/components/Tooltip/Tooltip.tsx
```diff
@@ -44,11 +45,11 @@ const Container = styled.div`
   }
 `
 
-const Tooltip: React.FunctionComponent<TooltipProps> = ({ content, children }) => {
+const Tooltip: React.FunctionComponent<TooltipProps> = ({ content, children, className }) => {
   return (
     <Container>
       {children}
-      <TooltipContent>{content}</TooltipContent>
+      <TooltipContent className={className}>{content}</TooltipContent>
```
ChefHutch, line 52: Required to allow tooltip to be restyled here https://github.com/pancakeswap/pancake-frontend/pull/773/files#diff-7139049a58c241ef6622e4a8f989946a026906be596a4b442f86a423d493a436R6.

By default it sits above the component, hence the restyling. 

I think work is already being done on this tooltip - hence not going further and making accept a 'position' prop and dynamically style it that way or similar

src/components/Timer/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+export interface TimerProps {
+  timerStage?: string
+  minutes?: number
+  hours?: number
+  days?: number
+  showTooltip?: boolean
+  blockNumber?: number
+  HeadingTextComponent?: any
+  BodyTextComponent?: any
```
hachiojidev, line 14: `ReactElement` is what I think you need here...

src/components/Timer/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+    {children}
+  </Text>
+)
+
+const Wrapper: React.FC<TimerProps> = ({
+  timerStage,
+  minutes,
+  hours,
+  days,
+  showTooltip,
```
hachiojidev, line 53: This should have a default value.

hachiojidev (changes requested):

RabbitDoge (commented):
The tooltip system is something we will reuse and should be in the uikit. So far there is only one direction, but maybe we can make a good use of popperjs later. Anyway, if you can start by making this one generic and add it to the uikit it would be a good start.

src/components/Timer/index.tsx
ChefHutch: Dead right. I left this in by accident - was experimenting with ways of passing a component as props it.

ChefHutch:
Made those changes @hachiojidev  

As discussed with @RabbitDoge - we'll look at moving this into the uikit after the trading competition (priority) is live.

hachiojidev (approved):

PR #782, opened 2021-04-03
Chart object change to be zoomable and cooler

YannSuissa:
Next step would be to add date time, and recover the cake price based on the date to display the real amounts of the pots in another serie.
To do it need to change the API request to add the timestamps and query for cake price history somewhere

some captures : 
Capture1
Capture2

Best, 

[ ] Explain what your PR does


memoyil:
Adding a lot of 3rd party libraries might affect page loading time and performance. Can we make sacrifices to make it light ? (may be we can sacrifice some functions)

YannSuissa:
> Adding a lot of 3rd party libraries might affect page loading time and performance. Can we make sacrifices to make it light ? (may be we can sacrifice some functions)

I Removed the react-chartjs-2 for react-apexcharts
The more the stats are sexy and shows potential prizes the more players will join it, imo..

memoyil:
> > Adding a lot of 3rd party libraries might affect page loading time and performance. Can we make sacrifices to make it light ? (may be we can sacrifice some functions)
> 
> I Removed the react-chartjs-2 for react-apexcharts
> The more the stats are sexy and shows potential prizes the more players will join it, imo..

I absolutely agree with you, but keep the site light and fast for all mobile devices and browser should also be our objective

hachiojidev (changes requested):
Thank you for submitting this. I have not seen Apex Charts before and the package looks really nice. There are a few issues that need to be addressed before we consider this PR.

1. Please follow our commit message rules

2. Please create a description branch with the word `feature/` in it so a preview build will be created. This will allow use to verify it's performance and let our designer look at it.

3. There are quite a few ESLint warning, please address them.

4. @memoyil raises a good point.

> I absolutely agree with you, but keep the site light and fast for all mobile devices and browsers should also be our objective.

Checking bundle phobia we see that [Apex Charts](https://bundlephobia.com/result?p=apexcharts@3.26.0) is 119kb while our current solution [Chart.js](https://bundlephobia.com/result?p=chart.js@3.0.1) is approximately half. This is not a deal breaker but something to be aware of.

5. The current way you build the data object is not optimal. From what I can tell there is no reason to use state or an effect to achieve the data set. Please have another look at it. Also, if I am mistaken and there is a reason please let me know.

RabbitDoge:
apexchart will add more than 1sec of loading time on mobile devices , just for a more fancy chart on a non essential feature. For me it's a deal breaker. Chartjs is already too heavy, if any change, it should be for something more light. The lottery page is currently a single bundle, so it will impact the whole page performaces.

hachiojidev:
@YannSuissa Give my concerns and @RabbitDoge's comments we will close this PR for now.

PR #783, opened 2021-04-05
Feature/easter nft giveaway

hachiojidev:
To be released 15:00 JST


src/views/Collectibles/components/ClaimNftModal.tsx
```diff
@@ -22,26 +24,23 @@ const Actions = styled.div`
...
   const { toastError, toastSuccess } = useToast()
-  const bunnySpecialContract = useBunnySpecialContract()
 
   const handleConfirm = async () => {
-    bunnySpecialContract.methods
-      .mintNFT(nft.bunnyId)
-      .send({ from: account })
-      .on('sending', () => {
+    onClaim()
+      .once('sending', () => {
```
hachiojidev, line 35: Turns out that `on` does not have the "sending" event.

RabbitDoge (approved):

PR #785, opened 2021-04-05
feat: Custom page meta

hachiojidev:
Adds ability to set a custom title, description and image per page.

#### Usage

Use the url (including slash) as the key. For example, to add custom info for farms you would add an entry like so:

```
{
  '/farms': {
    title: 'custom title',
    description: 'custom description',
    image: 'custom absolute URL'
  }
}
```

It also supports customizing 1 or all properties (will use the default if not specified).


package.json
```diff
@@ -59,6 +59,7 @@
     "react-countup": "^4.3.3",
     "react-dom": "^17.0.1",
     "react-feather": "^2.0.8",
+    "react-helmet-async": "^1.0.9",
```
hachiojidev, line 62: React Helmet is not maintained anymore.

src/components/layout/PageMeta.tsx
```diff
@@ -0,0 +1,30 @@
...
+  const pageMeta = customMeta[pathname] || {}
+  const { title, description, image } = { ...DEFAULT_META, ...pageMeta }
+  const titleWithPrice = [title, cakePriceUsdDisplay].join(' - ')
+
+  return (
+    <Helmet>
+      <title>{titleWithPrice}</title>
+      <meta property="og:title" content={title} />
+      <meta property="og:description" content={description} />
+      <meta property="og:image" content={image} />
```
Chef-Chungus, line 25: `<meta name=“twitter:card” content=“summary_large_image”>` 
Need to add this here, i think

RabbitDoge:
Suggestion : Merge `Page` and `PageMeta` files into `Page` since `PageMeta` is never supposed to be used alone

src/components/layout/PageMeta.tsx
Chef-Chungus, line 25: `<meta name=“twitter:card” content=“summary_large_image”>` 
Need to add this here, i think

hachiojidev: Per our conversation - this is already in `index.html`

Chef-Chungus (approved):

PR #795, opened 2021-04-05
fix: Page title shows "false" when loading

### LINKED ISSUE #794: Page title shows "false" when price is loading
Screenshot 2021-04-05 at 14 04 34

memoyil:
Fixes #794

### DIFF
--- a/src/components/layout/Page.tsx
+++ b/src/components/layout/Page.tsx
@@ -25,19 +25,19 @@ const StyledPage = styled(Container)`
 const PageMeta = () => {
   const { pathname } = useLocation()
   const cakePriceUsd = usePriceCakeBusd()
-  const cakePriceUsdDisplay =
-    cakePriceUsd.gt(0) &&
-    `$${cakePriceUsd.toNumber().toLocaleString(undefined, {
-      minimumFractionDigits: 3,
-      maximumFractionDigits: 3,
-    })}`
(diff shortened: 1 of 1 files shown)


src/components/layout/Page.tsx
```diff
@@ -25,19 +25,19 @@ const StyledPage = styled(Container)`
...
+  const cakePriceUsdDisplay = cakePriceUsd.eq(0)
+    ? ''
+    : `$${cakePriceUsd.toNumber().toLocaleString(undefined, {
+        minimumFractionDigits: 3,
+        maximumFractionDigits: 3,
+      })}`
   const pageMeta = customMeta[pathname] || {}
   const { title, description, image } = { ...DEFAULT_META, ...pageMeta }
-  const titleWithPrice = [title, cakePriceUsdDisplay].join(' - ')
+  const pageTitle = cakePriceUsdDisplay ? [title, cakePriceUsdDisplay].join(' - ') : title
```
Chef-Cheems, line 36: Was going to comment that we could just keep the original `cakePriceUsdDisplay` (with `gt(0) && ...`) and change only this line (36). But then figured out that with this change `cakePriceUsdDisplay` would be nice `string` type instead of `bool | string` 👍

Chef-Cheems (approved):
LGTM

hachiojidev (approved):

PR #797, opened 2021-04-05
fix: Increase translation limit, add and fix translations for hide and details

memoyil:


src/components/FoldableText/index.tsx
```diff
@@ -35,7 +37,7 @@ const FoldableText: React.FC<FoldableTextProps> = ({ title, children, ...props }
         </Text>
         <StyledExpandableLabelWrapper>
           <ExpandableLabel expanded={isExpanded} onClick={() => setIsExpanded(!isExpanded)}>
-            {isExpanded ? 'Hide' : 'Details'}
+            {isExpanded ? TranslateString(999, 'Hide') : TranslateString(658, 'Details')}
```
hachiojidev, line 40: Looks like Hide is `1066`

src/views/Pools/components/CardFooter.tsx
```diff
@@ -112,7 +112,7 @@ const CardFooter: React.FC<Props> = ({
           <Tag />
         </FlexFull>
         <StyledDetailsButton onClick={handleClick}>
-          {isOpen ? TranslateString(1066, 'Hide') : TranslateString(658, 'Details')} <Icon />
+          {isOpen ? TranslateString(999, 'Hide') : TranslateString(658, 'Details')} <Icon />
```
hachiojidev, line 115: Again, I see this as `1066` in Crowdin.

hachiojidev (changes requested):

src/components/FoldableText/index.tsx
memoyil: I checked the response that returns from crowdin, cannot see 1066 and when i put there it is not translated either. Could it be because our request has limit? 'https://api.crowdin.com/api/v2/projects/422458/languages/de/translations?fileId=8&limit=200'

memoyil:
@hachiojidev I found the issue, we have 309 translations total but our request limits it to 200, thats why we lose hide translation. I increased the limit and changed back to translation id correct

RabbitDoge (approved):

hachiojidev (approved):

PR #812, opened 2021-04-07
chore: Standardize confirming states

Chef-Chungus:


src/views/Collectibles/components/ClaimNftModal.tsx
```diff
@@ -62,11 +62,11 @@ const ClaimNftModal: React.FC<ClaimNftModalProps> = ({ nft, onSuccess, onClaim,
         <Button
           width="100%"
           onClick={handleConfirm}
-          disabled={!account}
+          disabled={!account || isConfirming}
```
hachiojidev, line 65: This is not necessary. If `isLoading` is true then it is disabled by default.

PR #824, opened 2021-04-07
feat: Use new price api for cake busd price

memoyil:


hachiojidev:
@memoyil I see why you made this change but we want to use the calculated CAKE price from the farms as it is much more accurate. Closing this for now.

PR #825, opened 2021-04-07
Feature: Top Traders - Top 20

ChefHutch:
Expand the Top Traders section to show the next 15 traders on `ExpandableLabel` click

- Between **576 - 852px** (where the card has the full-screen) and again above **1080px**, the expanded wrapper shows as a three-column grid.
- Above & below those values, it renders as a block element.
- Resulting in lots of media queries to make this work across screen sizes.

Preview https://elegant-payne-0f0170.netlify.app/competition

### Below 576 - 1 column
Screenshot 2021-04-08 at 13 25 54

### 576 - 852px - 3 column
Screenshot 2021-04-08 at 12 33 42

### 852 - 1080 - 1 column
Screenshot 2021-04-08 at 12 33 59

### Above 1080 - 3 column
Screenshot 2021-04-08 at 12 34 10

### DIFF
--- a/src/views/TradingCompetition/components/TeamRanks/TopTradersCard/ExpandedGridItem.tsx
+++ b/src/views/TradingCompetition/components/TeamRanks/TopTradersCard/ExpandedGridItem.tsx
@@ -84,7 +84,7 @@ const VolumeText = styled(Text)`
   }
 `
 
-const IconWrapper = styled(Flex)`
+const TeamImageWrapper = styled(Flex)`
   ${({ theme }) => theme.mediaQueries.xl} {
     position: absolute;
     right: 0;
@@ -92,7 +92,10 @@ const IconWrapper = styled(Flex)`
   }
 `
 
-const GridItem: React.FC<{ traderData?: LeaderboardDataItem; icons?: React.ReactNode[] }> = ({ traderData, icons }) => {
+const GridItem: React.FC<{ traderData?: LeaderboardDataItem; teamImages?: React.ReactNode[] }> = ({
+  traderData,
+  teamImages,
+}) => {
   const { address, volume, teamId, rank } = traderData
 
   return (
@@ -114,7 +117,7 @@ const GridItem: React.FC<{ traderData?: LeaderboardDataItem; icons?: React.React
           </Text>
         </Flex>
       </VolumeAddressWrapper>
-      <IconWrapper justifyContent="flex-end">{icons[teamId - 1]}</IconWrapper>
+      <TeamImageWrapper justifyContent="flex-end">{teamImages[teamId - 1]}</TeamImageWrapper>
     </Wrapper>
   )
 }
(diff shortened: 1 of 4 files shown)


src/views/TradingCompetition/components/TeamRanks/TopTradersCard/ExpandedGridItem.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  ${({ theme }) => theme.mediaQueries.md} {
+    margin-right: 16px;
+  }
+`
+
+const GridItem: React.FC<{ traderData?: LeaderboardDataItem }> = ({ traderData }) => {
+  const { address, volume, teamId, rank } = traderData
+
+  const icon = {
```
RabbitDoge, line 74: why don't you use an array ?

ChefHutch: Did this & moved to the level above as it's shared by both components 👌

src/views/TradingCompetition/components/TeamRanks/TopTradersCard/ExpandedGridItem.tsx
```diff
@@ -0,0 +1,125 @@
...
+  margin-left: 4px;
+
+  ${({ theme }) => theme.mediaQueries.xs} {
+    margin-left: 8px;
+  }
+
+  ${({ theme }) => theme.mediaQueries.md} {
+    margin-left: 16px;
+  }
+`
```
hachiojidev, line 47: Use responsive margin prop?

src/views/TradingCompetition/components/TeamRanks/TopTradersCard/ExpandedGridItem.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+const TeamImageWrapper = styled(Flex)`
+  ${({ theme }) => theme.mediaQueries.xl} {
+    position: absolute;
+    right: 0;
+    bottom: 0;
+  }
+`
+
+const GridItem: React.FC<{ traderData?: LeaderboardDataItem; teamImages?: React.ReactNode[] }> = ({
```
hachiojidev, line 95: I have noticed that you frequently make props optional. In general if a prop is optional then there should be code in your component dealing with the cases that is comes down undefined. In both cases I think this component will throw an error/look weird if a prop comes down as `undefined`.

hachiojidev (changes requested):

src/views/TradingCompetition/components/TeamRanks/TopTradersCard/ExpandedGridItem.tsx
ChefHutch: 👍

src/views/TradingCompetition/components/TeamRanks/TopTradersCard/ExpandedGridItem.tsx
ChefHutch: Thanks for this Hachi - I opted to leave that traderData optional but with some sensible defaults. TeamImages is required, passed down via static imports in this case.

hachiojidev (approved):

PR #830, opened 2021-04-08
fix: Fix color attribute shared styles

memoyil:


src/views/TradingCompetition/components/CompetitionHeadingText.tsx
```diff
@@ -21,8 +21,8 @@ const H2SizeStyles = (theme: DefaultTheme) => `
   }
 `
 
-const sharedStyles = (props: HeadingProps) => `
-  color: ${props.textColor ? props.textColor : '#ffffff'};
+const sharedStyles = (props) => `
+  color: ${props.textColor ? props.theme.colors[props.textColor] : '#ffffff'};
```
ChefHutch, line 25: I'm not sure any implementations of the `Heading1Text` / `Heading2Text` components even make use of this prop, I can probably rip this out. Will look tomorrow during trading comp code cleanup

hachiojidev: Please add the typing back unless Typescript can correctly infer the props.

hachiojidev (changes requested):
@ChefHutch Let me know if this PR is okay.

src/views/TradingCompetition/components/CompetitionHeadingText.tsx
memoyil: Done

src/views/TradingCompetition/components/CompetitionHeadingText.tsx
```diff
@@ -22,7 +22,6 @@ const H2SizeStyles = (theme: DefaultTheme) => `
 `
 
 const sharedStyles = (props: HeadingProps) => `
-  color: ${props.textColor ? props.textColor : '#ffffff'};
```
ChefHutch, line 25: We still need this color to be assigned, but the prop is OK to be removed, so just:

```css
color: #ffffff;
```

if removed entirely:
Screenshot 2021-04-09 at 15 02 08

ChefHutch (changes requested):

src/views/TradingCompetition/components/CompetitionHeadingText.tsx
memoyil: Done

ChefHutch (approved):

hachiojidev (approved):

PR #831, opened 2021-04-08
feat(trading-competition-share-img): Add button to generate downloadable sharing image

ChefHutch:
- Adding a 'Share Score' button that allows the user to generate & download a dynamic score image (generated using a canvas el).
- Clicking the button opens the expected 'download' dialog box on desktop browsers.
- As the component renders an image, it also allows mobile users to press & hold to share an image.
- On dApp browsers this functionality isn't there, hence the recommendation that the user screenshots the image.

Preview: https://hutch-share-score.netlify.app/

Screenshot 2021-04-08 at 17 04 05

Screenshot 2021-04-09 at 12 05 27


src/views/TradingCompetition/components/YourScore/CardUserInfo.tsx
```diff
@@ -149,65 +156,75 @@ const CardUserInfo: React.FC<YourScoreProps> = ({ hasRegistered, account, profil
         {subHeadingText}
       </Text>
       {shouldShowUserRanks && (
-        <RanksWrapper>
-          <Flex>
-            {volume > 0 ? (
+        <>
+          {profile.nft && volume > 0 && (
```
ChefHutch, line 160: Only show the button if the profile has an nft attached & the user trading volume is > 0

src/views/TradingCompetition/components/ShareImageModal.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+const ShareImageModal: React.FC<YourScoreProps> = ({ onDismiss, profile, userLeaderboardInformation }) => {
+  const TranslateString = useI18n()
+  const { global, team, volume } = userLeaderboardInformation
+  const [bgImage, setBgImage] = useState(null)
+  const [profileImage, setProfileImage] = useState(null)
+  const canvas = useRef(null)
+
+  useEffect(() => {
+    const bgImages = [StormShare, FlippersShare, CakersShare]
```
ChefHutch, line 23: Main thing I'm looking for advice on here are two things I want to investigate tomorrow morning - 

1. Should I investigate some use of loadable to only import the image for the team that the user belongs to?

2. Currently - all of the images I'm pulling in and the canvas el I'm drawing on are full-size (1600 x 900), even though the preview within the modal is small. This is because when the image is downloaded, it converts this canvas to a png and downloads that. Ideally - I'd separate these so that the user only downloads the full-size canvas when they click the button. I'm gonna play around with this but if you have any thoughts, please share em!

hachiojidev: > Should I investigate some use of loadable to only import the image for the team that the user belongs to?

Yes. You can use the `useProfile()` hook to get the profile (which contains the user's team) from state.

>Currently - all of the images I'm pulling in and the canvas el I'm drawing on are full-size (1600 x 900), even though the preview within the modal is small. This is because when the image is downloaded, it converts this canvas to a png and downloads that. Ideally - I'd separate these so that the user only downloads the full-size canvas when they click the button. I'm gonna play around with this but if you have any thoughts, please share em!

I have never done anything like that but I think that is a good idea. Test it out if it is feasible but don't worry about it too much unless we continue to use this pattern.

ChefHutch: On 1. - my worry was that because I was importing each of the teams' raw share images, it would actually download each of these big bastards regardless of whether the users need them. But - it doesn't look like it does, so I think we're good. 

Testing with syrup storm user, it only downloads the `storm-share` asset

Screenshot 2021-04-09 at 09 26 32

2. Good point on the limited utility unless we're going to reuse this component. I'm going to focus on getting this working on mobile ahead of this optimisation 👌

src/views/TradingCompetition/components/ShareImageModal.tsx
```diff
@@ -0,0 +1,144 @@
+import React, { useRef, useState, useEffect } from 'react'
+import { Modal, Flex, Button, Text, Skeleton, Box } from '@pancakeswap-libs/uikit'
+import styled from 'styled-components'
+import useI18n from 'hooks/useI18n'
+import FlippersShare from '../pngs/flippers-share.png'
+import StormShare from '../pngs/storm-share.png'
+import CakersShare from '../pngs/cakers-share.png'
+import ProfileMask from '../pngs/share-profile-mask.png'
+import MedalGold from '../pngs/medals/medal-gold.png'
```
ChefHutch, line 9: We already have these as .tsx / wrapped SVGs but canvas is dumb and it either wants a hacky way to get raw SVGs in, or images. Either way I'd need to re-add these assets. So... I've re-added these images (each one sub 10kb).

src/views/TradingCompetition/components/YourScore/CardUserInfo.tsx
```diff
@@ -149,65 +156,75 @@ const CardUserInfo: React.FC<YourScoreProps> = ({ hasRegistered, account, profil
...
+                </UserRankBox>
+              )}
               <UserRankBox
                 flex="1"
-                title={TranslateString(1222, 'Rank in team').toUpperCase()}
-                footer={`${TranslateString(999, `#${userLeaderboardInformation && global.toLocaleString()} Overall`)}`}
-                mr="8px"
+                title={TranslateString(1224, 'Your volume').toUpperCase()}
+                footer={TranslateString(1226, 'Since start')}
+                mr={[0, null, '8px']}
```
ChefHutch, line 193: The diff is a bit confused by the code changes early in this file nudging a few things around

```js
 {profile.nft && volume > 0 && (
      <Button mt="12px" variant="secondary" scale="sm" onClick={onPresentShareModal}>
              {TranslateString(999, 'Share Score')}
      </Button>
)}
```

I have changed this dynamic prop, but it's not replacing the 'Rank in Team' mr as the diff suggests

hachiojidev (approved):

PR #841, opened 2021-04-10
Use useTooltip for tooltips

Chef-Cheems:
This PR:
- removes `src/views/Farms/components/Tooltip/Tooltip.tsx` component
- replaces it in Farms table (`Multiplier` and `Liquidity` tooltip on Help icon) to use `useTooltip` from UIKit.
- replaces it in the competition `Timer` (we don't actually show that tooltip now though since we decided to not use block-based timer)

Visually not much has changed compared to old Tooltip. But thanks to Popper.js now it flips down if there is not enough space to show it on top.

## Desktop Before
Screenshot 2021-04-12 at 13 06 45

## Desktop After
Screenshot 2021-04-12 at 13 07 18

## Mobile Before
Screenshot 2021-04-12 at 13 08 05

## Mobile After
Screenshot 2021-04-12 at 13 09 32

## Timer tooltip (not used at the moment but I faked its presence) Before and After
Screenshot 2021-04-12 at 13 42 30
Screenshot 2021-04-12 at 13 53 03


hachiojidev (approved):

src/views/Farms/components/FarmTable/Liquidity.tsx
```diff
@@ -13,41 +15,42 @@ const LiquidityWrapper = styled.div`
...
     : '-'
   const TranslateString = useI18n()
+  const { targetRef, tooltip, tooltipVisible } = useTooltip(
+    TranslateString(999, 'The total value of the funds in this farm’s liquidity pool'),
+    'top-end',
+    'hover',
+    undefined,
+    undefined,
+    [20, 10],
+  )
```
hachiojidev, line 43: @Chef-Cheems Now that I see this in the wild I think it best to make the hooks second parameter an object. It will be much cleaner than this.

```
useTooltip('string', {
  placement: 'top-end',
  trigger: 'hover',
  tooltipOffset: [20,10]
})
```

What do you think?

Chef-Cheems: @Chef-Cheems yeah, agree. I started to think the same too. Will do it this week. 👍

PR #849, opened 2021-04-11
fix(auth): Eager connect not working with bsc extension

Tomburgs:
Fixes Binance Chain Wallet not working with eager connect, requiring users to re-authenticate on every refresh.

This fix defines `BinanceChain` property on window with a getter & setter so that we get notified once the extension has initialized and we can proceed with login.

### DIFF
--- a/src/hooks/useEagerConnect.ts
+++ b/src/hooks/useEagerConnect.ts
@@ -2,16 +2,38 @@ import { useEffect } from 'react'
 import { connectorLocalStorageKey, ConnectorNames } from '@pancakeswap-libs/uikit'
 import useAuth from 'hooks/useAuth'
 
+const _binanceChainListener = async () =>
+  new Promise<void>((resolve) =>
+    Object.defineProperty(window, 'BinanceChain', {
+      get() {
+        return this.bsc
+      },
+      set(bsc) {
+        this.bsc = bsc
+
+        resolve()
+      },
+    }),
+  )
+
 const useEagerConnect = () => {
   const { login } = useAuth()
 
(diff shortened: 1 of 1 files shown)


Chef-Chungus:
Hi - we disabled eager connector for bsc due to a race condition where if you clicked connect too fast it would give an error

Can you confirm is this resolved

https://github.com/pancakeswap/pancake-frontend/pull/579

Tomburgs:
> Hi - we disabled eager connector for bsc due to a race condition where if you clicked connect too fast it would give an error
> 
> Can you confirm is this resolved

Hej hej - I'm not sure if I understand correctly what you mean by clicked connect too fast, are you talking about calling `login` before the BSC extension is initialized?

The issue I fixed is what was described in the TODO comment, which was that the BSC extension does not immediately get injected on the window, thus when we call `activate` on the `useWeb3React` hook and it [tries to call `activate`](https://github.com/NoahZinsmeister/web3-react/blob/d0b038c748a42ec85641a307e6c588546d86afc2/packages/core/src/manager.ts#L132) on the bsc-connector it will [throw an error](https://github.com/binance-chain-npm/bsc-web3-connector/blob/main/src/index.ts#L70) if the extension is not yet initialized.

Now the BSC wallet behaves the same way Metamask and all the other wallets do, what it took was listening for when `BinanceChain` property gets set on the window and only then proceeding with connecting.

hachiojidev (approved):
Very nice.

@Chef-Chungus This _should_ fix the exact issue. It waits until the BinanceChain is available before trying to connect

PR #857, opened 2021-04-12
fix(auth): Clear profile state on logout

Tomburgs:
Currently user profile is not cleared from redux when `logout` action is called.
This causes things such as profile picture to remain in the navbar & edit profile button to remain on the profile page.

### DIFF
--- a/src/hooks/useAuth.ts
+++ b/src/hooks/useAuth.ts
@@ -13,8 +13,11 @@ import { ConnectorNames, connectorLocalStorageKey } from '@pancakeswap-libs/uiki
 import { useToast } from 'state/hooks'
 import { connectorsByName } from 'utils/web3React'
 import { setupNetwork } from 'utils/wallet'
+import { useDispatch } from 'react-redux'
+import { profileClear } from 'state/profile'
 
 const useAuth = () => {
+  const dispatch = useDispatch()
   const { activate, deactivate } = useWeb3React()
   const { toastError } = useToast()
 
@@ -51,7 +54,12 @@ const useAuth = () => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [])
 
-  return { login, logout: deactivate }
+  const logout = useCallback(() => {
+    dispatch(profileClear())
+    deactivate()
+  }, [deactivate, dispatch])
+
+  return { login, logout }
 }
 
 export default useAuth
--- a/src/hooks/useAuth.ts
+++ b/src/hooks/useAuth.ts
@@ -10,14 +10,14 @@ import {
   WalletConnectConnector,
 } from '@web3-react/walletconnect-connector'
 import { ConnectorNames, connectorLocalStorageKey } from '@pancakeswap-libs/uikit'
-import { useToast } from 'state/hooks'
 import { connectorsByName } from 'utils/web3React'
 import { setupNetwork } from 'utils/wallet'
-import { useDispatch } from 'react-redux'
+import { useToast } from 'state/hooks'
 import { profileClear } from 'state/profile'
+import { useAppDispatch } from 'state'
 
 const useAuth = () => {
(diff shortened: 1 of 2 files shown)


src/hooks/useAuth.ts
```diff
@@ -13,8 +13,11 @@ import { ConnectorNames, connectorLocalStorageKey } from '@pancakeswap-libs/uiki
 import { useToast } from 'state/hooks'
 import { connectorsByName } from 'utils/web3React'
 import { setupNetwork } from 'utils/wallet'
+import { useDispatch } from 'react-redux'
```
hachiojidev, line 16: Please use `useAppDispatch`

src/hooks/useAuth.ts
```diff
@@ -51,7 +54,12 @@ const useAuth = () => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [])
 
-  return { login, logout: deactivate }
+  const logout = useCallback(() => {
+    dispatch(profileClear())
+    deactivate()
+  }, [deactivate, dispatch])
+
+  return { login, logout }
```
hachiojidev, line 62: Why the need for `useCallback`?

hachiojidev (changes requested):

src/hooks/useAuth.ts
Tomburgs: So that we don't create a new reference every time this hook is invoked and accidentally trigger unnecessary recalculations / re-renders somewhere.

If that's not a concern I could of course remove it.

RabbitDoge: I would rather disable the eslint rule and put en empty array to make sure the fonction is never re created

src/state/profile/index.tsx
```diff
@@ -17,7 +18,7 @@ export const profileSlice = createSlice({
     profileFetchStart: (state) => {
       state.isLoading = true
     },
-    profileFetchSucceeded: (state, action: PayloadAction<GetProfileResponse>) => {
+    profileFetchSucceeded: (_state, action: PayloadAction<GetProfileResponse>) => {
```
RabbitDoge, line 21: what this change is about ? Please avoid making unneeded changes

src/hooks/useAuth.ts
hachiojidev: >If that's not a concern I could of course remove it.

A valid concern but is it happening? Both of the those functions should keep their reference.

>I would rather disable the eslint rule and put en empty array to make sure the fonction is never re created

While this case might be okay I think it will lead to bugs. As stated on the React docs it is recommended to include all dependencies.

src/state/profile/index.tsx
Tomburgs: It was due to tsserver complaining about a declared but unused argument.
Due to your comment I looked into it, and noticed that VSCode doesn't make these warning obvious, but other editors such as Vim do, I hope this change is alright?

image

See [tsserver 6133](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#flag-unused-declarations-with---nounusedparameters-and---nounusedlocals).

src/hooks/useAuth.ts
Tomburgs: > A valid concern but is it happening? Both of the those functions should keep their reference.

@hachiojidev both of these functions do keep their reference, don't they?
Could you please elaborate more on the desired outcome?

> I would rather disable the eslint rule and put en empty array to make sure the fonction is never re created

@RabbitDoge but we do want this function re-created if the dependencies change, otherwise if `dispatch` or `deactivate` change their reference, but we still call them by their old ones, then our function might not work properly anymore.

hachiojidev:
@Tomburgs Can you resolve conflicts.

hachiojidev (approved):

PR #870, opened 2021-04-13
fix: Missing margin lottery won card

### LINKED ISSUE #869: Lottery Won Card doesn't have margin
**Android:**

**Ipad:**

**Iphone:**

**Desktop:**

memoyil:
Fixes #869 

**Ipad:**

**Android:**

**Iphone:**

**Desktop:**

### DIFF
--- a/src/views/Lottery/components/YourPrizesCard/index.tsx
+++ b/src/views/Lottery/components/YourPrizesCard/index.tsx
@@ -7,21 +7,24 @@ import PrizesWonContent from './PrizesWonContent'
 import NoPrizesContent from './NoPrizesContent'
 
 const StyledCard = styled(Card)`
+  margin-top: 16px;
+
+  ${(props) => `
+    ${props.theme.mediaQueries.sm} {
+      margin-top: 24px;
+    }
+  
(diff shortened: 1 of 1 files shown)


src/views/Lottery/components/YourPrizesCard/index.tsx
```diff
@@ -7,21 +7,24 @@ import PrizesWonContent from './PrizesWonContent'
 import NoPrizesContent from './NoPrizesContent'
 
 const StyledCard = styled(Card)`
+  margin-top: 16px;
+
+  ${(props) => `
```
hachiojidev, line 12: Wrapping this in a function with props is not necessary. This is the pattern we use around the site.

```
${({ theme }) => theme.mediaQueries.sm} {

}
```

hachiojidev (changes requested):

src/views/Lottery/components/YourPrizesCard/index.tsx
memoyil: done

hachiojidev (approved):

PR #886, opened 2021-04-14
fix: Display stake balance correctly when lp price is too high that causes little lp's staked

### LINKED ISSUE #442: Staking liquidity before approving the contract can make the LP disappear?
**Bug Description**
Staking liquidity to a pool seems possible in the liquidity screen without approving the contract first.

**Steps to Reproduce**

1. I bought a smaller coin
2. Added that with BNB to a liquidity pool
3. Confirmed
4. Staked the LP pair directly from the liquidity page
5. Checked the farms and did not find the LP staked.

**Expected Behavior**
If you have the option to stake it at the liquidity page, you would expect that everything else works. If the contract first should be approved, that should be prompted. 

Perhaps I there is another issue but I think I just lost LP tokens.

**Additional Context**
I use Firefox on Ubuntu with the Binance Chain plug-in on the Binance Smart Chain Network.

memoyil:
Fixes #442

### DIFF
--- a/src/components/ModalInput/ModalInput.tsx
+++ b/src/components/ModalInput/ModalInput.tsx
@@ -2,8 +2,6 @@ import React from 'react'
 import styled from 'styled-components'
 import { Text, Button, Input, InputProps, Flex, Link } from '@pancakeswap-libs/uikit'
 import useI18n from 'hooks/useI18n'
-import { BigNumber } from 'bignumber.js'
-import { getBalanceNumber, getFullDisplayBalance } from '../../utils/formatBalance'
 
 interface ModalInputProps {
   max: string
--- a/src/components/ModalInput/ModalInput.tsx
+++ b/src/components/ModalInput/ModalInput.tsx
@@ -2,6 +2,8 @@ import React from 'react'
 import styled from 'styled-components'
 import { Text, Button, Input, InputProps, Flex, Link } from '@pancakeswap-libs/uikit'
 import useI18n from 'hooks/useI18n'
+import { BigNumber } from 'bignumber.js'
+import { getBalanceNumber, getFullDisplayBalance } from '../../utils/formatBalance'
 
 interface ModalInputProps {
   max: string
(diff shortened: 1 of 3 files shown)


src/components/ModalInput/ModalInput.tsx
```diff
@@ -68,15 +70,26 @@ const ModalInput: React.FC<ModalInputProps> = ({
   const TranslateString = useI18n()
   const isBalanceZero = max === '0' || !max
 
-  const displayBalance = isBalanceZero ? '0' : parseFloat(max).toFixed(4)
+  const displayBalance = (balance: string) => {
+    if (isBalanceZero) {
+      return '0'
+    }
+    const balanceNumber = new BigNumber(balance)
+    const balanceNumberFixed = balanceNumber.toFixed(4)
```
hachiojidev, line 78: This seems a bit unnecessary. Why not just use `toLocalString` for both cases?
```
return balance.toLocalString(undefined, { maximumFractionDigits: 4 });
```

src/views/Farms/components/FarmTable/Actions/StakedAction.tsx
```diff
@@ -40,8 +41,14 @@ const Staked: React.FunctionComponent<FarmWithStakedValue> = ({ pid, lpSymbol, l
...
   const addLiquidityUrl = `${BASE_ADD_LIQUIDITY_URL}/${liquidityUrlPathParts}`
-  const rawStakedBalance = getBalanceNumber(stakedBalance)
-  const displayBalance = rawStakedBalance.toLocaleString()
+
+  const displayBalance = useCallback(() => {
+    const balanceNumberLocale = getBalanceNumber(stakedBalance).toLocaleString()
+    if (stakedBalance.gt(0) && balanceNumberLocale === '0') {
+      return getFullDisplayBalance(stakedBalance).toLocaleString()
+    }
+    return balanceNumberLocale
```
hachiojidev, line 50: This too seems a bit off. I think it would be better to run the checks against the raw value `stakedBalance`, then do the necessary conversion. Is that not possible?

src/components/ModalInput/ModalInput.tsx
memoyil: Done

src/views/Farms/components/FarmTable/Actions/StakedAction.tsx
memoyil: We first have to convert it to balanceNumber since stakedBalance doesn't have precision value. But i changed the logic and now we checked against that balance number not the string

hachiojidev (approved):

PR #889, opened 2021-04-14
chore(trading-competition): Remove promotional LeadInBanner from page headers

ChefHutch:
Title says it all. Just removing the LeadInBanner.

We can do something nicer with a PromotionalBanner component when we have a plan for how to use the banners going forward

### DIFF
--- a/src/views/Home/Home.tsx
+++ b/src/views/Home/Home.tsx
@@ -3,6 +3,7 @@ import styled from 'styled-components'
 import { Heading, Text, BaseLayout } from '@pancakeswap-libs/uikit'
 import useI18n from 'hooks/useI18n'
 import Page from 'components/layout/Page'
+import PromotionalBanner from 'components/PromotionalBanner'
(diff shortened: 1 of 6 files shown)


Chef-Chungus (approved):

src/views/Home/Home.tsx
```diff
@@ -81,32 +80,29 @@ const Home: React.FC = () => {
   const TranslateString = useI18n()
 
   return (
-    <>
-      <LeadInBanner />
```
ChefHutch, line 85: Some fairly agressive diffs in here. In each instance the `<LeadInBanner>` is being removed as well as the wrapping fragment

memoyil:
LeadInBanner can stay, but we can empty it for future uses therefore we will not have to add it again in the future.

RabbitDoge (changes requested):
If the component is not used anymore, please remove it

RabbitDoge:
#910

PR #894, opened 2021-04-14
fix(pool): PoolCard height adjusting to siblings

Tomburgs:
Currently when a `PoolCard` gets expanded it's siblings which are on the same row also grow to the same height.

**Before**
image

**After**
image


hachiojidev (approved):
Thanks! The original work on this had this but somewhere along the way we lost it.

PR #895, opened 2021-04-14
refactor: Hide "Add to MetaMask" if no MetaMask in scope

Tomburgs:
In pools there's an option on each card to add that card's token to MetaMask.
Currently this option is still available if there's no MetaMask extension installed, in such cases clicking it will result in errors in console.

### DIFF
--- a/src/views/Pools/components/CardFooter.tsx
+++ b/src/views/Pools/components/CardFooter.tsx
@@ -105,7 +105,7 @@ const CardFooter: React.FC<Props> = ({
 
   const imageSrc = `${BASE_URL}/images/tokens/${tokenName.toLowerCase()}.png`
 
-  const isMetaMaskInScope = !!Reflect.get(window, 'ethereum')?.isMetaMask
+  const isMetaMaskInScope = !!(window as WindowChain).ethereum?.isMetaMask
 
   return (
     <StyledFooter isFinished={isFinished}>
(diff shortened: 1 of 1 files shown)


src/views/Pools/components/CardFooter.tsx
```diff
@@ -105,6 +105,8 @@ const CardFooter: React.FC<Props> = ({
 
   const imageSrc = `${BASE_URL}/images/tokens/${tokenName.toLowerCase()}.png`
 
+  const isMetaMaskInScope = !!Reflect.get(window, 'ethereum')?.isMetaMask
```
hachiojidev, line 108: I have never heard of [Reflect](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect) before. Is there a benefit to using it?

Tomburgs: I like `Reflect` object because it's very semantically correct as to what it does - it houses reflective functions.
Those functions also have very straight forward names, such as `get`, `has`, `set`, etc.

In this case it's also in a way a workaround as to not have to do type casting.

RabbitDoge: It's better to stay coherent with the rest of the codebase

Tomburgs: @RabbitDoge Sure thing!
Hadn't noticed before that there is a `WindowChain` interface I could use.

hachiojidev (approved):

PR #897, opened 2021-04-15
Feature/generic nft handling

hachiojidev:
### NFT Config Changes

**identifier**

This was added to replace `bunnyId` as the method to match an NFT from the `tokenURI` data. 3rd party NFTs will not have this. It is the file name of the NFT image.

**bunnyId**

Renamed `variationId`. Still needed to MINT.

### Contract Changes

Now NFT management is done using a generic `ERC721` helper.


src/config/constants/nfts.ts
```diff
@@ -1,4 +1,16 @@
...
+
+export const IPFS_GATEWAY = 'https://gateway.pinata.cloud'
+
+export const nftSources: NftSource = {
+  [NftType.PANCAKE]: {
+    address: {
+      56: '0xDf7952B35f24aCF7fC0487D01c8d5690a60DBa07',
+      97: '0x60935F36e4631F73f0f407e68642144e07aC7f5E',
+    },
+    identifierKey: 'image',
```
hachiojidev, line 11: Each `tokenURI` might (and probably will) have a different key to access the image (used in the identifier) so we keep track of it here.

src/hooks/useGetWalletNfts.ts
```diff
@@ -54,66 +52,84 @@ const useGetWalletNfts = () => {
...
+    const identifiers = Object.keys(state.tokenIds)
+    return Nfts.filter((nft) => identifiers.includes(nft.identifier))
+  }
 
-          for (let i = 0; i < balanceOf; i++) {
-            tokenIdPromises.push(getTokenIdAndBunnyId(i))
+  useEffect(() => {
+    const fetchCollectibles = async () => {
+      // For each nft source get nft data
+      Object.keys(nftSources).forEach(async (nftSourceType) => {
```
hachiojidev, line 73: The main structural change in this hook is here. We loop through all sources to get the tokenIds (versus just the pancake contract)

src/hooks/useGetWalletNfts.ts
```diff
@@ -54,66 +52,84 @@ const useGetWalletNfts = () => {
...
     if (account) {
-      fetchNfts()
+      fetchCollectibles()
     }
   }, [account, lastUpdated, dispatch])
 
   const refresh = () => dispatch({ type: 'refresh', timestamp: Date.now() })
 
-  return { ...state, refresh }
+  return { ...state, refresh, getTokenIdsByIdentifier, getNftsInWallet }
```
hachiojidev, line 132: Two helpers are the new additions.

src/utils/collectibles.ts
```diff
@@ -0,0 +1,69 @@
...
+    return null
+  }
+
+  if (!uriData[identifierKey]) {
+    return null
+  }
+
+  return Nfts.find((nft) => {
+    return uriData[identifierKey].includes(nft.identifier)
+  })
```
hachiojidev, line 68: This is where we match the NFT data in our config with the data in the `tokenURI`

Tomburgs, line 68: Could do an immediate return instead.

```suggestion
  return Nfts.find((nft) => (
    uriData[identifierKey].includes(nft.identifier)
  ))
```

src/views/Collectibles/components/NftGiveawayModal.tsx
```diff
@@ -41,8 +41,10 @@ const getClaimableNft = (profile: Profile): Nft => {
     return null
   }
 
-  const bunnyId = Object.keys(teamNftMap).find((mapBunnyId) => teamNftMap[mapBunnyId] === profile.team.id)
-  return nftList.find((nft) => nft.bunnyId === Number(bunnyId))
+  const identifier = Object.keys(teamNftMap).find(
+    (mapNftIdentifier) => teamNftMap[mapNftIdentifier] === profile.team.id,
+  )
```
Tomburgs, line 46: Wouldn't using `Object.values` make more sense?

```suggestion
  const identifier = Object.values(teamNftMap).find(
    (teamId) => teamId === profile.team.id,
  )
```

src/views/Collectibles/components/NftList.tsx
```diff
@@ -13,21 +14,30 @@ import EasterNftCard from './NftCard/EasterNftCard'
...
-        const Card = nftComponents[nft.bunnyId] || NftCard
+        const tokenIds = getTokenIdsByIdentifier(nft.identifier)
+
+        const getNftCardComponent = (sortedNft: Nft): FunctionComponent<NftCardProps> => {
+          if (!nftComponents[sortedNft.identifier]) {
+            return NftCard
+          }
+
+          return nftComponents[sortedNft.identifier]
+        }
```
Tomburgs, line 38: Perhaps this could make it clearer?

```suggestion
        const getNftCardComponent = (sortedNft: Nft): FunctionComponent<NftCardProps> => (
          nftComponents[sortedNft.identifier] ?? NftCard
        )
```

src/utils/collectibles.ts
```diff
@@ -0,0 +1,69 @@
...
+    return null
+  }
+
+  if (!identifierKey) {
+    return null
+  }
+
+  if (!uriData[identifierKey]) {
+    return null
+  }
```
Tomburgs, line 64: Why not do this in a single statement?

```suggestion
  if (!(uriData || identifierKey || uriData[identifierKey])) {
    return null
  }
```

src/hooks/useGetWalletNfts.ts
```diff
@@ -54,66 +52,84 @@ const useGetWalletNfts = () => {
...
+              }
 
-          dispatch({ type: 'set_nfts', data: nfts })
-        } else {
-          // Reset it in case of wallet change
+              const [tokenId, identifier] = association as [number, Nft['identifier']]
+
+              return {
+                ...accum,
+                [identifier]: accum[identifier] ? [...accum[identifier].tokenIds, tokenId] : [tokenId],
```
Tomburgs, line 113: Not sure if the cough syrup is getting to me, but how come we expect the accumulated value for identifier to be an object, but return an array?

i.e shouldn't it be this
```typescript
[...accum[identifier], tokenId]
```
instead of this?
```typescript
[...accum[identifier].tokenIds, tokenId]
```

Tomburgs (commented):
I humbly offer you my code review in an attempt to help and learn more of the code base.
Please let me know what you think! :')

src/hooks/useGetWalletNfts.ts
hachiojidev: Oh nice catch! It used to be that but I changed it and missed this. Thanks!

src/utils/collectibles.ts
hachiojidev: Just style. It is easier to read and test in my opinion.

src/views/Collectibles/components/NftList.tsx
hachiojidev: You're right. The original code had more than one case I was checking and I was too lazy to move it back.

src/views/Collectibles/components/NftGiveawayModal.tsx
hachiojidev: Not sure this will work. I need to get the nft `identifier`. That will return the team id

src/utils/collectibles.ts
hachiojidev: I have been moving away from that pattern. Not sure why, both are valid. I just write so much code and inevitably have to add the brackets to add more code.

src/hooks/useGetWalletNfts.ts
```diff
@@ -54,66 +52,84 @@ const useGetWalletNfts = () => {
...
-          const nfts: NftMap = tokenIdsOwnedByWallet.reduce((accum, association) => {
-            if (!association) {
-              return accum
+          if (balanceOf > 0) {
+            const nftDataFetchPromises = []
+
+            // For each index get the tokenId and data associated with it
+            for (let i = 0; i < balanceOf; i++) {
+              nftDataFetchPromises.push(getTokenIdAndData(i))
             }
```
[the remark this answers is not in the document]
hachiojidev: Perhaps, but that, in my opinion, is hard to read. I would rather be more verbose.

hachiojidev:
@Tomburgs Thanks so much for the review!

src/views/Collectibles/components/NftGiveawayModal.tsx
Tomburgs: Ah, yes, you're right, my bad.

src/utils/collectibles.ts
Tomburgs: That makes sense. I'm also at the point where I should have a vim macro for this kind of stuff 😁

src/hooks/useGetWalletNfts.ts
```diff
@@ -54,66 +53,84 @@ const useGetWalletNfts = () => {
...
+      Object.keys(nftSources).forEach(async (nftSourceType) => {
+        const { address: addressObj } = nftSources[nftSourceType as NftType]
+        const address = getAddress(addressObj)
+        const contract = getErc721Contract(address)
+
+        const getTokenIdAndData = async (index: number) => {
+          try {
+            const tokenId = await contract.methods.tokenOfOwnerByIndex(account, index).call()
+            const walletNft = await getNftByTokenId(address, tokenId)
+            return [Number(tokenId), walletNft.identifier]
```
RabbitDoge, line 83: isn't it better to use parseInt ?

src/state/collectibles/index.ts
```diff
@@ -0,0 +1,90 @@
...
+  async (account) => {
+    // For each nft source get nft data
+    const nftSourcePromises = Object.keys(nftSources).map(async (nftSourceType) => {
+      const { address: addressObj } = nftSources[nftSourceType as NftType]
+      const address = getAddress(addressObj)
+      const contract = getErc721Contract(address)
+
+      const getTokenIdAndData = async (index: number) => {
+        try {
+          const tokenId = await contract.methods.tokenOfOwnerByIndex(account, index).call()
```
RabbitDoge, line 30: this is still poorly optimised because we will do one request for each object. Instead we should batch the requests

src/utils/collectibles.ts
```diff
@@ -0,0 +1,73 @@
...
+import { Nft, NftType } from 'config/constants/types'
+import { getAddress } from './addressHelpers'
+import { getErc721Contract } from './contractHelpers'
+
+/**
+ * Gets the identifier key based on the nft address
+ * Helpful for looking up the key when all you have is the address
+ */
+export const getIdentifierKeyFromAddress = (nftAddress: string) => {
+  const nftSource = Object.values(nftSources).find((nftSourceEntry) => {
```
RabbitDoge, line 11: Simple helper functions should have unit tests

RabbitDoge (approved):

PR #855, opened 2021-04-12
feat(farms-pagination): Add pagination to farm table

noblica:
Hey, I saw that [this issue](https://github.com/pancakeswap/pancake-frontend/issues/422) was quite old, so I tried solving it.

Added pagination to the table view of Farms, since loading this many items in the DOM slows it down quite a lot.
First PR on here, I tried to follow all the laid out guidelines, but if you feel that I missed something, or that something should be done differently, please let me know so I can change it.

Here's the preview for large screens:
farms-large

And here's the preview for small screens:
farms-small

I purposely removed the "ROWS PER PAGE" selection for smaller screens, but it can be re-added easily. Also I only touched the table view, didn't touch the card view. Maybe something like infinite loading could be done there, if needed.

This PR closes: https://github.com/pancakeswap/pancake-frontend/issues/422

PR #909, opened 2021-04-15
Feat/farms infinite scroll

noblica:
As @memoyil mentioned in the [pagination PR](https://github.com/pancakeswap/pancake-frontend/pull/855), instead of adding pagination, an infinite scroll would be a better and more universal solution (can be applied to both the card and the table view).

I've created it using an Intersection Observer, and added the observed ref right after the Table. That way, the Table stays it's own component, only shows what is passed to it, and doesn't have to worry about whether it should infinitely scroll or not.

I picked 12 as a number of initially loaded rows/cards, mostly because of how the card view changes on smaller screens. That way, the last row rendered will always be "full".

There was also a type mismatch on the liquidity prop, so I took the liberty of fixing that.

Here's how the change looks like on larger screens:

https://user-images.githubusercontent.com/9339019/114942405-c9784800-9e44-11eb-9085-1228689c6647.mp4

And here's how the change looks like on smaller screens:

https://user-images.githubusercontent.com/9339019/114942457-d9902780-9e44-11eb-80b1-969399b968f7.mp4

Please let me know what you think.

### DIFF
--- a/src/views/Farms/Farms.tsx
+++ b/src/views/Farms/Farms.tsx
@@ -97,6 +97,7 @@ const StyledImage = styled(Image)`
   margin-right: auto;
   margin-top: 58px;
 `
+const NUMBER_OF_FARMS_VISIBLE = 12
 
 const Farms: React.FC = () => {
   const { path } = useRouteMatch()
@@ -163,8 +164,8 @@ const Farms: React.FC = () => {
 
   const loadMoreRef = useRef<HTMLDivElement>(null)
 
-  const [numberOfItemsToShow, setNumberOfItemsToShow] = useState(12)
-  const [observerIsSet, setObserverIsSet] = useState<boolean>()
+  const [numberOfFarmsVisible, setNumberOfFarmsVisible] = useState(NUMBER_OF_FARMS_VISIBLE)
+  const [observerIsSet, setObserverIsSet] = useState(false)
 
   const farmsStakedMemoized = useMemo(() => {
     let farmsStaked = []
@@ -194,7 +195,7 @@ const Farms: React.FC = () => {
       farmsStaked = stakedOnly ? farmsList(stakedInactiveFarms) : farmsList(inactiveFarms)
     }
 
-    return sortFarms(farmsStaked).slice(0, numberOfItemsToShow)
+    return sortFarms(farmsStaked).slice(0, numberOfFarmsVisible)
   }, [
     sortOption,
     activeFarms,
@@ -204,25 +205,23 @@ const Farms: React.FC = () => {
     stakedInactiveFarms,
     stakedOnly,
     stakedOnlyFarms,
-    numberOfItemsToShow,
+    numberOfFarmsVisible,
   ])
 
   useEffect(() => {
-    const options = {
-      rootMargin: '0px',
-      threshold: 1,
-    }
-
-    const callbackFunction = (entries) => {
+    const showMoreFarms = (entries) => {
       const [entry] = entries
       if (entry.isIntersecting) {
-        setNumberOfItemsToShow((currentlyShowing) => currentlyShowing + 12)
+        setNumberOfFarmsVisible((farmsCurrentlyVisible) => farmsCurrentlyVisible + NUMBER_OF_FARMS_VISIBLE)
       }
     }
 
     if (!observerIsSet) {
-      const newObserver = new IntersectionObserver(callbackFunction, options)
-      newObserver.observe(loadMoreRef.current)
+      const loadMoreObserver = new IntersectionObserver(showMoreFarms, {
+        rootMargin: '0px',
+        threshold: 1,
+      })
+      loadMoreObserver.observe(loadMoreRef.current)
       setObserverIsSet(true)
     }
   }, [farmsStakedMemoized, observerIsSet])
--- a/src/views/Farms/Farms.tsx
+++ b/src/views/Farms/Farms.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useCallback, useState } from 'react'
+import React, { useEffect, useCallback, useState, useMemo, useRef } from 'react'
 import { Route, useRouteMatch, useLocation } from 'react-router-dom'
 import { useAppDispatch } from 'state'
 import BigNumber from 'bignumber.js'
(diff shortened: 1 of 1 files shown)


memoyil:
Search and sort is working properly with that solution?

noblica:
@memoyil Yep, as far as I've tried it.
I haven't really touched sorting or filtering. All I did is limit the initial size of rendered data, and increase it every time you hit the end of the list.

Had to move some things inside the `useMemo` hook though, but everything seems to work, as far as I've checked. 

Thanks for the idea, btw!

memoyil:
It looks very good

src/views/Farms/components/FarmTable/Liquidity.tsx
```diff
@@ -2,13 +2,14 @@ import React from 'react'
...
 import { HelpIcon, Text, useTooltip } from '@pancakeswap-libs/uikit'
 import useI18n from 'hooks/useI18n'
+import BigNumber from 'bignumber.js'
 
 const ReferenceElement = styled.div`
   display: inline-block;
 `
 
 export interface LiquidityProps {
-  liquidity: number
```
memoyil, line 11: Why this is needed?

src/views/Farms/Farms.tsx
```diff
@@ -192,16 +174,73 @@ const Farms: React.FC = () => {
...
+    inactiveFarms,
+    isActive,
+    stakedInactiveFarms,
+    stakedOnly,
+    stakedOnlyFarms,
+    numberOfItemsToShow,
+  ])
+
+  useEffect(() => {
+    const options = {
```
memoyil, line 224: Can be defined outside of the function component, so that we don’t need to create a new object

src/views/Farms/components/FarmTable/Liquidity.tsx
noblica: It was a type mismatch, which TypeScript only caught after implementing the `useMemo` hook.

`farm.liquidity` gets assigned to this property, in `Farms.tsx` on line 269, and its type is `BigNumber`.

memoyil: Maybe we should open this as a seperate bugfix

src/views/Farms/Farms.tsx
noblica: Sure, if you think its necessary. We can also inline it. I just wanted to keep things "close" to each other.

Chef-Chungus:
Thanks for your contribution - just running this by @ChefBugs

src/views/Farms/Farms.tsx
```diff
@@ -192,16 +174,73 @@ const Farms: React.FC = () => {
...
 
-  let farmsStaked = []
-  if (isActive) {
-    farmsStaked = stakedOnly ? farmsList(stakedOnlyFarms) : farmsList(activeFarms)
-  } else {
-    farmsStaked = stakedOnly ? farmsList(stakedInactiveFarms) : farmsList(inactiveFarms)
-  }
+  const loadMoreRef = useRef<HTMLDivElement>(null)
+
+  const [numberOfItemsToShow, setNumberOfItemsToShow] = useState(12)
```
RabbitDoge, line 179: Avoid magic numbers in the code. Set this in a variable at the top of the file, with a wisely picked name please.

src/views/Farms/Farms.tsx
```diff
@@ -192,16 +174,73 @@ const Farms: React.FC = () => {
...
+  const [observerIsSet, setObserverIsSet] = useState<boolean>()
```
RabbitDoge, line 180: Initialise to `false` please, and the type will be inferred from the default value, so no need for `boolean`

src/views/Farms/Farms.tsx
```diff
@@ -192,16 +174,73 @@ const Farms: React.FC = () => {
...
+    numberOfItemsToShow,
+  ])
+
+  useEffect(() => {
+    const options = {
+      rootMargin: '0px',
+      threshold: 1,
+    }
+
+    const callbackFunction = (entries) => {
```
RabbitDoge, line 229: Can you use a more explicite name than `callbackFunction` ?

RabbitDoge (changes requested):

src/views/Farms/Farms.tsx
hachiojidev: Both patterns are acceptable.

src/views/Farms/Farms.tsx
noblica: Set it to a constant named `NUMBER_OF_FARMS_VISIBLE`. Also did some more renaming, so the code is a bit easier to understand.
Thanks a lot for the feedback!

src/views/Farms/Farms.tsx
noblica: Renamed to `showMoreFarms`.

src/views/Farms/Farms.tsx
noblica: Removed the `const` and inlined the object with the `IntersectionObserver` constructor.

src/views/Farms/components/FarmTable/Liquidity.tsx
noblica: Created a separate PR [here](https://github.com/pancakeswap/pancake-frontend/pull/915)

memoyil:
@noblica Is it possible to done it like recyclerview in android (always same amount of entry)? In that way we can keep the performance stable even the user reaches the bottom.

noblica:
Hey @memoyil , sorry, I'm not quite sure I understand what you mean? Can you please give me an example of the required behavior?
Currently, on initial load of the page, 12 farms will be loaded. Once the user reaches the end of the list, another 12 will be loaded, until all of them are loaded.
This means that the `To Top` button isn't really accessible anymore, until all the farms have been loaded. Is this what you had in mind, or?

memoyil:
@noblica No it is not required current implementation is sufficent imho. A further improvement can be to make that list always contains 12 farms or whatever the constant is at any time like recycler view of android. But if it makes the code complicated and harder to read, I would stick with this version.

RabbitDoge (commented):
The code looks good. The observer threshold will probably need to be updated, but we will see when we use it how it behave. You just need to rebase on develop and fix the conflict.

noblica:
@RabbitDoge rebase done :+1:

RabbitDoge (approved):

PR #930, opened 2021-04-19
feat: Hook to save value to local storage

hachiojidev:
Adds a generic hook to save values to local storage

- Pool card staked only setting
- Farm view and staked setting

### DIFF
--- a/src/hooks/usePersistState.ts
+++ b/src/hooks/usePersistState.ts
@@ -1,0 +1,24 @@
+import { useEffect, useState } from 'react'
+
+/**
+ * Same as "useState" but saves the value to local storage each time it changes
+ */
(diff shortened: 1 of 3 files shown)


src/hooks/usePersistState.ts
```diff
@@ -0,0 +1,24 @@
+import { useEffect, useState } from 'react'
+
+/**
+ * Same as "useState" but saves the value to local storage each time it changes
+ */
+const usePersistState = (initialValue: any, localStorageKey: string) => {
```
hachiojidev, line 6: Using `any` because it matches `JSON.stringify`

RabbitDoge (approved):

PR #936, opened 2021-04-19
Ifo v1 removal

RabbitDoge:
Remove IFO V1 cards


src/views/Ifos/components/IfoCardV1Data.tsx
```diff
@@ -0,0 +1,20 @@
...
+const IfoCardV1Data: React.FC<Props> = ({ ifo }) => {
+  const publicIfoData = useGetPublicIfoV1Data(ifo)
+  const walletIfoData = useGetWalletIfoV1Data(ifo)
+
+  return (
+    <IfoFoldableCard ifo={ifo} publicIfoData={publicIfoData} walletIfoData={walletIfoData} isInitiallyVisible={false} />
+  )
+}
+
+export default IfoCardV1Data
```
hachiojidev, line 20: Should this file be removed as it is V1?

hachiojidev (approved):

src/views/Ifos/components/IfoCardV1Data.tsx
RabbitDoge: No, this is a new component. The card is generic but we need a wrapper around the card to fetch the right data (V1 or V2)

PR #945, opened 2021-04-20
fix: Web3 utf8ToHex must be called to allow hex-like usernames

634396b:
When a user registers a username, `web3.eth.personal.sign` does not call `web3.utils.utf8ToHex("...")` if hex like string is entered.
Code unit responsible for this in web3-core: https://github.com/ChainSafe/web3.js/blob/5d027191c5cb7ffbcd44083528bdab19b4e14744/packages/web3-core-helpers/src/formatters.js#L225

This will result in a message of `AAAAA` being signed if a user inputs `0x4141414141`

image

### DIFF
--- a/src/views/Profile/ProfileCreation/UserName.tsx
+++ b/src/views/Profile/ProfileCreation/UserName.tsx
@@ -111,7 +111,10 @@ const UserName: React.FC = () => {
 
       const signature = library?.bnbSign
         ? (await library.bnbSign(account, userName))?.signature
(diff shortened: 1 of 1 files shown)


RabbitDoge:
Can you open this pr with a branch starting by `feature/` in order to trigger a preview, so we can easily test this change ?

634396b:
> Can you open this pr with a branch starting by `feature/` in order to trigger a preview, so we can easily test this change ?

#955 done thnx <3!

PR #952, opened 2021-04-20
removed missing twitter image from index.html

ciruz:
I noticed the twitter:image is missing ( https://pancakeswap.finance/hero.png ), so I removed it.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -32,7 +32,6 @@
     />
     <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=5, minimum-scale=1" />
     <meta name="theme-color" content="#000000" />
-    <meta name="twitter:image" content="https://pancakeswap.finance/hero.png" />
     <meta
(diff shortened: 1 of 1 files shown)


public/index.html
```diff
@@ -32,7 +32,6 @@
     />
     <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=5, minimum-scale=1" />
     <meta name="theme-color" content="#000000" />
-    <meta name="twitter:image" content="https://pancakeswap.finance/hero.png" />
```
RabbitDoge, line 35: I don't think we want to remove this. if the image doesn't exist we should find another one.
@ChefBugs  can we have a default image for the meta tag ?

RabbitDoge (changes requested):

public/index.html
ChefBugs: it is in chungie's paws as we speak

PR #955, opened 2021-04-21
fix: Web3 utf8ToHex must be called to allow hex-like usernames registration

634396b:
pr ref from #945


634396b:
Hi sorry is there a reason why actions failed? Is my commenting too long and verbose???

Sorry :( <333

hachiojidev (approved):

PR #960, opened 2021-04-21
feat: Binance token pair ticker

hachiojidev:
Hook for opening a web socket connection to Binance's ticker API. Currently only supports `bnbusdt`.

### DIFF
--- a/src/hooks/ticker.ts
+++ b/src/hooks/ticker.ts
@@ -70,7 +70,7 @@ export const useTokenPairTicker = (tokenPair: TokenPair, connectOnMount: boolean
   // Use a ref instead of state so we mark the connection as closed immediately without
   // triggering a re-render
   const isConnected = useRef(false)
-  const [stream, setStream] = useState(null)
+  const [stream, setStream] = useState<TickerStream>(null)
   const websocket = useRef<WebSocket>(null)
 
   const connect = useCallback(() => {
--- a/src/hooks/ticker.ts
+++ b/src/hooks/ticker.ts
@@ -1,0 +1,164 @@
+import { useCallback, useEffect, useRef, useState } from 'react'
+
+// Only support specific token pairs until we find a method to validate token pairs
+export enum TokenPair {
+  BNBUSDT = 'bnbusdt',
+}
+
+/**
+ * @see https://binance-docs.github.io/apidocs/spot/en/#individual-symbol-ticker-streams
+ */
+export interface StreamData {
+  e: string
+  E: number
+  s: string
+  p: string
+  P: string
+  w: string
+  x: string
+  c: string
+  Q: string
+  b: string
+  B: string
+  a: string
+  A: string
+  o: string
+  h: string
+  l: string
+  v: string
+  q: string
+  O: number
+  C: number
+  F: number
+  L: number
+  n: number
+}
+
+export interface TickerStream {
+  eventType: string
+  eventTime: number
+  symbol: string
+  priceChange: number
+  priceChangePercent: number
+  weightAveragePrice: number
+  firstTrade: number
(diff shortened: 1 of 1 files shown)


src/hooks/ticker.ts
```diff
@@ -0,0 +1,164 @@
...
+              totalTradedQuoteAssetVolume: Number(data.q),
+              statisticsOpenTime: Number(data.O),
+              statisticsCloseTime: Number(data.C),
+              firstTradeId: Number(data.F),
+              lastTradeId: Number(data.L),
+              totalNumberOfTrades: Number(data.n),
+            })
+          }
+        } catch (error) {
+          console.error(`Error parsing data from stream`, error)
```
Chef-Cheems, line 131: Not sure how consumer uses it, but we just `console.error()` the error here and nothing else it will keep old data in the `stream`. Might be not an issue if next event comes fine but if there are a lot of errors for some reason stream might get really out of date.
Not sure if its something that needs fixing, gotta check the consumer code too, but just thinking out loud.

src/hooks/ticker.ts
```diff
@@ -0,0 +1,164 @@
...
+        isConnected.current = true
+      }
+
+      ws.onclose = () => {
+        isConnected.current = false
+
+        if (retriesRemaining.current > 0 && !isManualClose.current && !isConnected.current) {
+          setTimeout(() => {
+            connect()
+            retriesRemaining.current -= 1
```
Chef-Cheems, line 90: I might be wrong, but isn't it the case that `retriesRemaining.current` will never reach 0?
- It starts with 3
- calls `connect()` here
- on line 135 it resets it back to 3 (`retriesRemaining.current = retryCount`) regardless if it re-connected or not
- decrements it by 1 here which makes it 2
- next time it retires it sets it back to 3 and so on

src/hooks/ticker.ts
```diff
@@ -0,0 +1,164 @@
...
+
+  // Use a ref instead of state so we mark the connection as closed immediately without
+  // triggering a re-render
+  const isConnected = useRef(false)
+  const [stream, setStream] = useState<TickerStream>(null)
+  const websocket = useRef<WebSocket>(null)
+
+  const connect = useCallback(() => {
+    if (navigator.onLine && !isConnected.current) {
+      const ws = new WebSocket(`wss://stream.binance.com:9443/ws/streams/${tokenPair}@ticker`)
```
Chef-Cheems, line 78: I wonder if disconnected WebSocket instances will pile up in memory when reconnecting. Sombody in stackoverflow comments also [was wondering about it](https://stackoverflow.com/questions/22431751/websocket-how-to-automatically-reconnect-after-it-dies#comment88921382_23176223) but nobody answered that.

My guess is once `ws.onclose()` handler finishes its job the memory gets cleaned, so it's probably safe to leave it like that.

src/hooks/ticker.ts
Chef-Cheems: Also another less likely scenario that came into my head:
- It tries to reconnect
- Creates new WebSocket instance
- It opens connection and then quickly closes it within 5 seconds setTimeout
- Original retry setTimeout will see isConnected.current as false and will try to reconnect again
- The second instances that opened and closed fires its own setTimeout and tries to reconnect in parallel with 1st reconnect routine.

Not sure how likely it is and if its worth researching, but just a though.

src/hooks/ticker.ts
hachiojidev: I don't think that is a problem here as the websocket instance gets overwritten each time connect is called and it can only do that if the previous once has closed. I am also not entirely sure...

Chef-Chungus (approved):

PR #961, opened 2021-04-21
feat: Get bnb balance hook

hachiojidev:
A hook for getting a wallet's BNB balance

### DIFF
--- a/src/hooks/useTokenBalance.ts
+++ b/src/hooks/useTokenBalance.ts
@@ -67,10 +67,10 @@ export const useGetBnbBalance = () => {
   const [balance, setBalance] = useState(new BigNumber(0))
   const { account } = useWeb3React()
   const { lastUpdated, setLastUpdated } = useLastUpdated()
+  const web3 = useWeb3()
 
   useEffect(() => {
     const fetchBalance = async () => {
-      const web3 = getWeb3NoAccount()
       const walletBalance = await web3.eth.getBalance(account)
       setBalance(new BigNumber(walletBalance))
     }
@@ -78,7 +78,7 @@ export const useGetBnbBalance = () => {
     if (account) {
       fetchBalance()
     }
-  }, [account, lastUpdated, setBalance])
+  }, [account, web3, lastUpdated, setBalance])
 
   return { balance, refresh: setLastUpdated }
 }
--- a/src/hooks/useTokenBalance.ts
+++ b/src/hooks/useTokenBalance.ts
@@ -1,10 +1,10 @@
 import { useEffect, useState } from 'react'
 import BigNumber from 'bignumber.js'
 import { useWeb3React } from '@web3-react/core'
+import { getWeb3NoAccount } from 'utils/web3'
 import { getBep20Contract, getCakeContract } from 'utils/contractHelpers'
 import useWeb3 from './useWeb3'
 import useRefresh from './useRefresh'
-import { getWeb3NoAccount } from 'utils/web3'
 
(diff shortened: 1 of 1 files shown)


src/hooks/useTokenBalance.ts
```diff
@@ -61,4 +63,24 @@ export const useBurnedBalance = (tokenAddress: string) => {
...
+      const walletBalance = await web3.eth.getBalance(account)
+      setBalance(new BigNumber(walletBalance))
+    }
+
+    if (account) {
+      fetchBalance()
+    }
+  }, [account, lastUpdated, setBalance])
+
+  return { balance, refresh: setLastUpdated }
```
hachiojidev, line 83: `refresh` is so the consumer of the hook can re-fetch the balance at their leisure

src/hooks/useTokenBalance.ts
```diff
@@ -61,4 +63,24 @@ export const useBurnedBalance = (tokenAddress: string) => {
...
+export const useGetBnbBalance = () => {
+  const [balance, setBalance] = useState(new BigNumber(0))
+  const { account } = useWeb3React()
+  const { lastUpdated, setLastUpdated } = useLastUpdated()
+
+  useEffect(() => {
+    const fetchBalance = async () => {
+      const web3 = getWeb3NoAccount()
+      const walletBalance = await web3.eth.getBalance(account)
+      setBalance(new BigNumber(walletBalance))
```
Chef-Cheems, line 75: Don't you need `web3.utils.fromWei(walletBalance, "ether")` here? Or it's supposed to be in Wei?

src/hooks/useTokenBalance.ts
ChefHutch: This is neat. I actually hadn't seen you add that `useLastUpdated` hook - it should be pretty useful! I found myself toggling a state boolean at points in the trading comp to effectively do the same thing.

src/hooks/useTokenBalance.ts
hachiojidev: I wasn't aware of that utility but I think most of our balances return as Wei.

ChefHutch (approved):
👍

src/hooks/useTokenBalance.ts
```diff
@@ -61,4 +63,24 @@ export const useBurnedBalance = (tokenAddress: string) => {
...
 }
 
+export const useGetBnbBalance = () => {
+  const [balance, setBalance] = useState(new BigNumber(0))
+  const { account } = useWeb3React()
+  const { lastUpdated, setLastUpdated } = useLastUpdated()
+
+  useEffect(() => {
+    const fetchBalance = async () => {
+      const web3 = getWeb3NoAccount()
```
Chef-Cheems, line 73: Shouldn't this be `const web3 = useWeb3()`?

hachiojidev: Yeah...why does this work? :)

hachiojidev: @Chef-Cheems updated

Chef-Cheems: @hachiojidev you forgot to remove getWeb3NoAccount import :)
Yeah regarding how it works.... I dunno, I guess Web3NoAccount is still valid Web3 that can query account balances via address, so in this case it doesn't really mater? 🤔 
I haven't looked that deep into web3 setup in PCS yet but some stuff seems a bit weird for me, there might be some stuff to optimize or at least to document 😅

Chef-Chungus (approved):

PR #968, opened 2021-04-21
feat(format-balance): Add optional decimalsToAppear argument to getFullDisplayBalance

ChefHutch:
- The addition of percentage sliders in the new pool modals means that the `getFullDisplayBalance` can return strings with numbers that have > 18 decimals.

i.e. try and stake a cool 63% of `18.402512381648115216`
and you receive `"11.59358280043831258608"` (20 decimals),
which throws an error

```
Error: invalid BigNumber string (argument="value", value="11593582800438312586.08", code=INVALID_ARGUMENT, 
```

code [here](https://github.com/pancakeswap/pancake-frontend/blob/55bb60b25cee3a83a997d01075a38fc5fc2f7cab/src/views/Pools/components/PoolCard/StakeModal.tsx#L57-L58) for this balance calculation that's occasionally returning 20-decimals

Screenshot 2021-04-21 at 15 41 51

- By adding an optional prop to `getFullDisplayBalance`, passing a decimal value into `toFixed()`, we can guarantee that the outputted string is 18 decimals, where important.
- We don't always need to guarantee 18 decimals (i.e. for 0 values), so this is an optional prop with a sensible default


src/utils/formatBalance.ts
```diff
@@ -5,8 +5,8 @@ export const getBalanceNumber = (balance: BigNumber, decimals = 18) => {
   return displayBalance.toNumber()
 }
 
-export const getFullDisplayBalance = (balance: BigNumber, decimals = 18) => {
-  return balance.dividedBy(new BigNumber(10).pow(decimals)).toFixed()
+export const getFullDisplayBalance = (balance: BigNumber, decimals = 18, decimalsToAppear = null) => {
```
hachiojidev, line 8: I think it is better in this case to leave `decimalsToAppear` as undefined.

hachiojidev (changes requested):

src/utils/formatBalance.ts
ChefHutch: Yea do you mean reassign the 'null' default to 'undefined', or just un-assign the `null` default? It looks like without some default it then becomes a required function argument.

hachiojidev:
```
(balance: BigNumber, decimals = 18, decimalsToAppear?: number)
```

ChefHutch: Ah yes, no idea what syntax I was playing with that it didn't like... but it wasn't this. I'll amend

ChefHutch: Done 👌

hachiojidev (approved):

PR #975, opened 2021-04-22
feat: Add migration banner

RabbitDoge:
Don't merge before the migration start

### DIFF
--- a/src/components/Banner/Banner.tsx
+++ b/src/components/Banner/Banner.tsx
@@ -1,0 +1,39 @@
+import React from 'react'
+import { Box, ExpandableLabel, Flex } from '@pancakeswap-libs/uikit'
+import styled from 'styled-components'
+import usePersistState from 'hooks/usePersistState'
+
+interface Props {
+  id: string
+  title: React.ReactNode
+}
+
+const Wrapper = styled(Box)<{ isVisible: boolean }>`
+  background: #452a7a;
+  box-shadow: inset 0px -4px 4px rgba(0, 0, 0, 0.25);
+  padding: 32px;
+  ${({ theme }) => theme.mediaQueries.sm} {
+    flex-direction: row;
+  }
+`
+
+const Content = styled.div<{ isVisible: boolean }>`
+  display: ${({ isVisible }) => (isVisible ? 'block' : 'none')};
+`
+
+const Banner: React.FC<Props> = ({ id, title, children, ...props }) => {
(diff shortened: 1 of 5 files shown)


src/components/Banner/MigrationV2.tsx
```diff
@@ -0,0 +1,31 @@
...
+    <Banner id="v2-migration">
+      <Flex mb="16px">
+        <ErrorIcon color="white" width="32px" mr="16px" />
+        <Text color="white" fontSize="24px" bold>
+          ACTION REQUIRED for all LP token holders
+        </Text>
+      </Flex>
+      <Box ml="48px">
+        <Text color="warning" mb="16px" bold>
+          You must complete migration if you want to keep earning from any LP tokens that aren’t staked in PancakeSwap
```
ChefHutch, line 16: All LGTM. I think this text is a bit misleading? Should "aren't" become **are**?

> You must complete migration if you want to keep earning from any LP tokens that **are** staked in PancakeSwap

ChefHutch (approved):

src/components/Banner/MigrationV2.tsx
RabbitDoge: I just copy Bug's text

src/components/Banner/Banner.tsx
```diff
@@ -0,0 +1,34 @@
...
+`
+
+const Content = styled.div<{ isVisible: boolean }>`
+  display: ${({ isVisible }) => (isVisible ? 'block' : 'none')};
+`
+
+const Banner: React.FC<{ id: string }> = ({ id, children, ...props }) => {
+  const [isVisible, setIsVisible] = usePersistState(true, `banner-${id}`)
+  return (
+    <Wrapper isVisible={isVisible} {...props}>
```
ChefHutch, line 25: Totally depends on how much time you have to give to this, but might be useful to have the option to render some content when the banner is collapsed? In this instance it would maybe be helpful to users to show:

```
<Text color="white" fontSize="24px" bold>
 ACTION REQUIRED for all LP token holders
</Text>
```

RabbitDoge: yes, will do

src/components/Banner/MigrationV2.tsx
hachiojidev: Confirmed with Chungus and Bugs, the text should be corrected to **are**

hachiojidev (changes requested):

hachiojidev (approved):

PR #987, opened 2021-04-23
feat(competition-migration-banner): Fix Page styles and banner on trading competition page

ChefHutch:
The TradingCompetition `<Section>` component extends the `<Page>`, component (we wanted the page styles) - but this means the banner is added to every section on the page.

This fixes that, and because of styling difficulties when wrapping the competition page within the `<Page>` component, instead this adds the banner directly to the competition page (note for when we need to remove it).

### DIFF
--- a/src/views/TradingCompetition/index.tsx
+++ b/src/views/TradingCompetition/index.tsx
@@ -14,6 +14,7 @@ import {
   OVER,
   REGISTRATION,
 } from 'config/constants/trading-competition/easterPhases'
+import MigrationV2 from 'components/Banner/MigrationV2'
 import { PrizesIcon, RanksIcon, RulesIcon } from './svgs'
 import {
   DARKBG,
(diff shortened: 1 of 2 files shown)


src/views/TradingCompetition/index.tsx
```diff
@@ -205,108 +206,17 @@ const TradingCompetition = () => {
...
-        <Rules />
-      </Section>
-      <Section backgroundStyle={DARKBG} svgFill={DARKFILL} index={4} intersectionPosition="top">
-        <Flex alignItems="center">
-          {shouldHideCta ? null : (
-            <Flex height="fit-content">
+    <>
+      <Box pb={[null, '32px', null, '42px', null, '52px']}>
+        <MigrationV2 />
+      </Box>
```
ChefHutch, line 212: This is the actual addition. The rest is just a diff spasm

Chef-Cheems (approved):

hachiojidev (approved):

PR #988, opened 2021-04-23
Feature/language provider

hachiojidev:
### Remove translation id and fallback removed
Translations no longer use the translation id or fallback

```
TranslateString(200, 'My Translation')
TranslateString(201, `${variable} times`, { num: variable })
```
...becomes...
```
t'My Translation')
t('%num% times', { num: variable ))
```

### Translation source

Translations will now be from Crowdin GH integration. The master file is located at [src/config/localization/translations.json](https://github.com/pancakeswap/pancake-frontend/blob/bdfe0c36469d7652ebc6c06c4ede19f009831e3b/src/config/localization/translations.json). Anytime a new translation is added to this file Crowdin will automatically add it to the project. Then a PR will be submitted to update the generated locales at `/public/locales/[locale].json`.

Note: The master file is kept inside the `src` directory so we can include it in the bundle. This allows us to have English ready to go without an extra fetch.


src/contexts/Localization/Provider.tsx
```diff
@@ -0,0 +1,106 @@
...
+      currentLanguage: languages[codeFromStorage],
+    }
+  })
+  const { currentLanguage } = state
+
+  useEffect(() => {
+    const fetchInitialLocales = async () => {
+      const codeFromStorage = getLanguageCodeFromLS()
+
+      if (codeFromStorage !== EN.code) {
```
hachiojidev, line 33: We only need to fetch an additional locale if the locale is not English.

src/contexts/Localization/Provider.tsx
```diff
@@ -0,0 +1,106 @@
...
+        ...prevState,
+        isFetching: false,
+      }))
+    }
+
+    fetchInitialLocales()
+  }, [setState])
+
+  const setLanguage = async (language: Language) => {
+    if (!languageMap.has(language.code)) {
```
hachiojidev, line 49: Fetch the new locale

src/contexts/Localization/Provider.tsx
```diff
@@ -0,0 +1,106 @@
...
+      // Merge the EN locale to ensure that any locale fetched has all the keys
+      languageMap.set(language.code, { ...enLocale, ...locale })
+      localStorage.setItem(LS_KEY, language.code)
+
+      setState((prevState) => ({
+        ...prevState,
+        isFetching: false,
+        currentLanguage: language,
+      }))
+    } else {
```
hachiojidev, line 67: Or, if we have it already just use it.

src/contexts/Localization/Provider.tsx
```diff
@@ -0,0 +1,106 @@
...
+        return key
+      }
+
+      const includesVariable = translatedText.includes('%')
+
+      if (includesVariable) {
+        let interpolatedText = translatedText
+
+        Object.keys(data).forEach((dataKey) => {
+          const templateKey = new RegExp(`%${dataKey}%`, 'g')
```
ChefHutch, line 93: Sort of just rubber ducking here, but I think this'll be fine even in the scenario where we have text that includes a %, which isn't referencing a variable?

i.e. Translation: 
> Auto-Compound Bounty: %num%% of all Auto CAKE pool users’ pending yield

Code:
```
t(xxx, `Auto-Compound Bounty: ${callFee}% of all Auto CAKE pool users’ pending yield``, { num: callFee })
```

src/contexts/Localization/Provider.tsx
```diff
@@ -0,0 +1,106 @@
...
+  })
+  const { currentLanguage } = state
+
+  useEffect(() => {
+    const fetchInitialLocales = async () => {
+      const codeFromStorage = getLanguageCodeFromLS()
+
+      if (codeFromStorage !== EN.code) {
+        const enLocale = languageMap.get(EN.code)
+        const currentLocale = await fetchLocale(codeFromStorage)
```
ChefHutch, line 35: Is a 'locale' here describing a json of all translations for that specific language?

src/contexts/Localization/helpers.ts
```diff
@@ -0,0 +1,21 @@
+import { EN, LanguageCode } from '../../config/localization/languages'
+
+const publicUrl = process.env.PUBLIC_URL
+
+export const LS_KEY = 'pancakeswap_lang'
+
+export const fetchLocale = async (code: LanguageCode) => {
+  const response = await fetch(`${publicUrl}/locales/${code}.json`)
```
ChefHutch, line 8: I'm still slightly confused as to how `translations.json` is populated and maintained, and then where the per-language translations are stored? Will we also have `.json` files for each language in the codebase, or are the non-EN translations fetched remotely?

ChefHutch (approved):
All looks great man.

Also - the lightweight docs are a nice touch.

I have more questions than CR feedback, so will give this my blessing and wishing you god speed on Monday 👍

src/contexts/Localization/Provider.tsx
hachiojidev: Good eye - I have confirmed that it works.

src/contexts/Localization/helpers.ts
hachiojidev: Magic Hutch. Magic.

hachiojidev: Once the integration is complete it will all become clear.

RabbitDoge (approved):

PR #991, opened 2021-04-23
Farm Archived tab

Chef-Cheems:
New tab in Farms page for Archived farms.
Fetching logic slightly reworked so it doesn't fetch data for archived pools unless you're looking at them.


hachiojidev (approved):

ChefHutch (approved):
Sorry for the delay man, was having to dig around the code in my editor, not too familiar with the farms stuff. This looks good - nice one.

I wondered if it was worth visually distinguishing the 'archived' option somehow... shown here with 50%. What do you think? One of the UX gang maybe.

Thinking about it now... I guess I'm a bit lost with what the distinction between Archived (v2, discontinued - I think?) and Finished (v1, 0 multiplier - I think), so I suspect users might feel similar..? Might this be worth a tooltip?

Screenshot 2021-04-23 at 14 46 20

PR #992, opened 2021-04-23
Fix userdata for farm cards

Chef-Cheems:
[ ] Explain what your PR does

### DIFF
--- a/src/views/Farms/components/FarmCard/CardActionsContainer.tsx
+++ b/src/views/Farms/components/FarmCard/CardActionsContainer.tsx
@@ -1,11 +1,11 @@
 import React, { useState, useCallback } from 'react'
 import styled from 'styled-components'
 import { provider as ProviderType } from 'web3-core'
+import BigNumber from 'bignumber.js'
 import { getAddress } from 'utils/addressHelpers'
 import { getBep20Contract } from 'utils/contractHelpers'
 import { Button, Flex, Text } from '@pancakeswap-libs/uikit'
 import { Farm } from 'state/types'
-import { useFarmFromSymbol, useFarmUser } from 'state/hooks'
 import useI18n from 'hooks/useI18n'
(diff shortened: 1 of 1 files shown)


src/views/Farms/components/FarmCard/CardActionsContainer.tsx
```diff
@@ -30,8 +30,17 @@ interface FarmCardActionsProps {
...
   const TranslateString = useI18n()
   const [requestedApproval, setRequestedApproval] = useState(false)
-  const { pid, lpAddresses } = useFarmFromSymbol(farm.lpSymbol)
-  const { allowance, tokenBalance, stakedBalance, earnings } = useFarmUser(pid)
+  const { pid, lpAddresses } = farm
+  const {
+    allowance: allowanceAsString = 0,
+    tokenBalance: tokenBalanceAsString = 0,
+    stakedBalance: stakedBalanceAsString = 0,
+    earnings: earningsAsString = 0,
```
hachiojidev, line 38: I think these names are a bit confusing. You have `asString` but they are initialized as numbers.

hachiojidev (approved):

src/views/Farms/components/FarmCard/CardActionsContainer.tsx
Chef-Cheems: ah yeah, I initially called them "xAsNumber", initialized to `0`, then I remembered that they are stored as strings in Redux, renamed to `xAsStrings` but forgot to change default value to `"0"` 😅  Thankfully BigNumber doesn't care what you put it the constructor.
Also - userData in Redux said to have `BigNumber` type but its incorrect, all userData numbers are strings converted later to BigNumber, its just that we don't have much types defined in Redux part of the code so TS never complained. Anyway, I fixed the types in the table improvements PR I made the other day.

PR #1006, opened 2021-04-26
fix: (farms) Disable stake lp button on finished farms

memoyil:


Chef-Cheems (approved):

src/views/Farms/components/FarmTable/Actions/StakedAction.tsx
```diff
@@ -130,7 +130,7 @@ const Staked: React.FunctionComponent<StackedActionProps> = ({
             width="100%"
             onClick={onPresentDeposit}
             variant="secondary"
-            disabled={location.pathname.includes('archived')}
+            disabled={location.pathname.includes('history') || location.pathname.includes('archived')}
```
hachiojidev, line 133: I think `['history', 'archived'].includes(location.pathname)` would be cleaner.

hachiojidev (changes requested):

src/views/Farms/components/FarmTable/Actions/StakedAction.tsx
memoyil: That doesn't work, but i changed to 

`["history", "archived"].some(item => location.pathname.includes(item))`

hachiojidev (approved):

PR #874, opened 2021-04-13
feat: New syrup pool cards

ChefHutch:
**Quite a complete rewrite of the existing `PoolCard` component. The pools view itself has had small changes & the pools / userData state is untouched.**

Preview: https://new-pool-cards.netlify.app/

The whole rework only adds 100 lines of code (net) to the codebase! ✊ 

Screenshot 2021-04-22 at 21 22 37

- There's a lot of UI stuff in here. 
- I have highlighted each smart contract interaction and anything of note via code comments.

PR #1007, opened 2021-04-26
(feat): Cake Vault 

ChefHutch:
🏦 Cake Vault
Preview - https://wen-vault-ser.netlify.app/

- Tooltips PR here can be merged into this main `cake-vault` branch, once reviewed https://github.com/ChefHutch/pancake-frontend/pull/10

**Time-based caveats**: I'm aware of two code stinks 🙏 
- **The modals**. A lot of copy + pasta from the pool cards (rather than making one component that works for both), and some logic that probably shouldn't be in there (i.e. modals being responsible for the approval / confirmation contract call). I'd love to break this out once shipped, and make less-opinionated modal that work across both the vault & the pool cards.
- **State management**. There's some props being passed 4 levels - I know it could (and should) be improved by implementing redux here. I'm not familiar with our redux lib and could probably benefit from 1/2 days to understand it. Maybe implementing it here after shipping would be a good way for me to learn?

**Notes**
- Doge picked up some good stuff in his Syrup pool review, all in this commit - https://github.com/pancakeswap/pancake-frontend/pull/874/commits/dfa6bc5aa092e1b9cbd1775d013df125568f660a . There will be some repeat offences in this PR, I'll catch em.


src/hooks/cakeVault/useGetVaultBountyInfo.ts
```diff
@@ -0,0 +1,50 @@
...
+  }, [cakeVaultContract, refresh])
+
+  useEffect(() => {
+    // Call contract to get total pending cake rewards (doesn't update on refresh)
+    const calculateTotalPendingCakeRewards = async () => {
+      const pendingCakeRewards = await cakeVaultContract.methods.calculateTotalPendingCakeRewards().call()
+      setTotalPendingCakeRewards(new BigNumber(pendingCakeRewards))
+    }
+    calculateTotalPendingCakeRewards()
+  }, [cakeVaultContract])
```
hachiojidev, line 33: Not critical but these might be better put together and batched with `makeBatchRequest`

src/hooks/cakeVault/useGetVaultFees.ts
```diff
@@ -0,0 +1,74 @@
...
+        return { ...prevState, withdrawalFee: contractWithdrawalFee }
+      })
+    }
+
+    const getWithdrawalFeePeriod = async () => {
+      const contractWithdrawalFeeTimePeriod = await cakeVaultContract.methods.withdrawFeePeriod().call()
+      setFees((prevState) => {
+        return { ...prevState, withdrawalFeePeriod: contractWithdrawalFeeTimePeriod }
+      })
+    }
```
hachiojidev, line 54: These should definitely be batched.

src/hooks/cakeVault/useGetVaultSharesInfo.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  useEffect(() => {
+    const getTotalShares = async () => {
+      const shares = await cakeVaultContract.methods.totalShares().call()
+      const { cakeAsBigNumber } = convertSharesToCake(new BigNumber(shares), pricePerFullShare)
+      setTotalShares(new BigNumber(shares))
+      setTotalCakeInVault(cakeAsBigNumber)
+    }
+    getTotalShares()
+  }, [cakeVaultContract, lastUpdated, pricePerFullShare])
```
hachiojidev, line 28: Why are these separate effects?

src/hooks/cakeVault/useGetVaultUserInfo.ts
```diff
@@ -0,0 +1,36 @@
...
+
+const useGetVaultUserInfo = (lastUpdated?: number) => {
+  const { account } = useWeb3React()
+  const cakeVaultContract = useCakeVaultContract()
+  const [userInfo, setUserInfo] = useState({
+    shares: null,
+    cakeAtLastUserAction: null,
+    lastDepositedTime: null,
+    lastUserActionTime: null,
+  })
```
hachiojidev, line 14: Not critical but for state like this please add types.

src/views/Pools/components/CakeVaultCard/VaultStakeModal.tsx
```diff
@@ -0,0 +1,234 @@
...
+    } else {
+      cakeVaultContract.methods
+        .withdraw(sharesAsBigNumber.toString())
+        // .toString() being called to fix a BigNumber error in prod
+        // as suggested here https://github.com/ChainSafe/web3.js/issues/2077
+        .send({ from: account })
+        .on('sending', () => {
+          setPendingTx(true)
+        })
+        .on('receipt', async () => {
```
hachiojidev, line 110: Does this need to be async?

src/views/Pools/components/CakeVaultCard/VaultStakeModal.tsx
```diff
@@ -0,0 +1,234 @@
...
+  const handleDeposit = async (convertedStakeAmount: BigNumber) => {
+    cakeVaultContract.methods
+      .deposit(convertedStakeAmount.toString())
+      // .toString() being called to fix a BigNumber error in prod
+      // as suggested here https://github.com/ChainSafe/web3.js/issues/2077
+      .send({ from: account })
+      .on('sending', () => {
+        setPendingTx(true)
+      })
+      .on('receipt', async () => {
```
hachiojidev, line 137: Same here, async needed?

src/hooks/cakeVault/useGetVaultFees.ts
ChefHutch: My _idea_ is that they can be called selectively. 

like here: https://github.com/pancakeswap/pancake-frontend/pull/1007/files#diff-80c27d122268c36a58ec1320fde37aaf56dcf8085b3a08450a164d04dc2d1ef7R27

`const { callFee } = useGetVaultFees([FeeFunctions.callFee])`

Would that impact the use of `makeBatchRequest`?

Separately, I had an issue where passing the functionsArray as a dependency created an infinite loop ([ln 67 - 69](https://github.com/pancakeswap/pancake-frontend/pull/1007/files#diff-ce17aa16fd3da44cb39ee0bf5c238fac3f55368fe1f62c59d0256d341b507e57R67-R68)). 

Still figuring out how best to manage this 'only make the contract calls you need' pattern.

src/hooks/cakeVault/useGetVaultSharesInfo.ts
ChefHutch: Sometimes, Hachioji, even I don't know why I do certain things.

src/components/Tags.tsx
```diff
@@ -25,4 +25,16 @@ const DualTag = (props) => (
   </Tag>
 )
 
-export { CoreTag, CommunityTag, BinanceTag, DualTag }
+const ManualPoolTag = (props) => (
+  <Tag variant="secondary" outline startIcon={<RefreshIcon width="18px" color="secondary" mr="3px" />} {...props}>
```
RabbitDoge, line 29: margin should be a multiple of 2

src/components/Tags.tsx
```diff
@@ -25,4 +25,16 @@ const DualTag = (props) => (
...
+    Manual
+  </Tag>
+)
+
+const CompoundingPoolTag = (props) => (
+  <Tag variant="success" outline startIcon={<AutoRenewIcon width="18px" color="success" mr="3px" />} {...props}>
```
RabbitDoge, line 35: same

src/config/constants/contracts.ts
```diff
@@ -51,6 +51,10 @@ export default {
     56: '0x23c41D28A239dDCAABd1bb1deF8d057189510066',
     97: '0x24ec6962dbe874F6B67B5C50857565667fA0854F',
   },
+  cakeVault: {
+    56: '0xa80240Eb5d7E05d3F250cF000eEc0891d00b51CC',
+    97: '',
```
RabbitDoge, line 56: do we have a testnet version ? @ChefNyan

src/hooks/cakeVault/useGetVaultBountyInfo.ts
RabbitDoge: definitely yes

src/hooks/cakeVault/useGetVaultBountyInfo.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  useEffect(() => {
+    // Convert estimated rewards to dollars and a cake display value
+    if (estimatedCallBountyReward && stakingTokenPrice) {
+      // Reduce decimals for production
+      const estimatedDollars = getFullDisplayBalance(estimatedCallBountyReward.multipliedBy(stakingTokenPrice), 18, 4)
```
RabbitDoge, line 39: You need to put these magic numbers into constants, or add a comment to explain what it is

src/hooks/cakeVault/useGetVaultFees.ts
RabbitDoge: here you do 4 node requests on each effect run, it's really inefficient. A batch request will send all the 4 calls in 1 request

src/views/Pools/components/CakeVaultCard/UnstakingFeeCountdownRow.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+const UnstakingFeeCountdownRow: React.FC<UnstakingFeeCountdownRowProps> = ({
+  account = true,
+  withdrawalFee,
+  lastDepositedTime,
+  withdrawalFeePeriod = '259200',
+}) => {
+  const TranslateString = useI18n()
+  const { secondsRemaining, hasPerformanceFee } = useWithdrawalFeeTimer(
+    parseInt(lastDepositedTime),
```
RabbitDoge, line 22: add the radix on parseInt

src/views/Pools/components/CakeVaultCard/UnstakingFeeCountdownRow.tsx
```diff
@@ -0,0 +1,39 @@
...
+    parseInt(withdrawalFeePeriod),
```
RabbitDoge, line 23: same

src/views/Pools/components/CakeVaultCard/VaultStakeModal.tsx
```diff
@@ -0,0 +1,237 @@
...
+  const { stakingToken } = pool
+  const cakeVaultContract = useCakeVaultContract()
+  const TranslateString = useI18n()
+  const { theme } = useTheme()
+  const { toastSuccess, toastError } = useToast()
+  const [pendingTx, setPendingTx] = useState(false)
+  const [stakeAmount, setStakeAmount] = useState('')
+  const [percent, setPercent] = useState(0)
+  const { hasPerformanceFee } = useWithdrawalFeeTimer(parseInt(userInfo.lastDepositedTime))
+  const usdValueStaked = stakeAmount && formatNumber(parseFloat(stakeAmount) * stakingTokenPrice)
```
RabbitDoge, line 56: arithmetic operation on balances must use Bignumber

src/views/Pools/components/CakeVaultCard/VaultStakeModal.tsx
```diff
@@ -0,0 +1,237 @@
...
+
+  const handleStakeInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const inputValue = event.target.value ? event.target.value : '0'
```
RabbitDoge, line 59: event.target.value || 0

src/views/Pools/components/PoolCard/CardFooter/ExpandedFooter.tsx
```diff
@@ -75,12 +92,22 @@ const ExpandedFooter: React.FC<ExpandedFooterProps> = ({ pool, account }) => {
...
+          </Flex>
+        </Flex>
+      )}
       <Flex mb="2px" justifyContent="flex-end">
         <LinkExternal bold={false} fontSize="14px" href={earningToken.projectLink}>
           {TranslateString(412, 'View Project Site')}
         </LinkExternal>
       </Flex>
-      {poolContractAddress && (
+      {!isAutoVault && poolContractAddress && (
```
Chef-Cheems, line 110: No need to hide contract for auto I think.

src/views/Pools/index.tsx
```diff
@@ -61,6 +67,7 @@ const Pools: React.FC = () => {
         <FlexLayout>
           <Route exact path={`${path}`}>
             <>
+              <CakeVaultCard pool={cakePool} account={account} />
```
Chef-Cheems, line 70: As discussed - does not get filtered by StakedOnly toggle

src/views/Pools/index.tsx
```diff
@@ -34,11 +36,12 @@ const Pools: React.FC = () => {
     () => finishedPools.some((pool) => pool.userData && new BigNumber(pool.userData.stakedBalance).isGreaterThan(0)),
     [finishedPools],
   )
+  const cakePool = useMemo(() => openPools.filter((pool) => pool.sousId === 0)[0], [openPools])
```
Chef-Cheems, line 39: - `autoCakePool` or `cakeVault` would be better name I think, for a sec I thought its manual cake pool
- better to use `openPools.find((pool) => pool.sousId === 0)` instead of `.filter(...)[0]`

src/hooks/cakeVault/useGetVaultUserInfo.ts
```diff
@@ -0,0 +1,36 @@
+import { useEffect, useState } from 'react'
+import BigNumber from 'bignumber.js'
+import { useWeb3React } from '@web3-react/core'
+import { useCakeVaultContract } from 'hooks/useContract'
+
+const useGetVaultUserInfo = (lastUpdated?: number) => {
```
Chef-Cheems, line 6: Can we put return types in functions? Apart form maybe some simple `number` or `string`, although I'd personally prefer even them to be explicitly defined.
I assume this returns VaultUser that defined in types, but would be nice not to jump around to verify that. Wastes extra 2-3 seconds but it adds up.

Chef-Cheems (commented):
Truckload of nitpicks below. Otherwise seems to work well.

src/config/constants/types.ts
```diff
@@ -76,6 +77,16 @@ export interface PoolConfig {
...
 }
 
+export interface VaultConfig {
+  earningToken: Token
+  stakingToken: Token
+  contractAddress: Address
+  poolCategory: PoolCategory
+  tokenPerBlock: string
+  sortOrder?: number
+  isFinished?: boolean
```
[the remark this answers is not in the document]
Chef-Cheems: ~oh, actually scratch that - if its optional then you don't have to pass it at all which can be nice~

actualllly.... is it even used anywhere?

src/hooks/cakeVault/useGetVaultFees.ts
Chef-Cheems: You could both batch and select which one to call if you don't want all of them all the time.
```
// prepare the calls
...
const feeCalls = {
      performanceFee: getPerformanceFeeCall,
      callFee: getCallFeeCall,
      withdrawalFee: getWithdrawalFeeCall,
      withdrawalFeePeriod: getWithdrawalFeePeriodCall,
    }
const calls = functionsArray.map((functionName) => feeCalls[functionName])
makeBatchRequest(calls)
```
If the call is not in the array it just doesn't get mapped and your batching 3 calls instead of 4 if needed.

As for array as dependency - might need to google that up a bit, its annoying yeah...

OR you can just redefine the parameter as object `calls: FeeCalls` and pass something like
```
{
  performanceFee: false
  callFee: true
  withdrawalFee: true
  withdrawalFeePeriod: true
}
```

and in deps array
```
const { performanceFee, callFee, withdrawalFee, withdrawalFeePeriod } = calls;
useEffect(() => {
  ...
}, [performanceFee, callFee, withdrawalFee, withdrawalFeePeriod ])
```

then you can extract convert it to array to map for batching stuff above.

ChefHutch: Batched and simplified - thanks Cheems but out of time for anything else

src/hooks/cakeVault/useGetVaultSharesInfo.ts
ChefHutch: Oh sorry - not, there is logic here. The second effect is reliant on the thing that the first effect fetches 

`pricePerFullShare` (first effect) is necessary to figure out the `totalCakeInVault` (second effect)

ChefHutch: Batched requests and blended the effects

src/views/Pools/components/CakeVaultCard/VaultStakeModal.tsx
ChefHutch: Cheers Doge, going to go through and check the vault against the kind of changes I made following your reviews in the pool tomorrow (once shipped). I'll be able to strip out some repeated code, too. https://github.com/pancakeswap/pancake-frontend/commit/dfa6bc5aa092e1b9cbd1775d013df125568f660a

src/config/constants/types.ts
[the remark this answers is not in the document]
ChefHutch: Yea I once had big plans for this interface. Good catch! Nothing uses it.

src/views/Pools/components/CakeVaultCard/FeeSummary.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const TranslateString = useI18n()
+  const feeAsDecimal = parseInt(vaultFees.withdrawalFee) / 100
+  const feeInCake = (parseFloat(stakeAmount) * (feeAsDecimal / 100)).toFixed(4)
+
+  return (
+    <>
+      <Flex mt="24px" alignItems="center" justifyContent="space-between">
+        <Text fontSize="14px">{TranslateString(999, 'Unstaking Fee')}</Text>
+        <Text fontSize="14px">
+          {stakeAmount ? feeInCake : '-'} {stakingTokenSymbol}
```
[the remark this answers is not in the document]
ChefHutch: Yea good point... I'm gonna leave it as a `-` just... bcos. 🍆 🍆 🍆 🍆

src/views/Pools/components/PoolCard/CardFooter/ExpandedFooter.tsx
```diff
@@ -44,7 +54,14 @@ const ExpandedFooter: React.FC<ExpandedFooterProps> = ({ pool, account }) => {
...
-              <Balance fontSize="14px" value={getBalanceNumber(totalStaked, stakingToken.decimals)} />
+              <Balance
+                fontSize="14px"
+                value={
+                  isAutoVault
+                    ? getBalanceNumber(totalCakeInVault, stakingToken.decimals)
+                    : getBalanceNumber(totalStaked, stakingToken.decimals)
+                }
+              />
               <Text ml="4px" fontSize="14px">
```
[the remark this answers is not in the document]
ChefHutch: It's a bloody fontSize = 14px epidemic in here!

The small prop is neat, I actually didn't know about it - thanks!

src/views/Pools/components/PoolCard/CardFooter/ExpandedFooter.tsx
ChefHutch: Yea nice, slipped the cake vault contract in there

src/views/Pools/index.tsx
ChefHutch: Too important. Ahaha. I've added it to the Trello 👌

src/views/Pools/index.tsx
ChefHutch: Yea it _is_ actually the manual pool. I just use the `pool` for some data in the vault. I thought that in theory I could make the vault more easily expandable to other tokens that way. 

Will amend the filter > find, amend the name & add a comment here to clarify.

src/hooks/cakeVault/useGetVaultUserInfo.ts
ChefHutch: Typed the state here, I think that'll now get inferred in the return values too

src/views/Pools/components/PoolCard/AprRow.tsx
```diff
@@ -41,17 +64,25 @@ const AprRow: React.FC<AprRowProps> = ({ pool, stakingTokenPrice }) => {
...
     <Flex alignItems="center" justifyContent="space-between">
-      <Text fontSize="16px">{TranslateString(736, 'APR')}:</Text>
+      <Text fontSize="16px">{isAutoVault ? TranslateString(999, 'APY') : TranslateString(736, 'APR')}:</Text>
       {isFinished || !apr ? (
         <Skeleton width="82px" height="32px" />
       ) : (
         <Flex alignItems="center">
-          <Balance fontSize="16px" isDisabled={isFinished} value={apr} decimals={2} unit="%" />
+          <Balance
+            fontSize="16px"
```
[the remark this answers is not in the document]
ChefHutch: I think the Balance component works a bit differently and we need the fontSize here 👌

src/views/Pools/components/PoolCard/CardFooter/ExpandedFooter.tsx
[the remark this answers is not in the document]
ChefHutch: I'll catch some more of these 14px > `small` conversions tomorrow!

RabbitDoge (approved):

PR #1012, opened 2021-04-27
refactor: Move toasts to context to support components

hachiojidev:
### Change

- Toasts are now stored in its own context versus Redux. This was done so we can support adding components to the toast description


src/contexts/ToastsContext/Provider.tsx
```diff
@@ -0,0 +1,55 @@
...
+            title,
+            description,
+            type,
+          },
+          ...currentToasts,
+        ]
+      })
+    },
+    [setToasts],
+  )
```
hachiojidev, line 31: Same logic as the one that was in state.

src/state/achievements/index.ts
```diff
@@ -39,8 +36,6 @@ export const fetchAchievements = (account: string) => async (dispatch: Dispatch)
     dispatch(setAchievements(achievements))
   } catch (error) {
     console.error(error)
-    const title = 'Error fetching achievements'
-    dispatch(push({ id: kebabCase(title), type: toastTypes.DANGER, title }))
```
hachiojidev, line 43: I removed this for 2 reasons:

1. Cannot dispatch toasts anymore.
2. Not sure we should have done this in the first place.

RabbitDoge (approved):

PR #1013, opened 2021-04-27
feat: Use bignumber zero and ten from cache 

memoyil:
Most used static bignumbers in project


Chef-Cheems (approved):
Hm.... at first I though "wait, wouldn't it create bunch of references and stuff?", but it seems like all BigNumber methods return new BigNumber, none of them modifies the object in-place. So I think it should be safe. 
Although I'm not sure how much it really affects performance, but I guess might be tiny bit helpful on low spec devices.
Let's see what other devs think.

Otherwise, all the values replaced properly with BIG_ZERO and BIG_TEN. 👍

hachiojidev:
@memoyil Please fix conflicts

memoyil:
@hachiojidev done

hachiojidev (approved):

PR #1015, opened 2021-04-28
refactor: Block countdown hook

hachiojidev:
This fixes an issue that can arise if the timer is accepting new block numbers to calculate from.

### DIFF
--- a/src/hooks/useGetBlockCountdown.ts
+++ b/src/hooks/useGetBlockCountdown.ts
@@ -1,6 +1,6 @@
 import { useEffect, useRef, useState } from 'react'
 import { BSC_BLOCK_TIME } from 'config'
-import { useBlock } from 'state/hooks'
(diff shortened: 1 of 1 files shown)


src/hooks/useGetBlockCountdown.ts
```diff
@@ -1,41 +1,45 @@
...
+        // Clear previous interval
+        if (timer.current) {
+          clearInterval(timer.current)
+        }
 
         timer.current = setInterval(() => {
-          setSecondsRemaining((prevSecondsRemaining) => {
-            if (prevSecondsRemaining === 0) {
+          setSecondsRemaining((prevSecondsRemaning) => {
+            if (prevSecondsRemaning === 1) {
```
hachiojidev, line 27: Need `1` here - we want it to tick down to zero.

RabbitDoge (approved):

PR #1019, opened 2021-04-28
fix(predictions): Fetch user bets on load

hachiojidev:
`<RoundCard />` now makes a fetch on initial load for the user's bets. This will ensure that the entered tag always shows even if a round has greater than 1,000 bets.

### DIFF
--- a/src/state/hooks.ts
+++ b/src/state/hooks.ts
@@ -11,7 +11,7 @@ import { getAddress } from 'utils/addressHelpers'
 import { getBalanceNumber } from 'utils/formatBalance'
 import useRefresh from 'hooks/useRefresh'
 import { fetchFarmsPublicDataAsync, fetchPoolsPublicDataAsync, fetchPoolsUserDataAsync, setBlock } from './actions'
-import { State, Farm, Pool, ProfileState, TeamsState, AchievementState, PriceState, FarmsState, Bet } from './types'
+import { State, Farm, Pool, ProfileState, TeamsState, AchievementState, PriceState, FarmsState } from './types'
 import { fetchProfile } from './profile'
 import { fetchTeam, fetchTeams } from './teams'
 import { fetchAchievements } from './achievements'
@@ -261,16 +261,6 @@ export const useGetMinBetAmount = () => {
   return useMemo(() => new BigNumber(minBetAmount), [minBetAmount])
 }
 
-export const useGetUserBetByRound = (id: string, account: string): Bet => {
-  const round = useGetRound(id)
-
-  if (!account) {
-    return undefined
-  }
-
-  return round.bets.find((bet) => bet.user.address.toLowerCase() === account.toLocaleLowerCase())
-}
-
 export const useGetIsFetchingHistory = () => {
   return useSelector((state: State) => state.predictions.isFetchingHistory)
 }
@@ -282,6 +272,20 @@ export const useGetHistory = () => {
 export const useGetHistoryByAccount = (account: string) => {
   const bets = useGetHistory()
   return bets ? bets[account] : []
+}
+
+export const useGetBetByRoundId = (account: string, roundId: string) => {
+  const bets = useSelector((state: State) => state.predictions.bets)
+
+  if (!bets[account]) {
+    return null
+  }
+
+  if (!bets[account][roundId]) {
+    return null
+  }
+
+  return bets[account][roundId]
 }
 
 // Collectibles
--- a/src/state/predictions/helpers.ts
+++ b/src/state/predictions/helpers.ts
@@ -221,7 +221,7 @@ type BetHistoryWhereClause = Record<string, string | number | boolean>
 
 export const getBetHistory = async (
   where: BetHistoryWhereClause = {},
-  first = 100,
+  first = 1000,
   skip = 0,
 ): Promise<BetResponse[]> => {
   const response = await request(
--- a/src/state/predictions/helpers.ts
+++ b/src/state/predictions/helpers.ts
@@ -172,12 +172,6 @@ export const getMarketData = async (): Promise<{
       query getMarketData {
         rounds(first: 5, orderBy: epoch, orderDirection: desc) {
           ${getRoundBaseFields()}
-          bets(first: 1000) {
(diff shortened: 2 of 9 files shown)


src/state/hooks.ts
```diff
@@ -261,16 +261,6 @@ export const useGetMinBetAmount = () => {
...
-export const useGetUserBetByRound = (id: string, account: string): Bet => {
-  const round = useGetRound(id)
-
-  if (!account) {
-    return undefined
-  }
-
-  return round.bets.find((bet) => bet.user.address.toLowerCase() === account.toLocaleLowerCase())
-}
-
```
hachiojidev, line 273: This was the core of the problem. It was looking in the round data which could only contain a max number of 1,000 bets. Replaced by `useGetBetByRoundId`

src/state/hooks.ts
```diff
@@ -284,6 +274,20 @@ export const useGetHistoryByAccount = (account: string) => {
...
+    return null
+  }
+
+  if (!bets[account][roundId]) {
+    return null
+  }
+
+  return bets[account][roundId]
+}
+
```
hachiojidev, line 290: Checks the new namespace inside predictions `bets` (which is scoped by account)

src/state/predictions/helpers.ts
```diff
@@ -221,7 +221,7 @@ type BetHistoryWhereClause = Record<string, string | number | boolean>
 
 export const getBetHistory = async (
   where: BetHistoryWhereClause = {},
-  first = 100,
+  first = 1000,
```
hachiojidev, line 224: Just to make sure.

src/state/predictions/index.ts
```diff
@@ -37,6 +38,25 @@ export const fetchBet = createAsyncThunk<{ account: string; bet: Bet }, { accoun
...
+  // This should always return 0 or 1 bet because a user can only place
+  // one bet per round
+  if (betResponses && betResponses.length === 1) {
+    const [betResponse] = betResponses
+    return { account, roundId, bet: transformBetResponse(betResponse) }
+  }
+
+  return { account, roundId, bet: null }
+})
+
```
hachiojidev, line 59: This thunk fetches the user's bet by round id.

src/state/predictions/index.ts
```diff
@@ -85,6 +109,31 @@ export const predictionsSlice = createSlice({
...
+        state.bets = {
+          ...state.bets,
+          [account]: {
+            ...state.bets[account],
+            ...betData,
+          },
+        }
+      }
+
```
hachiojidev, line 136: I added this here because we already have 1,000 bets coming each 10 seconds. It would be a shame not to update the `bets` namespace with the relevant data.

src/state/types.ts
```diff
@@ -217,10 +217,16 @@ export interface RoundData {
...
+export interface HistoryData {
   [key: string]: Bet[]
 }
 
+export interface BetData {
+  [key: string]: {
+    [key: string]: Bet
+  }
+}
+
```
hachiojidev, line 229: This is the new namespace scoped by wallet address.

src/views/Predictions/components/RoundCard/index.tsx
```diff
@@ -13,23 +15,32 @@ interface RoundCardProps {
...
   const currentEpoch = useGetCurrentEpoch()
   const { account } = useWeb3React()
-  const bet = useGetUserBetByRound(round.id, account)
-  const hasEnteredUp = bet && bet.position === BetPosition.BULL
-  const hasEnteredDown = bet && bet.position === BetPosition.BEAR
+  const dispatch = useAppDispatch()
+  const bet = useGetBetByRoundId(account, id)
+  const hasEntered = bet !== null
+  const hasEnteredUp = hasEntered && bet.position === BetPosition.BULL
+  const hasEnteredDown = hasEntered && bet.position === BetPosition.BEAR
```
hachiojidev, line 25: We now have a more reliable way to determine if the user has entered a round.

Chef-Cheems (approved):

ChefHutch (approved):

src/state/predictions/helpers.ts
```diff
@@ -172,12 +172,6 @@ export const getMarketData = async (): Promise<{
       query getMarketData {
         rounds(first: 5, orderBy: epoch, orderDirection: desc) {
           ${getRoundBaseFields()}
-          bets(first: 1000) {
-            ${getBetBaseFields()}
-            user {
-              ${getUserBaseFields()}
-            }
-          }
```
hachiojidev, line 180: This reduces the load significantly. And embarrassingly, it was no longer used.

src/views/Predictions/components/RoundCard/OpenRoundCard.tsx
```diff
@@ -38,16 +42,13 @@ const OpenRoundCard: React.FC<OpenRoundCardProps> = ({
   const [state, setState] = useState<State>({
     isSettingPosition: false,
     position: BetPosition.BULL,
-
-    // We keep this flag in state because it is possible despite polling for new round data
-    // after a user enters a position it returns without the position data. So we want
-    // to be able to update it immediately
-    hasEnteredPosition: false,
```
hachiojidev, line 45: This is no longer need because it resides in state now.

src/views/Predictions/components/RoundCard/OpenRoundCard.tsx
```diff
@@ -74,13 +75,21 @@ const OpenRoundCard: React.FC<OpenRoundCardProps> = ({
...
+      markPositionAsEntered({
+        account,
+        roundId: round.id,
+        partialBet: {
+          hash,
+          position,
+          amount: getBnbAmount(decimalValue).toNumber(),
+        },
+      }),
+    )
```
hachiojidev, line 92: Here is where we set the partial bet immediately so the UI updates. I tried fetching the bet from the graph but it is not updated quick enough.

RabbitDoge (approved):

src/state/predictions/index.ts
hachiojidev: Disregard this comment.

PR #1027, opened 2021-04-29
refactor(predictions): Mobile pages hidden instead of unmounted

hachiojidev:
Update the structure of the mobile view on Predictions so all 3 views are loaded.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "@binance-chain/bsc-connector": "^1.0.0",
     "@crowdin/crowdin-api-client": "^1.10.4",
     "@ethersproject/abi": "^5.0.7",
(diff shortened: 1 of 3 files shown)


Chef-Cheems (approved):
Initially thought that `visibility: hidden` would screw up the layout (since layout space is still allocated), but local testing proved me wrong, works fine 👍

hachiojidev:
> Initially thought that `visibility: hidden` would screw up the layout (since layout space is still allocated), but local testing proved me wrong, works fine 👍

You are 100% right. So I did the lazy method and made the pages `position: absolute;`

RabbitDoge (approved):

PR #1031, opened 2021-04-29
Predictions/clickable history row

hachiojidev:


src/views/Predictions/components/History/HistoricalBet.tsx
```diff
@@ -136,7 +137,7 @@ const HistoricalBet: React.FC<BetProps> = ({ bet }) => {
 
   return (
     <>
-      <StyledBet>
+      <StyledBet onClick={toggleOpen}>
```
RabbitDoge, line 140: Add an `aria-role` on a clickable item that is not supposed to be clickable

RabbitDoge (approved):

PR #1036, opened 2021-04-30
fix: (pools): Finished farms by pool endblock instead of isFinished not shown properly

memoyil:
Issue:

### DIFF
--- a/src/state/hooks.ts
+++ b/src/state/hooks.ts
@@ -33,25 +33,25 @@ import { fetchPoolsStakingLimitsAsync } from './pools'
 export const useFetchPublicData = () => {
   const dispatch = useAppDispatch()
   const { slowRefresh } = useRefresh()
-  const { currentBlock } = useBlock()
+  const web3 = getWeb3NoAccount()
   useEffect(() => {
+    const fetchPoolsPublicData = async () => {
+      const blockNumber = await web3.eth.getBlockNumber()
+      dispatch(fetchPoolsPublicDataAsync(blockNumber))
+    }
     dispatch(fetchFarmsPublicDataAsync())
+    fetchPoolsPublicData()
     dispatch(fetchPoolsStakingLimitsAsync())
-  }, [dispatch, slowRefresh])
+  }, [dispatch, slowRefresh, web3])
 
   useEffect(() => {
-    dispatch(fetchPoolsPublicDataAsync(currentBlock))
-  }, [dispatch, currentBlock])
-
-  useEffect(() => {
-    const web3 = getWeb3NoAccount()
     const interval = setInterval(async () => {
       const blockNumber = await web3.eth.getBlockNumber()
       dispatch(setBlock(blockNumber))
     }, 6000)
 
     return () => clearInterval(interval)
-  }, [dispatch])
+  }, [dispatch, web3])
 }
 
 // Farms
--- a/src/state/hooks.ts
+++ b/src/state/hooks.ts
@@ -36,9 +36,12 @@ export const useFetchPublicData = () => {
   const { currentBlock } = useBlock()
   useEffect(() => {
     dispatch(fetchFarmsPublicDataAsync())
-    dispatch(fetchPoolsPublicDataAsync(currentBlock))
     dispatch(fetchPoolsStakingLimitsAsync())
-  }, [dispatch, slowRefresh, currentBlock])
+  }, [dispatch, slowRefresh])
+
+  useEffect(() => {
+    dispatch(fetchPoolsPublicDataAsync(currentBlock))
+  }, [dispatch, currentBlock])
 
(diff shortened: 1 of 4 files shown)


ChefHutch (commented):
Nice one man, this definitely needs doing else our UI will always lag behind when a pool is finished, but the config hasn't been updated to reflect this.

I wonder if we should store this composite `isPoolFinished` Boolean in redux? We've gone from one, consistent Boolean being used for this logic, to redeclaring the Boolean `const isPoolFinished = pool.isFinished || currentBlock > pool.endBlock` 5 times. That adds some maintenance if any of this changes in future.

I think this is where the public pool data is fetched & stored in redux https://github.com/pancakeswap/pancake-frontend/blob/e4edfe130b1117ec47c816a4d979b20a80b8c8a3/src/state/pools/index.ts#L45

One thing to then solve is how to store this elegantly. It may be a little confusing to have `pool.isFinished` (coming from the config) & `pool.isPoolFinished` (coming from `fetchPoolsPublicDataAsync`), but maybe unavoidable. Have a think? Be good to get the others' thoughts on this too.

memoyil:
@ChefHutch can you review it again?

src/state/pools/index.ts
```diff
@@ -42,16 +42,20 @@ export const PoolsSlice = createSlice({
 export const { setPoolsPublicData, setPoolsUserData, updatePoolsUserData } = PoolsSlice.actions
 
 // Thunks
-export const fetchPoolsPublicDataAsync = () => async (dispatch) => {
+export const fetchPoolsPublicDataAsync = (currentBlock) => async (dispatch) => {
```
ChefHutch, line 45: Can we type this please @memoyil?

src/state/pools/index.ts
```diff
@@ -42,16 +42,20 @@ export const PoolsSlice = createSlice({
...
   const blockLimits = await fetchPoolsBlockLimits()
   const totalStakings = await fetchPoolsTotalStaking()
 
   const liveData = poolsConfig.map((pool) => {
     const blockLimit = blockLimits.find((entry) => entry.sousId === pool.sousId)
     const totalStaking = totalStakings.find((entry) => entry.sousId === pool.sousId)
+    const isPoolFinished =
+      pool.isFinished || (currentBlock > 0 && blockLimit ? currentBlock > Number(blockLimit.endBlock) : false)
```
ChefHutch, line 53: This condition is a bit confusing, can we try and improve readability? One step would be to break out the second part into one or more variables to help describe what each does?

```
currentBlock > 0 && blockLimit ? currentBlock > Number(blockLimit.endBlock) : false
```

ChefHutch (commented):
Cheers Memoyil, couple of bits of feedback. I think these updates may also mean that the `finishedPools` const in the pools index no longer needs the currentBlock check?

https://github.com/pancakeswap/pancake-frontend/blob/0d89801e5b00d3d279791d463836bbc745a5b0c1/src/views/Pools/index.tsx#L28

src/state/pools/index.ts
memoyil: done

src/state/pools/index.ts
memoyil: done

memoyil (commented):
> Cheers Memoyil, couple of bits of feedback. I think these updates may also mean that the `finishedPools` const in the pools index no longer needs the currentBlock check?
> 
> https://github.com/pancakeswap/pancake-frontend/blob/0d89801e5b00d3d279791d463836bbc745a5b0c1/src/views/Pools/index.tsx#L28

removed

ChefHutch (approved):
lgtm, thanks for this Memoyil. Be good to see what the others think 👍

src/state/hooks.ts
```diff
@@ -22,10 +22,11 @@ import { fetchWalletNfts } from './collectibles'
 export const useFetchPublicData = () => {
   const dispatch = useAppDispatch()
   const { slowRefresh } = useRefresh()
+  const { currentBlock } = useBlock()
   useEffect(() => {
     dispatch(fetchFarmsPublicDataAsync())
-    dispatch(fetchPoolsPublicDataAsync())
-  }, [dispatch, slowRefresh])
+    dispatch(fetchPoolsPublicDataAsync(currentBlock))
+  }, [dispatch, slowRefresh, currentBlock])
```
hachiojidev, line 29: Adding the current block here invalidates the necessity for the `slowRefresh`. Is this necssary?

memoyil: I keep the slowrefresh and remove the useblock otherwise we have to send requests unnecessarily every 6 seconds instead of 60. I am getting latest block number in useEffect directly

hachiojidev (approved):

PR #1043, opened 2021-04-30
fix: Change lottery banner

Chef-Cheems:
Screenshot 2021-04-30 at 11 51 21
Screenshot 2021-04-30 at 11 52 09


src/views/Lottery/components/Hero.tsx
```diff
@@ -1,25 +1,41 @@
...
+  color: #ffffff;
   margin-bottom: 24px;
+  text-shadow: 2px 2px 2px #00000040;
 `
 
-const Blurb = styled(Text)`
-  color: #ffffff;
-  font-size: 20px;
+const ComeBack = styled(Text)`
+  background: -webkit-linear-gradient(#ffd800, #eb8c00);
```
hachiojidev, line 14: Why the prefix?

hachiojidev: Styled components automatically adds the vendor prefixes.

Chef-Cheems: I'm not sure why but it doesn't work without it. With prefix works on Chorm, Firefox and Safari.

Chef-Cheems: That what it shows if I remove the prefixes. I can take a look at it later if you don't mind, gonna remove the nextdraw page in later PR. Might need some special wrapper for it to work without prefix or some stupid css thing like that....
Screenshot 2021-04-30 at 12 00 29

src/views/Lottery/components/Hero.tsx
```diff
@@ -60,26 +81,51 @@ const RightWrapper = styled.div`
...
 const Hero = () => {
   const TranslateString = useI18n()
+  const { isXs, isSm } = useMatchBreakpoints()
+
+  const text = (
+    <LeftWrapper>
+      <Title>{TranslateString(999, 'The Lottery Is Changing!')}</Title>
+      <ComeBack>{TranslateString(999, 'Come back soon!')}</ComeBack>
+      <LearnMore>
+        <LearnMoreLink href="https://voting.pancakeswap.finance/#/pancake/proposal/QmU8pcbmBrfbfVQXMMxmkExDq3mYq4s5cbBuFe6uCZzdmX">
```
ChefHutch, line 99: I think you can use the uikit `<Link>` with the `external` prop here, it'll also nicely handle some of the styles you need.

```
<Link bold  mt="20" external href={`https://google.com`}>
```

src/views/Lottery/components/Hero.tsx
```diff
@@ -70,12 +94,17 @@ const Hero = () => {
     <StyledHero>
       <StyledContainer>
         <LeftWrapper>
-          <Title>{TranslateString(708, 'The CAKE Lottery')}</Title>
-          <Blurb>{TranslateString(710, 'Buy tickets with CAKE')}</Blurb>
-          <Blurb>{TranslateString(712, 'Win if 2, 3, or 4 of your ticket numbers match!')}</Blurb>
+          <Title>{TranslateString(999, 'The Lottery Is Changing!')}</Title>
+          <ComeBack>{TranslateString(999, 'Come back soon!')}</ComeBack>
+          <LearnMore>
```
ChefHutch, line 99: I think you could simplify this too and avoid the styled comp?

`<Text bold color="primary">`

hachiojidev (approved):

PR #1045, opened 2021-04-30
fix: (pools) Change claim button to unlock wallet in auto claim bounty when no wallet connected

memoyil:
Before:

image

After:

Screenshot 2021-04-30 at 11 02 30


memoyil:
better to show claim amount even if the wallet is not connected

PR #1046, opened 2021-04-30
feat: (lottery) Add come back soon to lottery card temporarily

memoyil:
Before:

image

After:

Screenshot 2021-04-30 at 11 14 05


Chef-Cheems (commented):
Hey, I'm afraid we have to close this one since we're adjusting the lottery page that gonna stay until new lottery is live. Gonna come today-tomorrow already, sorry...

Although if @hachiojidev or @RabbitDoge wants to merge it then yeah, go ahead.
There are some linter warnings, you can fix them or just yeet it like that, its all gonna change anyway.

PR #1048, opened 2021-04-30
fix: (bounty) Add unlock wallet button bounty modal when no wallet connected

memoyil:
Because it causes an error when user clicks confirm without wallet

Before: 

After:


memoyil:
Closed because https://github.com/pancakeswap/pancake-frontend/pull/1054 contains this fix

PR #1056, opened 2021-04-30
fix: (pools) Non staked Cake vault shown when staked only selected

### LINKED ISSUE #1042: [BUG] Syrup pool : staked only / new pool auto CAKE
In Syrup Pools, check staked only
--> the new pool auto CAKE is displayed even if im not in.

memoyil:
Fixes #1042


src/views/Pools/components/CakeVaultCard/index.tsx
```diff
@@ -35,6 +39,10 @@ const CakeVaultCard: React.FC<{ pool: Pool; account: string }> = ({ pool, accoun
   const stakingTokenPrice = useGetApiPrice(stakingToken.address ? getAddress(stakingToken.address) : '')
   const isLoading = !pool.userData || !userInfo.shares
 
+  if (showStakedOnly && !accountHasSharesStaked) {
+    return <></>
```
Chef-Cheems, line 43: I think `return null` is more common way to render nothing

memoyil: Changed to null

src/views/Pools/components/CakeVaultCard/index.tsx
```diff
@@ -22,7 +22,11 @@ const StyledCardBody = styled(CardBody)<{ isLoading: boolean }>`
   min-height: ${({ isLoading }) => (isLoading ? '0' : '254px')};
 `
 
-const CakeVaultCard: React.FC<{ pool: Pool; account: string }> = ({ pool, account }) => {
+const CakeVaultCard: React.FC<{ pool: Pool; account: string; showStakedOnly: boolean }> = ({
```
ChefHutch, line 25: Let's turn it into an interface as there are 3 props now?

ChefHutch (approved):
Not sure there is a perfectly neat solution to this, but this LGTM.

`The CakeVaultCard` will still make its smart contract calls via the `useGetVaultUserInfo`, `useGetVaultFees` & `useGetVaultSharesInfo` even though nothing is rendered to the DOM, but the alternative (making the `useGetVaultUserInfo` call within the pools view, declaring the `accountHasSharesStaked` boolean and then passing these down) isn't exactly neat.

memoyil:
> Not sure there is a perfectly neat solution to this, but this LGTM.
> 
> The `CakeVaultCard` will still make its smart contract calls via the `useGetVaultUserInfo`, `useGetVaultFees` & `useGetVaultSharesInfo` even though nothing is rendered to the DOM, but the alternative (making the `useGetVaultUserInfo` call within the pools view, declaring the `accountHasSharesStaked` boolean and then passing these down) isn't exactly neat.

Yes either way we have to call, thats why i selected workaroundish way

src/views/Pools/components/CakeVaultCard/index.tsx
memoyil: done

ChefHutch (approved):

hachiojidev (approved):

PR #1059, opened 2021-04-30
chore(apr-row): Make tooltip text conditional depending on whether it's an auto vault

ChefHutch:
Adds conditional 'APR' / 'APY' tooltip text:

APY
Screenshot 2021-04-30 at 17 41 05

APR
Screenshot 2021-04-30 at 17 41 00


Chef-Cheems (approved):
f&*king hell, I was doing similar shit in one of my branches, lol 😅  I'd better check what PRs are out there. 
I actually find a way to (over)optimize it by not showing tooltip and not even using hook at all for APR (since in Figma APR doesn't have tooltip), but yeah I think tooltip on both APR and APY is good.

hachiojidev (approved):

PR #1066, opened 2021-05-02
feat: Change pool collect modal title to Collect from Harvest when compound available

memoyil:
To comply with button in pool card


ChefHutch (approved):

hachiojidev (approved):
I recommend rebasing

PR #1067, opened 2021-05-03
fix: Harvest all button in homepage creates multiple transactions with same nonce value

memoyil:
That causes the second transaction will fail with error, if user sends the first one

### DIFF
--- a/src/views/Home/components/FarmStakingCard.tsx
+++ b/src/views/Home/components/FarmStakingCard.tsx
@@ -1,10 +1,11 @@
 import React, { useState, useCallback } from 'react'
 import styled from 'styled-components'
 import { Heading, Card, CardBody, Button } from '@pancakeswap/uikit'
+import { harvest } from 'utils/callHelpers'
 import { useWeb3React } from '@web3-react/core'
 import { useTranslation } from 'contexts/Localization'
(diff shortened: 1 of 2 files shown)


src/views/Home/components/FarmStakingCard.tsx
```diff
@@ -38,20 +39,22 @@ const FarmedStakingCard = () => {
...
-    } finally {
-      setPendingTx(false)
+    // eslint-disable-next-line no-restricted-syntax
+    for (const farmWithBalance of balancesWithValue) {
+      try {
+        // eslint-disable-next-line no-await-in-loop
+        await harvest(masterChefContract, farmWithBalance.pid, account)
+      } catch (error) {
+        // TODO: find a way to handle when the user rejects transaction or it fails
+      }
```
hachiojidev, line 54: I think a cleaner way is to use `Promise.all`.

```
const harvestPromises = balancesWithValue.map((farmWithBalance) => {
 return harvest(masterChefContract, farmWithBalance.pid, account)
})
 
await Promise.all(harvestPromises)
 ```

memoyil: Nope, that approach doesn't show the next transaction window. It gets stuck at the first one

hachiojidev (approved):

PR #1090, opened 2021-05-03
fix: Unapproved pool does not show enable button

Chef-Cheems:
If user unapproved a pool (e.g. using https://app.unrekt.info/) the pool will not show Enable button (and you'll get failed transaction if you'd try to stake more CAKE). The check for staked balance seems to be not needed at all here.


ChefHutch (approved):
Good catch man 👍

Chef-Cheems:
> Good catch man 👍 

Chungus found it, not me 😄

hachiojidev (approved):

PR #1095, opened 2021-05-03
fix: (prediction): Collect winnings button still enabled after claim

### LINKED ISSUE #1094: Predictions collect winning transaction bug
**Bug Description**
Hi, because i didn't find any repo with the predictions contracts source code, i'm posting here this bug.
When trying to collect winnings, sometimes after collecting them, the button for claim winnings doesn't disappear. If i try to collect them again, instead of an error, there's a transaction with an abnourmous gas fee. I don't know what can happen if someone try to accept and pay that transaction, so i'm posting here a screenshot. Can i ask why there isn't still a repo with predictions smart contracts source code?

Cattura

memoyil:
Fixes #1094 

@hachiojidev I am not sure on that but I think graph api returns old result for a short amount of time (one of the triggers is epoch changes) that makes re-enabling the collect winnings button. Thats why I added a local claimed state that checks conflicts and if so avoid updating state with old graph api state


hachiojidev (approved):
You know I have been aware of this and spent a lot of time overthinking it. This solution works and is simple. Thanks!

PR #1100, opened 2021-05-04
feat: Use BIG_ZERO for conditional initializers

memoyil:


ChefHutch (approved):
Sure - not sure this is cleaner than just conditionally passing 0 into the BigNumber, but good to get the `BIG_ZERO` pattern propagated throughout the codebase 👌

hachiojidev (approved):

PR #1103, opened 2021-05-04
fix(fee-countdown-row): Conditional for handling fee row if a user's fee time has expired

ChefHutch:
Condition for when a user has deposited to the vault but no longer has a withdrawal fee (i.e. 72 hrs has passed since depositing).

- If a user has made a deposit to the CAKE vault (`lastDepositedTime` is true), but doesn't have a fee to pay (`hasUnstakingFee` is false): then show a 0% staking fee
- The impacts both the card, and the withdrawal modal

Staked, no fee
Screenshot 2021-05-04 at 15 05 12

Staked, has fee
Screenshot 2021-05-04 at 13 44 32

Not staked
Screenshot 2021-05-04 at 13 45 25


Chef-Chungus:
@ChefHutch i think here it is better to have "0% unstaking fee" or "no unstaking fee" rather than omitting the row, as people could think it's a bug

ChefHutch:
That makes sense @Chef-Chungus - I've amended.

Chef-Cheems (approved):

RabbitDoge (approved):

PR #1112, opened 2021-05-05
test: Use the new pool abi for tests

RabbitDoge:
The olf ABI is currently still compatible with the new contract.
Execpt for the tests, where we need to fetch the staking token.

[ ] Explain what your PR does

### DIFF
--- a/src/__tests__/config/pools.test.ts
+++ b/src/__tests__/config/pools.test.ts
@@ -1,5 +1,5 @@
 import pools from 'config/constants/pools'
-import { getSouschefContract } from 'utils/contractHelpers'
+import { getSouschefContract, getSouschefV2Contract } from 'utils/contractHelpers'
 
 // Pool 0 is special (cake pool)
 // Pool 78 is a broken pool, not used, and break the tests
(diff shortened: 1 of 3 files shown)


src/utils/contractHelpers.ts
```diff
@@ -71,6 +72,10 @@ export const getSouschefContract = (id: number, web3?: Web3) => {
   const abi = config.poolCategory === PoolCategory.BINANCE ? sousChefBnb : sousChef
   return getContract(abi, getAddress(config.contractAddress), web3)
 }
+export const getSouschefV2Contract = (id: number, web3?: Web3) => {
```
ChefHutch, line 75: I don't really like littering the codebase with comments... but do you think it's worth adding something here explaining that this contract is only needed for the pools tests when calling the `stakedToken` method?

src/__tests__/config/pools.test.ts
```diff
@@ -29,8 +29,15 @@ describe('Config pools', () => {
   it.each(poolsToTest.filter((pool) => pool.stakingToken.symbol !== 'BNB'))(
     'Pool %p has the correct staking token',
     async (pool) => {
-      const contract = getSouschefContract(pool.sousId)
-      const stakingTokenAddress = await contract.methods.syrup().call()
+      let stakingTokenAddress = null
+      try {
+        const contract = getSouschefV2Contract(pool.sousId)
+        stakingTokenAddress = await contract.methods.stakedToken().call()
+      } catch (error) {
```
ChefHutch, line 36: Is the `contract.methods.stakedToken()` call compatible (in theory) with _all_ pools, or just the new ones? 

Just wondering whether it's going to attempt this call, error, and then attempt the `syrup` call for every pool other than the new ones?

ChefHutch (approved):
Couple of Qs but lgtm

Chef-Chungus (approved):

src/__tests__/config/pools.test.ts
RabbitDoge: syrup has been replaced by stakedToken in the new contract, it's basically the only breaking change on the abi, and it was used only in the tests. I didn't find a cleaner solution , but if you have any suggestion you can propose a different implementation

PR #1054, opened 2021-04-30
chore(cake-vault-bounty): Reorganise vault bounty hook, Improve var names. Account check

ChefHutch:
- Reorganise vault bounty hook to no longer convert the contract responses into string display balances
- Improve var names. Fewer 'magic numbers'
- Add ternary to check user account is connected on the confirm button

PR #1113, opened 2021-05-05
[DNM] feature(cake-vault): Implement vault redux store & calculate manual CAKE pool total stake

ChefHutch:
## This work has been amalgamated with a range of improvements in another vault redux PR, all work now here - https://github.com/pancakeswap/pancake-frontend/pull/1208

---

Do not merge - The user redux PR https://github.com/ChefHutch/pancake-frontend/pull/11 should be merged into this PR before ultimately merging into develop.

Preview: https://intial-vault-redux-implementation.netlify.app/

Original goal was to be able to minus the auto CAKE vault total from the manual CAKE pool. This really needed the redux store to be set up for the vault, so I did.

UserData redux PR: https://github.com/ChefHutch/pancake-frontend/pull/11

**Code changes:**
- Setup redux store for public vault information: Fees & Share / Bounty data
- Deleting a load of props passing & old, redundant hooks

**Features**
- Calculate manual CAKE vault staked total correctly:
Screenshot 2021-05-05 at 16 06 54


src/hooks/cakeVault/useGetVaultBountyInfo.ts
```diff
@@ -1,33 +1,23 @@
...
 import { getFullDisplayBalance } from 'utils/formatBalance'
-import makeBatchRequest from 'utils/makeBatchRequest'
 import { getCakeAddress } from 'utils/addressHelpers'
+import BigNumber from 'bignumber.js'
+
+const useGetVaultBountyInfo = () => {
+  const {
+    estimatedCallBountyReward: estimatedCallBountyRewardAsString,
+    totalPendingCakeRewards: totalPendingCakeRewardsAsString,
+  } = useCakeVault()
```
ChefHutch, line 11: Get these values from redux. 

Some calcs to get the $ bounty & CAKE value are still being done in this hook. This will change once this PR is merged, it would be great to get reviews on it https://github.com/pancakeswap/pancake-frontend/pull/1054 🙏

src/state/hooks.ts
```diff
@@ -95,6 +102,23 @@ export const usePoolFromPid = (sousId): Pool => {
   return pool
 }
 
+export const useFetchCakeVault = () => {
+  const { fastRefresh } = useRefresh()
+  const dispatch = useAppDispatch()
+
+  useEffect(() => {
+    dispatch(fetchCakeVaultPublicData())
+  }, [dispatch, fastRefresh])
```
ChefHutch, line 111: This data changes frequently, so fetch on `fastRefresh`

src/state/hooks.ts
```diff
@@ -95,6 +102,23 @@ export const usePoolFromPid = (sousId): Pool => {
...
+
+  useEffect(() => {
+    dispatch(fetchCakeVaultFees())
```
ChefHutch, line 114: This data is relatively static, so fetch once

src/state/pools/fetchVaultPublic.ts
```diff
@@ -0,0 +1,50 @@
+import BigNumber from 'bignumber.js'
+import { convertSharesToCake } from 'views/Pools/helpers'
+import { getCakeVaultContract } from 'utils/contractHelpers'
+import makeBatchRequest from 'utils/makeBatchRequest'
+
+const cakeVaultContract = getCakeVaultContract()
+
+export const fetchPublicVaultData = async () => {
```
ChefHutch, line 8: Calls for public vault data that changes with relative frequency:

> totalShares
> pricePerFullShare
> totalCakeInVault
> estimatedCallBountyReward
> totalPendingCakeRewards

src/state/pools/fetchVaultPublic.ts
```diff
@@ -0,0 +1,50 @@
...
+      totalCakeInVault: totalCakeInVaultEstimate.cakeAsBigNumber.toJSON(),
+      estimatedCallBountyReward: new BigNumber(estimatedCallBountyReward as string).toJSON(),
+      totalPendingCakeRewards: new BigNumber(totalPendingCakeRewards as string).toJSON(),
+    }
+  } catch (error) {
+    return null
+  }
+}
+
+export const fetchVaultFees = async () => {
```
ChefHutch, line 31: Calls for public vault data that will rarely change

> performanceFee
> callFee
> withdrawalFee
> withdrawalFeePeriod

src/state/types.ts
```diff
@@ -58,8 +58,24 @@ export interface FarmsState {
   userDataLoaded: boolean
 }
 
+export interface CakeVault {
+  // When the balance strings are typed as BigNumbers, it throws an error after serialising the BigNumber with .toJSON(): Type 'string' is not assignable to type 'BigNumber'
+  totalShares?: string
```
ChefHutch, line 63: When the balance strings are typed as BigNumbers, it throws an error after serialising the BigNumber with .toJSON():

```
Type 'string' is not assignable to type 'BigNumber'
```

Hence typing them as strings within redux.

src/views/Pools/components/BountyCard.tsx
```diff
@@ -32,10 +31,11 @@ const InlineText = styled(Text)`
 
 const BountyCard = () => {
   const { t } = useTranslation()
-  const { fastRefresh } = useRefresh()
-  const { dollarCallBountyToDisplay, cakeCallBountyToDisplay, totalPendingCakeRewards } = useGetVaultBountyInfo(
-    fastRefresh,
-  )
+  const { dollarCallBountyToDisplay, cakeCallBountyToDisplay, totalPendingCakeRewards } = useGetVaultBountyInfo()
+  const {
+    fees: { callFee },
```
ChefHutch, line 36: Nested deconstruction, hope this is OK

src/views/Pools/components/CakeVaultCard/RecentCakeProfitBalance.tsx
```diff
@@ -2,20 +2,18 @@ import React from 'react'
...
 }
 
-const RecentCakeProfitBalance: React.FC<RecentCakeProfitBalanceProps> = ({
-  cakeAtLastUserAction,
-  userShares,
-  pricePerFullShare,
-}) => {
+const RecentCakeProfitBalance: React.FC<RecentCakeProfitBalanceProps> = ({ cakeAtLastUserAction, userShares }) => {
+  const { pricePerFullShare: pricePerFullShareAsString } = useCakeVault()
+  const pricePerFullShare = new BigNumber(pricePerFullShareAsString)
```
ChefHutch, line 16: Not sure if there's a more elegant solution for this?

1. Serialise BigNumber as string when storing it in Redux
2. Fetch string from Redux store
3. Initialise as BigNumber when used in code

src/views/Pools/components/PoolCard/CardFooter/ExpandedFooter.tsx
```diff
@@ -62,21 +61,25 @@ const ExpandedFooter: React.FC<ExpandedFooterProps> = ({
     { placement: 'bottom-end' },
   )
 
+  const getTotalStakedBalance = () => {
+    if (isAutoVault) {
+      return getBalanceNumber(totalCakeInVault, stakingToken.decimals)
+    }
+    if (isManualCakePool) {
+      const manualCakeTotalMinusAutoVault = new BigNumber(totalStaked).minus(totalCakeInVault)
+      return getBalanceNumber(manualCakeTotalMinusAutoVault, stakingToken.decimals)
```
ChefHutch, line 70: This is the logic to calculate the correct manual cake vault total

src/views/Pools/index.tsx
```diff
@@ -17,6 +17,7 @@ import PoolTabButtons from './components/PoolTabButtons'
 import BountyCard from './components/BountyCard'
 
 const Pools: React.FC = () => {
+  useFetchCakeVault()
```
ChefHutch, line 20: Populating the vault Redux store. Opted to do it within the Pools view rather than app.tsx but easily changed if you'd prefer that?

src/state/types.ts
```diff
@@ -58,8 +58,23 @@ export interface FarmsState {
   userDataLoaded: boolean
 }
 
+export interface CakeVault {
+  totalShares?: string
```
ChefHutch, line 62: When the balance strings are typed as BigNumbers, it throws an error after serialising the BigNumber with .toJSON():

```
Type 'string' is not assignable to type 'BigNumber'
```

Hence typing them as strings within redux.

src/views/Pools/components/CakeVaultCard/RecentCakeProfitBalance.tsx
ChefHutch: So I've changed this in this commit - https://github.com/pancakeswap/pancake-frontend/pull/1113/commits/92f04d72bf2037e8a7a151a70199b1f792c6ef36 

There's now an intermediary hook `useGetVaultPublicData` that parses the serialised string and returns an actual BigNumber. This feels slightly like overkill, but it _does_ save on a lot of repeated code when instantiating these strings as BigNumbers.

New hook - https://github.com/pancakeswap/pancake-frontend/pull/1113/commits/92f04d72bf2037e8a7a151a70199b1f792c6ef36#diff-6a15e2082467c89c26382ca7c5b411677f8abc61512c4ecefe4515fa5aeb4123R1-R18

Example of how this saves on doing this repeatedly in the codebase - https://github.com/pancakeswap/pancake-frontend/pull/1113/commits/92f04d72bf2037e8a7a151a70199b1f792c6ef36#diff-2fadc663a1c0af7a175e5fe2156f5eac5c330a048c2daef4586416b4dfd0a6e3L15-R15

Let me know if you think there's a better pattern for this?

src/state/types.ts
ChefHutch: Farm & Pool state doesn't seem to need to type these as BigNumbers... although I'm struggling to see how they've got around it.

Generally looking for feedback on serialisation of these contract responses, example data:

`totalShares: "18574889782429049977066502"`

src/hooks/cakeVault/useGetVaultPublicData.ts
```diff
@@ -0,0 +1,18 @@
+import BigNumber from 'bignumber.js'
```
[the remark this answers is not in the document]
ChefHutch: `useCakeVault` retrieves the serialised string values from state, this hook just instantiates them as `BigNumber`s (to save having to do that each time the strings are used elsewhere in the codebase)

The proper state storage & retrieval of BigNumbers is probably my biggest question in this PR & the branched one - [#2](https://github.com/ChefHutch/pancake-frontend/pull/11).

I'm not sure the best way to do this though, these two code comments are on this topic. I'd appreciate your input Doge

https://github.com/pancakeswap/pancake-frontend/pull/1113#discussion_r626666314

https://github.com/pancakeswap/pancake-frontend/pull/1113#discussion_r626678246

src/state/types.ts
Chef-Cheems: Farms having values as BigNumber is just a mistake, and it wasn't throwing errors just cause TS didn't actually check them after bunch of hops through helper functions/hooks. It was actually fixed in some places to use string. So everything's fine here 👍

src/views/Pools/components/CakeVaultCard/RecentCakeProfitBalance.tsx
Chef-Cheems: I think this is fine for now.
Other way would be to just convert all values that should be BigNumber to BigNumber inside `useCakeVault` and keep other values as strings if needed (like fees here)? Then you have just one hook for everything, destructure what you need and here you go.

Actually interesting topic of storing class instances inside global state. As I understand with redux its not possible, I wonder if Context could be better option in this case (not sure though, might have problems with that too).

src/state/types.ts
ChefHutch: Ah excellent news.

src/state/pools/fetchVaultPublic.ts
```diff
@@ -0,0 +1,61 @@
...
+  try {
+    const [performanceFee, callFee, withdrawalFee, withdrawalFeePeriod] = await makeBatchRequest([
+      cakeVaultContract.methods.performanceFee().call,
+      cakeVaultContract.methods.callFee().call,
+      cakeVaultContract.methods.withdrawFee().call,
+      cakeVaultContract.methods.withdrawFeePeriod().call,
+    ])
+    return {
+      performanceFee: parseInt(performanceFee as string, 10),
```
Chef-Cheems, line 46: Is `as string` needed here? I guess its because the return type of calls is `any` but it looks like TS doesn't complain if you pass `any` to `parseInt`

ChefHutch: You could go `as any` but it seems reasonable to type it?

Removing the type all together throws an error:
```
Argument of type 'unknown' is not assignable to parameter of type 'string'
```

src/state/types.ts
hachiojidev: If the number is a balance that needs decimal conversion it should be stored as a big number. In Redux `bn.toJSON()` and out of redux `new BigNumber(bn)`. It is a more overhead I admit but I think much better if we consistently do that.

src/views/Pools/index.tsx
hachiojidev: I think this is what we should do now moving forward with the exception of data that is used many places throughout the site.

ChefHutch:
This work has been amalgamated with a range of improvements in another vault PR, all work now here - #1208

PR #1132, opened 2021-05-06
Remove discontinued farms button and banner

Chef-Chungus:


src/views/Farms/components/FarmTabButtons/index.tsx
```diff
@@ -9,7 +9,7 @@ interface FarmTabButtonsProps {
   hasStakeInArchivedFarms: boolean
```
hachiojidev, line 9: We need to remove this and the passing of the prop in `Farms.tsx` line 391.

hachiojidev (changes requested):

hachiojidev (approved):

PR #1156, opened 2021-05-07
feat(withdrawal-fee-timer): Make vault timer update per second that passes

ChefHutch:
Previously, the Vault withdrawal fee timer would fetch the seconds remaining once on component mount, now it sets an interval that updates the `currentSeconds` value every 1000ms, meaning the timer will count down as the user sits on the page.

Screenshot 2021-05-07 at 13 27 38

### DIFF
--- a/src/hooks/cakeVault/useWithdrawalFeeTimer.ts
+++ b/src/hooks/cakeVault/useWithdrawalFeeTimer.ts
@@ -6,14 +6,6 @@ const useWithdrawalFeeTimer = (lastDepositedTime: number, withdrawalFeePeriod = 
   const [currentSeconds, setCurrentSeconds] = useState(Math.floor(Date.now() / 1000))
 
   useEffect(() => {
-    const tick = () => {
-      setCurrentSeconds((prevSeconds) => prevSeconds + 1)
-    }
-    const timerInterval = setInterval(() => tick(), 1000)
-    return () => clearInterval(timerInterval)
-  }, [])
-
-  useEffect(() => {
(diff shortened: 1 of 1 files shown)


src/hooks/cakeVault/useWithdrawalFeeTimer.ts
```diff
@@ -3,17 +3,25 @@ import { useEffect, useState } from 'react'
...
+
+  useEffect(() => {
+    const feeEndTime = lastDepositedTime + withdrawalFeePeriod
+    const secondsRemainingCalc = feeEndTime - currentSeconds
     const doesUnstakingFeeApply = secondsRemainingCalc > 0
     if (doesUnstakingFeeApply) {
       setSecondsRemaining(secondsRemainingCalc)
       setHasUnstakingFee(true)
     }
-  }, [lastDepositedTime, withdrawalFeePeriod, setSecondsRemaining])
```
hachiojidev, line 16: I don't see a clear reason why this needs to be 2 hooks. Am I missing something?

ChefHutch: Yea I thought that setting the interval within the same `useEffect` that has the `currentSeconds` dependency would create new intervals every time `currentSeconds` is updated - creating a rerender loop..

But it looks like the `return` also cleans up effects from the previous render before running the effects next time? Interesting, I only thought it did this on unmount. 

Is it not less performance to have this all within the one effect, if it's having to clear & set an interval each time the effect fires (happening every 1000ms), rather than just once on mount? I've pushed up this change regardless.

hachiojidev (approved):

PR #1163, opened 2021-05-08
Remove unused addresses

Akkii4:


hachiojidev:
Thank you for submitting this. At this time we will keep it in here because soon we will be moving these addresses to an external package.

PR #1164, opened 2021-05-09
fix: specific section is not displayed in Prediction page

zpdldhkdl:
The screen is not displayed when the width is 853 to 967.

issue

### DIFF
--- a/src/views/Predictions/Mobile.tsx
+++ b/src/views/Predictions/Mobile.tsx
@@ -21,7 +21,7 @@ const StyledMobile = styled.div`
   height: 100%;
   max-height: 100%;
 
-  ${({ theme }) => theme.mediaQueries.lg} {
+  ${({ theme }) => theme.mediaQueries.xl} {
     display: none;
   }
 `
--- a/src/views/Predictions/index.tsx
+++ b/src/views/Predictions/index.tsx
@@ -34,7 +34,7 @@ const Predictions = () => {
   const isChartPaneOpen = useIsChartPaneOpen()
   const dispatch = useAppDispatch()
   const initialBlock = useInitialBlock()
-  const isDesktop = isXl || false
+  const isDesktop = isXl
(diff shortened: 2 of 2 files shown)


src/views/Predictions/index.tsx
```diff
@@ -27,14 +27,14 @@ import ChartDisclaimer from './components/ChartDisclaimer'
...
-  const { isLg, isXl } = useMatchBreakpoints()
+  const { isXl } = useMatchBreakpoints()
   const [hasAcceptedRisk, setHasAcceptedRisk] = usePersistState(false, 'pancake_predictions_accepted_risk')
   const [hasAcceptedChart, setHasAcceptedChart] = usePersistState(false, 'pancake_predictions_chart')
   const status = useGetPredictionsStatus()
   const isChartPaneOpen = useIsChartPaneOpen()
   const dispatch = useAppDispatch()
   const initialBlock = useInitialBlock()
-  const isDesktop = isLg || isXl
+  const isDesktop = isXl || false
```
hachiojidev, line 37: No need for `|| false`

src/views/Predictions/Mobile.tsx
```diff
@@ -21,7 +21,7 @@ const StyledMobile = styled.div`
   height: 100%;
   max-height: 100%;
 
-  ${({ theme }) => theme.mediaQueries.lg} {
+  ${({ theme }) => theme.mediaQueries.xl} {
```
hachiojidev, line 24: Can you double check the rest of the media queries in Predictions, they all need to go to `xl`

hachiojidev (changes requested):

src/views/Predictions/Mobile.tsx
zpdldhkdl: I checked it again and there is no need to change anything else.
@hachiojidev

hachiojidev (approved):

PR #1166, opened 2021-05-10
refactor(predictions): Sync bet status

hachiojidev:
"Collect Winnings" on card and history pane is now synced.

### DIFF
--- a/src/state/predictions/helpers.ts
+++ b/src/state/predictions/helpers.ts
@@ -141,6 +141,21 @@ export const makeRoundData = (rounds: Round[]): RoundData => {
 }
 
 /**
+ * Given a bet object, check if it is eligible to be claimed or refunded
+ */
+export const getCanClaim = (bet: Bet) => {
+  return !bet.claimed && (bet.position === bet.round.position || bet.round.failed === true)
+}
+
+/**
+ * Returns only bets where the user has won.
+ * This is necessary because the API currently cannot distinguish between an uncliamed bet that has won or lost
+ */
+export const getUnclaimedWinningBets = (bets: Bet[]): Bet[] => {
+  return bets.filter(getCanClaim)
+}
+
+/**
  * Gets static data from the contract
  */
 export const getStaticPredictionsData = async () => {
@@ -211,7 +226,7 @@ export const getRound = async (id: string) => {
   return response.round
 }
 
-type BetHistoryWhereClause = Record<string, string | number | boolean>
+type BetHistoryWhereClause = Record<string, string | number | boolean | string[]>
 
 export const getBetHistory = async (
   where: BetHistoryWhereClause = {},
--- a/src/state/predictions/index.ts
+++ b/src/state/predictions/index.ts
@@ -1,6 +1,7 @@
 /* eslint-disable no-param-reassign */
 import { createAsyncThunk, createSlice, PayloadAction } from '@reduxjs/toolkit'
-import { maxBy } from 'lodash'
+import maxBy from 'lodash/maxBy'
+import merge from 'lodash/merge'
 import { BIG_ZERO } from 'utils/bigNumber'
 import { Bet, HistoryFilter, Market, PredictionsState, PredictionStatus, Round } from 'state/types'
 import {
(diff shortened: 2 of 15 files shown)


src/state/predictions/helpers.ts
```diff
@@ -140,6 +140,21 @@ export const makeRoundData = (rounds: Round[]): RoundData => {
...
+  return !bet.claimed && (bet.position === bet.round.position || bet.round.failed === true)
+}
+
+/**
+ * Returns only bets where the user has won.
+ * This is necessary because the API currently cannot distinguish between an uncliamed bet that has won or lost
+ */
+export const getUnclaimedWinningBets = (bets: Bet[]): Bet[] => {
+  return bets.filter(getCanClaim)
+}
```
hachiojidev, line 156: This was used in more than one spot so put it here as a helper.

src/state/predictions/index.ts
```diff
@@ -59,6 +60,21 @@ export const fetchRoundBet = createAsyncThunk<
...
+/**
+ * Used to poll the user bets of the current round cards
+ */
+export const fetchCurrentBets = createAsyncThunk<
+  { account: string; bets: Bet[] },
+  { account: string; roundIds: string[] }
+>('predictions/fetchCurrentBets', async ({ account, roundIds }) => {
+  const betResponses = await getBetHistory({
+    user: account.toLowerCase(),
+    round_in: roundIds,
```
hachiojidev, line 72: I realized that we can query the current round's bets more efficiently by using the `round_in` clause. Before each card was querying its bet individually.

src/state/predictions/index.ts
```diff
@@ -184,6 +208,18 @@ export const predictionsSlice = createSlice({
...
+      const betData = bets.reduce((accum, bet) => {
+        return {
+          ...accum,
+          [bet.round.id]: bet,
+        }
+      }, {})
+
+      state.bets = merge({}, state.bets, {
+        [account]: betData,
+      })
```
hachiojidev, line 222: Every time we fetch the bets to display in the pane we can save the results in the `bets` object.

src/state/types.ts
```diff
@@ -195,12 +195,12 @@ export interface Market {
...
 export interface Bet {
-  id: string
-  hash: string
+  id?: string
+  hash?: string
   amount: number
   position: BetPosition
   claimed: boolean
-  user: PredictionUser
+  user?: PredictionUser
```
hachiojidev, line 203: This is in place of using `Partial<Bet>` which was causing all sorts of problems.

src/views/Predictions/hooks/useOnNextRound.ts
```diff
@@ -15,11 +21,14 @@ const useOnNextRound = () => {
     if (swiper && currentEpoch !== undefined && previousEpoch !== undefined && currentEpoch !== previousEpoch) {
       const currentEpochIndex = rounds.findIndex((round) => round.epoch === currentEpoch)
 
+      // Fetch data on current unclaimed bets
+      dispatch(fetchCurrentBets({ account, roundIds: rounds.map((round) => round.id) }))
+
```
hachiojidev, line 26: When the round changes we go and get the bets displayed again to keep it in sync.

src/views/Predictions/index.tsx
```diff
@@ -64,6 +66,9 @@ const Predictions = () => {
       const { currentEpoch, intervalBlocks, bufferBlocks } = staticPredictionsData
       const latestRound = marketData.rounds.find((round) => round.epoch === currentEpoch)
 
+      // Fetch data on current unclaimed bets
+      dispatch(fetchCurrentBets({ account, roundIds: marketData.rounds.map((round) => round.id) }))
+
```
hachiojidev, line 71: This is the replacement query on load.

RabbitDoge (approved):

hachiojidev:
> @hachiojidev Should we also refactor CollectWinningsPopup?

To do what? This change updates the bet's claimable status in the global store.

memoyil:
> > @hachiojidev Should we also refactor CollectWinningsPopup?
> 
> To do what? This change updates the bet's claimable status in the global store.

It takes bets directly from graphql every 30 sec. Should we point to global store or keep it as it is?

hachiojidev:
@memoyil I see what you mean. Yeah, we should still query the API but simply have the API update the store. Will tackle that in another PR.

PR #1171, opened 2021-05-10
refactor(predictions): Add resizable panes

hachiojidev:


package.json
```diff
@@ -63,6 +63,7 @@
     "react-router-dom": "^5.2.0",
     "react-scripts": "^4.0.1",
     "react-transition-group": "^4.4.1",
+    "split-grid": "^1.0.11",
```
RabbitDoge, line 66: Do we really need a deps for this ???

hachiojidev: Absolutely.

RabbitDoge (approved):

PR #1179, opened 2021-05-11
fix(headings): Fix smol headings

ChefHutch:
Change heading size prop to scale throughout app, following a recent uikit change

### DIFF
--- a/src/views/Profile/components/ClaimPointsCallout .tsx
+++ b/src/views/Profile/components/ClaimPointsCallout .tsx
@@ -54,7 +54,9 @@ const ClaimPointsCallout = () => {
         <Flex flexDirection={['column', null, 'row']} justifyContent={['start', null, 'space-between']}>
           <Flex alignItems="center" mb={['16px', null, 0]}>
             <PrizeIcon width="32px" mr="8px" />
-            <Heading size="lg">{t(`${totalPointsToCollect} Points to Collect`, { num: totalPointsToCollect })}</Heading>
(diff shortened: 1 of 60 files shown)


src/views/Profile/components/ClaimPointsCallout .tsx
```diff
@@ -54,7 +54,9 @@ const ClaimPointsCallout = () => {
         <Flex flexDirection={['column', null, 'row']} justifyContent={['start', null, 'space-between']}>
           <Flex alignItems="center" mb={['16px', null, 0]}>
             <PrizeIcon width="32px" mr="8px" />
-            <Heading size="lg">{t(`${totalPointsToCollect} Points to Collect`, { num: totalPointsToCollect })}</Heading>
+            <Heading scale="lg">
+              {t(`${totalPointsToCollect} Points to Collect`, { num: totalPointsToCollect })}
```
hachiojidev, line 58: This won't work. Needs to be
```
{t(`%num% Points to Collect`, { num: totalPointsToCollect })}
```

hachiojidev: Please add the key to `translations.json` as well.

hachiojidev (changes requested):

src/views/Profile/components/ClaimPointsCallout .tsx
ChefHutch: Nice spot.

PR #1183, opened 2021-05-11
fix: (pools) Zero with decimals shown in amount when bunny slider moved back to zero

memoyil:
Before: 

Screenshot 2021-05-11 at 12 53 37

After: 

Screenshot 2021-05-11 at 13 36 56


ChefHutch (approved):
Thanks for taking a look at this Memoyil. lgtm

hachiojidev (approved):

PR #1190, opened 2021-05-12
Feature/staking limits

hachiojidev:
Adds check on staking limits for pools

### DIFF
--- a/src/config/constants/pools.ts
+++ b/src/config/constants/pools.ts
@@ -1,5 +1,3 @@
-import BigNumber from 'bignumber.js'
-import { getDecimalAmount } from '../../utils/formatBalance'
 import tokens from './tokens'
 import { PoolConfig, PoolCategory } from './types'
 
@@ -1106,7 +1104,6 @@ const pools: PoolConfig[] = [
     sousId: 62,
     stakingToken: tokens.tpt,
     earningToken: tokens.cake,
-    stakingLimit: getDecimalAmount(new BigNumber(25000)),
     contractAddress: {
       97: '',
       56: '0x0c3D6892aa3b23811Af3bd1bbeA8b0740E8e4528',
@@ -1219,7 +1216,6 @@ const pools: PoolConfig[] = [
     sousId: 54,
     stakingToken: tokens.sushi,
     earningToken: tokens.cake,
-    stakingLimit: getDecimalAmount(new BigNumber(50)),
     contractAddress: {
       97: '',
       56: '0xb69b6e390cba1F68442A886bC89E955048DAe7E3',
@@ -1304,7 +1300,6 @@ const pools: PoolConfig[] = [
     sousId: 48,
     stakingToken: tokens.comp,
     earningToken: tokens.cake,
-    stakingLimit: getDecimalAmount(new BigNumber(1)),
     contractAddress: {
       97: '',
       56: '0xE0565fBb109A3f3f8097D8A9D931277bfd795072',
@@ -1319,7 +1314,6 @@ const pools: PoolConfig[] = [
     sousId: 47,
     stakingToken: tokens.comp,
     earningToken: tokens.cake,
-    stakingLimit: getDecimalAmount(new BigNumber(1)),
     contractAddress: {
       97: '',
       56: '0xc3693e3cbc3514d5d07EA5b27A721F184F617900',
@@ -1614,7 +1608,6 @@ const pools: PoolConfig[] = [
     sousId: 26,
     stakingToken: tokens.twt,
     earningToken: tokens.cake,
-    stakingLimit: getDecimalAmount(new BigNumber(1000)),
     contractAddress: {
       97: '',
       56: '0x0554a5D083Abf2f056ae3F6029e1714B9A655174',
@@ -1671,7 +1664,6 @@ const pools: PoolConfig[] = [
     sousId: 20,
     stakingToken: tokens.bnb,
     earningToken: tokens.cake,
-    stakingLimit: getDecimalAmount(new BigNumber(10)),
     contractAddress: {
       97: '',
       56: '0x555Ea72d7347E82C614C16f005fA91cAf06DCB5a',
--- a/src/config/constants/pools.ts
+++ b/src/config/constants/pools.ts
@@ -1,3 +1,5 @@
+import BigNumber from 'bignumber.js'
+import { getDecimalAmount } from '../../utils/formatBalance'
 import tokens from './tokens'
 import { PoolConfig, PoolCategory } from './types'
 
@@ -1104,7 +1106,7 @@ const pools: PoolConfig[] = [
     sousId: 62,
     stakingToken: tokens.tpt,
     earningToken: tokens.cake,
-    stakingLimit: 25000,
+    stakingLimit: getDecimalAmount(new BigNumber(25000)),
     contractAddress: {
       97: '',
       56: '0x0c3D6892aa3b23811Af3bd1bbeA8b0740E8e4528',
@@ -1217,7 +1219,7 @@ const pools: PoolConfig[] = [
     sousId: 54,
     stakingToken: tokens.sushi,
     earningToken: tokens.cake,
-    stakingLimit: 50,
+    stakingLimit: getDecimalAmount(new BigNumber(50)),
     contractAddress: {
       97: '',
       56: '0xb69b6e390cba1F68442A886bC89E955048DAe7E3',
@@ -1302,7 +1304,7 @@ const pools: PoolConfig[] = [
     sousId: 48,
     stakingToken: tokens.comp,
     earningToken: tokens.cake,
-    stakingLimit: 1,
+    stakingLimit: getDecimalAmount(new BigNumber(1)),
     contractAddress: {
       97: '',
       56: '0xE0565fBb109A3f3f8097D8A9D931277bfd795072',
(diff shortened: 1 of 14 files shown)


src/config/constants/types.ts
```diff
@@ -68,7 +68,6 @@ export interface PoolConfig {
   sousId: number
   earningToken: Token
   stakingToken: Token
-  stakingLimit?: number
```
hachiojidev, line 71: Not used (nor should be)

src/state/pools/fetchPools.ts
```diff
@@ -69,3 +71,29 @@ export const fetchPoolsTotalStaking = async () => {
...
+
+export const fetchPoolStakingLimit = async (sousId: number): Promise<BigNumber> => {
+  try {
+    const sousContract = getSouschefV2Contract(sousId)
+    const stakingLimit = await sousContract.methods.poolLimitPerUser().call()
+    return new BigNumber(stakingLimit)
+  } catch (error) {
+    return BIG_ZERO
+  }
+}
```
hachiojidev, line 83: The `catch` here is important because for V1 pools

src/state/pools/helpers.ts
```diff
@@ -0,0 +1,24 @@
...
+export const transformPool = (pool: Pool): Pool => {
+  const { totalStaked, stakingLimit, userData, ...rest } = pool
+
+  return {
+    ...rest,
+    userData: transformUserData(userData),
+    totalStaked: new BigNumber(totalStaked),
+    stakingLimit: new BigNumber(stakingLimit),
+  } as Pool
+}
```
hachiojidev, line 24: While developing I noticed that pool data was incorrectly typed. They were typed as `BigNumber` but were actually coming back as strings. Nothing happened because we were casting the value as a `BigNumber` each time and/or we were not using the values a big numbers.

src/state/pools/index.ts
```diff
@@ -33,7 +34,10 @@ export const PoolsSlice = createSlice({
     updatePoolsUserData: (state, action) => {
       const { field, value, sousId } = action.payload
       const index = state.data.findIndex((p) => p.sousId === sousId)
-      state.data[index] = { ...state.data[index], userData: { ...state.data[index].userData, [field]: value } }
+
+      if (index >= 0) {
```
hachiojidev, line 38: This was not causing an issue or anything in production but I noticed with test data (where the sousId was mismatched) this threw an error so I added this to be safe.

src/utils/formatBalance.ts
```diff
@@ -20,7 +20,7 @@ export const getBalanceNumber = (balance: BigNumber, decimals = 18) => {
 }
 
 export const getFullDisplayBalance = (balance: BigNumber, decimals = 18, decimalsToAppear?: number) => {
-  return balance.dividedBy(BIG_TEN.pow(decimals)).toFixed(decimalsToAppear)
+  return getBalanceAmount(balance, decimals).toFixed(decimalsToAppear)
```
hachiojidev, line 23: Just reducing duplicate code.

src/views/Pools/components/PoolCard/CardActions/StakeActions.tsx
```diff
@@ -32,17 +32,16 @@ const StakeAction: React.FC<StakeActionsProps> = ({
   isStaked,
   isLoading = false,
 }) => {
-  const { stakingToken, earningToken, stakingLimit, isFinished } = pool
+  const { stakingToken, earningToken, stakingLimit, isFinished, userData } = pool
   const { t } = useTranslation()
-  const convertedLimit = getDecimalAmount(new BigNumber(stakingLimit), earningToken.decimals)
+  const convertedLimit = getDecimalAmount(stakingLimit, earningToken.decimals)
```
hachiojidev, line 37: Here is an example of what I mentioned earlier. `stakingLimit` was already typed as a BigNumber so we shouldn't need to cast it again `new BigNumber(stakingLimit)`.

src/views/Pools/components/PoolCard/CardActions/StakeActions.tsx
```diff
@@ -81,14 +85,23 @@ const StakeAction: React.FC<StakeActionsProps> = ({
...
           </IconButton>
-          <IconButton
-            variant="secondary"
-            onClick={stakingTokenBalance.gt(0) ? onPresentStake : onPresentTokenRequired}
-            disabled={isFinished}
-          >
-            <AddIcon color="primary" width="24px" height="24px" />
-          </IconButton>
+          {userData.stakedBalance.gte(stakingLimit) ? (
+            <span ref={targetRef}>
```
hachiojidev, line 89: Wrapped in `<span />` because the button is disabled (so tooltip won't trigger)

src/views/Pools/components/PoolCard/Modals/StakeModal.tsx
```diff
@@ -33,20 +33,28 @@ const StakeModal: React.FC<StakeModalProps> = ({
...
 
+  useEffect(() => {
+    const stakingLimitBn = new BigNumber(stakingLimit)
+
+    if (stakingLimitBn.gt(0) && !isRemovingStake) {
+      const fullDecimalStakeAmount = getDecimalAmount(new BigNumber(stakeAmount), stakingToken.decimals)
+      setHasReachedStakedLimit(fullDecimalStakeAmount.plus(userData.stakedBalance).gt(stakingLimitBn))
+    }
+  }, [stakeAmount, stakingLimit, userData, stakingToken, isRemovingStake, setHasReachedStakedLimit])
+
```
hachiojidev, line 57: This is the effect to check if the user can add the amount given the limit. We need to add their current stake to what the input is.

src/views/Pools/components/PoolCard/Modals/StakeModal.tsx
```diff
@@ -33,20 +33,28 @@ const StakeModal: React.FC<StakeModalProps> = ({
...
-
   const [pendingTx, setPendingTx] = useState(false)
   const [stakeAmount, setStakeAmount] = useState('')
+  const [hasReachedStakeLimit, setHasReachedStakedLimit] = useState(false)
   const [percent, setPercent] = useState(0)
 
   const usdValueStaked = stakeAmount && formatNumber(new BigNumber(stakeAmount).times(stakingTokenPrice).toNumber())
 
+  useEffect(() => {
+    const stakingLimitBn = new BigNumber(stakingLimit)
```
memoyil, line 50: Isnt the stakingLimit bigNumber already?

hachiojidev: Nice catch!

src/config/constants/pools.ts
```diff
@@ -1076,7 +1078,7 @@ const pools: PoolConfig[] = [
     sousId: 62,
     stakingToken: tokens.tpt,
     earningToken: tokens.cake,
-    stakingLimit: 25000,
+    stakingLimit: getDecimalAmount(new BigNumber(25000)),
```
hachiojidev, line 1081: This is to bring the old staking limits in line with the new format.

src/config/constants/pools.ts
```diff
@@ -1076,7 +1076,6 @@ const pools: PoolConfig[] = [
     sousId: 62,
     stakingToken: tokens.tpt,
     earningToken: tokens.cake,
-    stakingLimit: 25000,
```
hachiojidev, line 1079: Confirmed with Chungus that these are longer relevant.

src/state/pools/fetchPools.ts
```diff
@@ -69,3 +71,29 @@ export const fetchPoolsTotalStaking = async () => {
...
+
+export const fetchPoolsStakingLimits = async (): Promise<{ [key: string]: BigNumber }> => {
+  const validPools = poolsConfig.filter((p) => p.stakingToken.symbol !== 'BNB' && !p.isFinished)
+
+  // Get the staking limit for each valid pool
+  // Note: We cannot batch the calls via multicall because V1 pools do not have "poolLimitPerUser" and will throw an error
+  const stakingLimitPromises = validPools.map((validPool) => fetchPoolStakingLimit(validPool.sousId))
```
ChefHutch, line 90: So are we attempting to fetch a staking limit for every pool apart from BNB staking pools & finished pools?

Are these staking limits going to become ubiquitous, or will it only be for specific pools? If it's the latter... maybe we should assign a boolean to these `hasStakingLimit` pools within the pools config to help filter the number of calls we need to make here? Might it also then allow us to do a multicall as V1 pools' `hasStakingLimit` would be false/undefined?

Doing anything for _every_ pool gives me the fear because what we do once, we do 60+ times. How's the performance with 60+ calls?

hachiojidev: Yeah...I agree. I am hoping we move a more efficient design where we don't fetch everything at once.

> Might it also then allow us to do a multicall as V1 pools' hasStakingLimit would be false/undefined?
This might be the solution we have to take.

src/state/pools/fetchPools.ts
RabbitDoge: this should be a comment in the code

RabbitDoge (approved):

Chef-Cheems (approved):

src/views/Pools/components/PoolCard/CardActions/StakeActions.tsx
```diff
@@ -32,33 +32,41 @@ const StakeAction: React.FC<StakeActionsProps> = ({
...
+      isBnbPool={isBnbPool}
+      pool={pool}
+      stakingTokenBalance={stakingTokenBalance}
+      stakingTokenPrice={stakingTokenPrice}
+    />,
   )
 
   const [onPresentUnstake] = useModal(
     <StakeModal
-      stakingMax={stakedBalance}
```
ChefHutch, line 54: Ah fuck. This is the crucial bit ahaha. `stakingMax` was different depending on whether it was a staking/unstaking action.

PR #1199, opened 2021-05-13
fix(stake-modal): Catch for zero staking limit & handling for unstaking

ChefHutch:
Two issues in live
- STAKING: Pools with a `stakingLimit` of `BIG_ZERO` are returning the stakingLimit value (0), not the user token balance (`stakingTokenBalance`)
- UNSTAKING: There's no specific handling for determining how much a user can unstake in the FE. Currently - it uses the same as the amount that you can stake.

Preview - https://pools-fixes-13-5.netlify.app/

### DIFF
--- a/src/views/Pools/components/PoolCard/Modals/StakeModal.tsx
+++ b/src/views/Pools/components/PoolCard/Modals/StakeModal.tsx
@@ -100,7 +100,6 @@ const StakeModal: React.FC<StakeModalProps> = ({
       }
     } else {
       try {
-        debugger // eslint-disable-line
         // staking
         await onStake(stakeAmount, stakingToken.decimals)
         toastSuccess(`${t('Staked')}!`, t(`Your ${stakingToken.symbol} funds have been staked in the pool!`))
--- a/src/views/Pools/components/PoolCard/Modals/StakeModal.tsx
+++ b/src/views/Pools/components/PoolCard/Modals/StakeModal.tsx
@@ -43,7 +43,13 @@ const StakeModal: React.FC<StakeModalProps> = ({
   const [stakeAmount, setStakeAmount] = useState('')
   const [hasReachedStakeLimit, setHasReachedStakedLimit] = useState(false)
   const [percent, setPercent] = useState(0)
-  const calculatedStakingLimit = stakingTokenBalance.gt(stakingLimit) ? stakingLimit : stakingTokenBalance
+  const getCalculatedStakingLimit = () => {
+    if (isRemovingStake) {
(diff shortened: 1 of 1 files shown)


src/views/Pools/components/PoolCard/Modals/StakeModal.tsx
```diff
@@ -43,7 +43,13 @@ const StakeModal: React.FC<StakeModalProps> = ({
   const [stakeAmount, setStakeAmount] = useState('')
   const [hasReachedStakeLimit, setHasReachedStakedLimit] = useState(false)
   const [percent, setPercent] = useState(0)
-  const calculatedStakingLimit = stakingTokenBalance.gt(stakingLimit) ? stakingLimit : stakingTokenBalance
+  const getCalculatedStakingLimit = () => {
+    if (isRemovingStake) {
+      return userData.stakedBalance
+    }
+    return stakingTokenBalance.gt(stakingLimit) && stakingLimit.gt(0) ? stakingLimit : stakingTokenBalance
```
ChefHutch, line 50: Added `&& stakingLimit.gt(0)` - to catch the majority of pools which will have a stakingLimit of 0

src/views/Pools/components/PoolCard/Modals/StakeModal.tsx
```diff
@@ -43,7 +43,13 @@ const StakeModal: React.FC<StakeModalProps> = ({
   const [stakeAmount, setStakeAmount] = useState('')
   const [hasReachedStakeLimit, setHasReachedStakedLimit] = useState(false)
   const [percent, setPercent] = useState(0)
-  const calculatedStakingLimit = stakingTokenBalance.gt(stakingLimit) ? stakingLimit : stakingTokenBalance
+  const getCalculatedStakingLimit = () => {
+    if (isRemovingStake) {
+      return userData.stakedBalance
```
ChefHutch, line 48: Currently there's no specific handling for unstaking from pools, this sets the limit to the total user's stakedBalance when unstaking

src/views/Pools/components/PoolCard/Modals/StakeModal.tsx
hachiojidev: `stakingLimit.gt(0)` should be the first clause so it bails out earlier. (micro optimization I know)

ChefHutch: Tasty. Will do

hachiojidev (approved):

PR #1209, opened 2021-05-14
fix: Change ONE pool and farm logos

Chef-Cheems:


Chef-Chungus (approved):

Chef-Chungus:
@Chef-Cheems still need to replace the img in /tokens

Chef-Cheems:
@Chef-Chungus good catch, added.

PR #1211, opened 2021-05-14
feat: (pools) Add infinite scroll and lazy loading pool cards

memoyil:


Chef-Cheems (approved):

hachiojidev (approved):
Thanks for adding this. @Chef-Cheems is working on the overhaul of this page which includes lazy loading as well but we will merge this anyway because it is sorely needed.

PR #1216, opened 2021-05-16
refactor: Make code style better

luckyJeffy:
May make the code more readable 😆

### DIFF
--- a/src/views/TradingCompetition/index.tsx
+++ b/src/views/TradingCompetition/index.tsx
@@ -99,8 +99,11 @@ const TradingCompetition = () => {
   const [team3LeaderboardInformation, setTeam3LeaderboardInformation] = useState({ teamId: 3, leaderboardData: null })
 
   const isCompetitionLive = currentPhase.state === LIVE
-  const hasCompetitionEnded =
-    currentPhase.state === FINISHED || currentPhase.state === CLAIM || currentPhase.state === OVER
+  const hasCompetitionEnded = [
+    FINISHED,
+    CLAIM,
+    OVER
+  ].some((phasesItem) => phasesItem === currentPhase.state)
 
   const { hasUserClaimed, userCakeRewards, userPointReward, canClaimNFT } = userTradingInformation
 
@@ -201,8 +204,12 @@ const TradingCompetition = () => {
   }, [profileApiUrl])
 
   // Don't hide when loading. Hide if the account is connected && the user hasn't registered && the competition is live or finished
-  const shouldHideCta =
-    !isLoading && account && !userTradingInformation.hasRegistered && (isCompetitionLive || hasCompetitionEnded)
+  const shouldHideCta = [
+    !isLoading,
+    account,
+    !userTradingInformation.hasRegistered,
+    isCompetitionLive || hasCompetitionEnded
+  ].every(i => Boolean(i))
 
   return (
     <CompetitionPage>
--- a/src/views/TradingCompetition/index.tsx
+++ b/src/views/TradingCompetition/index.tsx
@@ -8,10 +8,9 @@ import useTheme from 'hooks/useTheme'
 import {
   SmartContractPhases,
   CompetitionPhases,
+  EndedPhasesSet,
   LIVE,
-  FINISHED,
   CLAIM,
-  OVER,
   REGISTRATION,
 } from 'config/constants/trading-competition/easterPhases'
 import { PrizesIcon, RanksIcon, RulesIcon } from './svgs'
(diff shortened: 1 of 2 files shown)


src/views/TradingCompetition/index.tsx
```diff
@@ -99,8 +99,11 @@ const TradingCompetition = () => {
   const [team3LeaderboardInformation, setTeam3LeaderboardInformation] = useState({ teamId: 3, leaderboardData: null })
 
   const isCompetitionLive = currentPhase.state === LIVE
-  const hasCompetitionEnded =
-    currentPhase.state === FINISHED || currentPhase.state === CLAIM || currentPhase.state === OVER
+  const hasCompetitionEnded = [
+    FINISHED,
+    CLAIM,
+    OVER
+  ].some((phasesItem) => phasesItem === currentPhase.state)
```
hachiojidev, line 106: In this case I think it is better to use `includes`
```
const hasCompetitionEnded = [
  FINISHED,
  CLAIM,
  OVER
].includes(currentPhase.state)
```

src/views/TradingCompetition/index.tsx
```diff
@@ -201,8 +204,12 @@ const TradingCompetition = () => {
...
 
   // Don't hide when loading. Hide if the account is connected && the user hasn't registered && the competition is live or finished
-  const shouldHideCta =
-    !isLoading && account && !userTradingInformation.hasRegistered && (isCompetitionLive || hasCompetitionEnded)
+  const shouldHideCta = [
+    !isLoading,
+    account,
+    !userTradingInformation.hasRegistered,
+    isCompetitionLive || hasCompetitionEnded
+  ].every(i => Boolean(i))
```
hachiojidev, line 212: I am not convinced this is more _readable_. I would opt for an inline function.
```
const shouldHideCta = () => {
  if (!isLoading) {
    return false
  }  

// Guard for each item  
...etc
  
  return true
}
```

hachiojidev (changes requested):

src/views/TradingCompetition/index.tsx
luckyJeffy: Yeah, It was much better 👍

src/views/TradingCompetition/index.tsx
luckyJeffy: I got it, this case can be more controllable.
But now the constance "shouldHideCta" type will become to Function.

src/views/TradingCompetition/index.tsx
luckyJeffy: `
export const EndedPhasesSet = new Set([
  FINISHED,
  CLAIM,
  OVER
])
`

In this case, Set may better than Array.
So I just try modify the constant file for a new set, then import by "TradingCompetition" to use it.

d68109852c96cc46d309396f0e54c258c9274505

src/views/TradingCompetition/index.tsx
luckyJeffy: Updated.

d68109852c96cc46d309396f0e54c258c9274505

src/views/TradingCompetition/index.tsx
```diff
@@ -99,9 +98,7 @@ const TradingCompetition = () => {
   const [team3LeaderboardInformation, setTeam3LeaderboardInformation] = useState({ teamId: 3, leaderboardData: null })
 
   const isCompetitionLive = currentPhase.state === LIVE
-  const hasCompetitionEnded =
-    currentPhase.state === FINISHED || currentPhase.state === CLAIM || currentPhase.state === OVER
-
+  const hasCompetitionEnded = EndedPhasesSet.has(currentPhase.state);
```
hachiojidev, line 101: Nice. Have not considered this pattern but it is definitely more readable.

hachiojidev (approved):

src/views/TradingCompetition/index.tsx
luckyJeffy: 🎉

luckyJeffy:
`
Run yarn format:check
yarn run v1.22.10
$ prettier --check --loglevel error 'src/**/*.{js,jsx,ts,tsx}'
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
Error: Process completed with exit code 1.
`

Seam like something went wrong, I got same issue here when I try to commit the code.

Chef-Chungus:
@luckyJeffy - Try doing `yarn install` then running prettier on the files again.

hachiojidev:
@luckyJeffy 

>Seam like something went wrong, I got same issue here when I try to commit the code.

run `yarn format:write`

Likely there is a file that was not formatted. If it is a file you have not touched try rebasing develop.

(site note: need to figure out why prettier is not reporting the warnings in the console)

Chef-Chungus:
@luckyJeffy We will close this PR in one day if there's no update from you

luckyJeffy:
> @luckyJeffy We will close this PR in one day if there's no update from you

Sorry guys, I am going through a very busy stage (Maybe two weeks), I will continue to update when I take a little break

PR #1226, opened 2021-05-17
feat(cake-api-price): Replace API price with Farms price for CAKE

ChefHutch:
PR is doing a couple of things 
- Replace any instances where `useGetApiPrice` is being used to access the price of CAKE, and replace it with `usePriceCakeBusd`
- Clear timer for `useWithdrawalFeeTimer` when `doesUnstakingFeeApply` is false, to prevent unnecessary re-renders
- useMemo to wrap the returns of all BigNumbers within the `useCakeVault`. If one of the return values was used as a useEffect/Memo dependency, then it would trigger a re-render as it is a new class instance with a different reference

### DIFF
--- a/src/hooks/cakeVault/useWithdrawalFeeTimer.ts
+++ b/src/hooks/cakeVault/useWithdrawalFeeTimer.ts
@@ -9,14 +9,17 @@ const useWithdrawalFeeTimer = (lastDepositedTime: number, withdrawalFeePeriod = 
     const feeEndTime = lastDepositedTime + withdrawalFeePeriod
     const secondsRemainingCalc = feeEndTime - currentSeconds
     const doesUnstakingFeeApply = secondsRemainingCalc > 0
-    if (doesUnstakingFeeApply) {
-      setSecondsRemaining(secondsRemainingCalc)
-      setHasUnstakingFee(true)
-    }
     const tick = () => {
       setCurrentSeconds((prevSeconds) => prevSeconds + 1)
     }
     const timerInterval = setInterval(() => tick(), 1000)
+    if (doesUnstakingFeeApply) {
+      setSecondsRemaining(secondsRemainingCalc)
+      setHasUnstakingFee(true)
+    } else {
+      setHasUnstakingFee(false)
+      clearInterval(timerInterval)
+    }
     return () => clearInterval(timerInterval)
   }, [lastDepositedTime, withdrawalFeePeriod, setSecondsRemaining, currentSeconds])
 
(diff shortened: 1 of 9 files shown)


src/hooks/cakeVault/useWithdrawalFeeTimer.ts
```diff
@@ -9,14 +9,17 @@ const useWithdrawalFeeTimer = (lastDepositedTime: number, withdrawalFeePeriod =
...
     const tick = () => {
       setCurrentSeconds((prevSeconds) => prevSeconds + 1)
     }
     const timerInterval = setInterval(() => tick(), 1000)
+    if (doesUnstakingFeeApply) {
+      setSecondsRemaining(secondsRemainingCalc)
+      setHasUnstakingFee(true)
+    } else {
+      setHasUnstakingFee(false)
+      clearInterval(timerInterval)
```
ChefHutch, line 21: Stop timer when it's no longer used to prevent re-renders each second

src/state/hooks.ts
```diff
@@ -142,17 +142,33 @@ export const useCakeVault = () => {
...
-  const pricePerFullShare = new BigNumber(pricePerFullShareAsString)
-  const totalPendingCakeHarvest = new BigNumber(totalPendingCakeHarvestAsString)
-  const totalCakeInVault = new BigNumber(totalCakeInVaultAsString)
-  const userShares = new BigNumber(userSharesAsString)
-  const cakeAtLastUserAction = new BigNumber(cakeAtLastUserActionAsString)
+  const totalPendingCakeHarvest = useMemo(() => {
+    return new BigNumber(totalPendingCakeHarvestAsString)
+  }, [totalPendingCakeHarvestAsString])
+
+  const totalShares = useMemo(() => {
```
ChefHutch, line 153: Memoise so that if the BigNumbers returned by `useCakeVault` are used in dependency arrays - they don't trigger re-renders

src/views/Pools/components/BountyCard.tsx
```diff
@@ -30,21 +29,17 @@ const StyledCard = styled(Card)`
...
 const BountyCard = () => {
   const { t } = useTranslation()
-  const [estimatedDollarBountyReward, setEstimatedDollarBountyReward] = useState(null)
   const {
     estimatedCakeBountyReward,
     totalPendingCakeHarvest,
     fees: { callFee },
   } = useCakeVault()
+  const cakePriceBusd = usePriceCakeBusd()
+  const cakePriceBusdAsNumber = cakePriceBusd.toNumber()
```
ChefHutch, line 38: The return of `usePriceCakeBusd` has a different reference on each render, even if the num value is the same. Calling `.toNumber()` to get the primitive num value makes it suitable for use in a deps array.

Chef-Cheems (approved):
Excellent 👍

hachiojidev (approved):

PR #1228, opened 2021-05-17
feat (prediction) use user language for tradingview

ciruz:
Noticed the language in the TradingView is hardcoded so I extended it, now the language from the user settings getting used in the TradingView.

image

image

If the language isn't available, the TradingView widget will be english by default (so no issues with incompatible codes).

Some language codes from PCS are not compatible with the TradingView widget, since their language codes are also a little inconsistent at some point (german is "de_DE", spanish is "es").

image

I'm not sure about this one, maybe one day someone extend the language interface ex. for a tradingview language, to be fully compatible:

image


src/views/Predictions/components/TradingView.tsx
```diff
@@ -38,19 +39,20 @@ const initializeTradingView = (TradingViewObj: any, theme: DefaultTheme) => {
 }
 
 const TradingView = () => {
+  const { currentLanguage } = useContext(LanguageContext);
```
hachiojidev, line 42: `const { currentLanguage } = useTranslation()`

hachiojidev (changes requested):
Nice work, just one thing to adjust.

Chef-Chungus:
@ciruz - Please run prettier on this one too

src/views/Predictions/components/TradingView.tsx
ciruz: thanks

ciruz:
done @Chef-Chungus

hachiojidev (approved):

PR #1235, opened 2021-05-18
Bugfix/ifo harvest

hachiojidev:


src/config/constants/ifo.ts
```diff
@@ -32,7 +32,7 @@ const ifos: Ifo[] = [
     campaignId: '511100000',
     articleUrl: 'https://pancakeswap.medium.com/hot-cross-hotcross-ifo-to-be-hosted-on-pancakeswap-10e70f1f6841',
     tokenOfferingPrice: 0.05,
-    isV1: false,
+    version: 2,
```
hachiojidev, line 35: Using `version` is more scalable in my opinion. Undoubtedly there will be a version 3+ as we grow.

hachiojidev:
This PR is waiting on a community member to verify it works.

Chef-Chungus (approved):

PR #1242, opened 2021-05-18
test: Added test to check new farms' factory address

hachiojidev:


src/__tests__/config/farms.test.ts
```diff
@@ -1,18 +1,21 @@
...
+import { Farm } from 'state/types'
 import { getBep20Contract, getLpContract } from 'utils/contractHelpers'
 
-const farmsToTest = farms.filter((farm) => farm.pid !== 0)
+const farmsToTest: [number, Farm][] = farms.filter((farm) => farm.pid !== 0).map((farm) => [farm.pid, farm])
 
 describe('Config farms', () => {
-  it.each(farms.map((farm) => farm.pid))('Farm #%d has an unique pid', (pid) => {
+  it.each(farmsToTest)('Farm #%d has an unique pid', (pid) => {
     const duplicates = farms.filter((f) => pid === f.pid)
```
ChefHutch, line 9: I'm kinda surprised this works... but I might be being dim.

Isn't `f` now an array like: `[number, Farm]`. So `f.pid` would be undefined?

ChefHutch: Being dim. `farms` is the const, `farmsToTest` is the array

hachiojidev: This is just so that when it prints in the console it just prints the `pid`

Chef-Chungus (approved):

PR #1244, opened 2021-05-18
Chore/Replace api prices with farm prices

ChefHutch:
Preview: https://farm-pools-prices.netlify.app/pools

### Replace API price data with LP price data for:

**Pools**
- APR / APY balances & calculation modals
- Staked/earned $ values
- Staking/unstaking modal $ values

**IFO**
- LP token price

-------

**To do in next PR:**
- Farms
- Possible optimisations of using the hooks to store price in state and then access them in unified way across farms & pools

### DIFF
--- a/src/state/hooks.ts
+++ b/src/state/hooks.ts
@@ -108,9 +108,6 @@ export const useTokenPriceBusd = (pid: number): BigNumber => {
   // we find the pBTC farm, pBTC - BNB
   // from the BNB - pBTC BUSD price, we can get the PNT - BUSD price
 
-  if (!quoteTokenFarm) {
-    debugger // eslint-disable-line
-  }
   if (quoteTokenFarm.quoteToken.symbol === 'wBNB') {
     const quoteTokenInBusd = bnbPriceBusd.gt(0) && bnbPriceBusd.times(quoteTokenFarm.tokenPriceVsQuote)
     return farm.tokenPriceVsQuote ? new BigNumber(farm.tokenPriceVsQuote).times(quoteTokenInBusd) : BIG_ZERO
--- a/src/state/hooks.ts
+++ b/src/state/hooks.ts
@@ -7,7 +7,6 @@ import { orderBy } from 'lodash'
 import { Team } from 'config/constants/types'
 import Nfts from 'config/constants/nfts'
 import { getWeb3NoAccount } from 'utils/web3'
-import { getAddress } from 'utils/addressHelpers'
 import { getBalanceNumber } from 'utils/formatBalance'
 import { BIG_ZERO } from 'utils/bigNumber'
 import useRefresh from 'hooks/useRefresh'
@@ -66,7 +65,7 @@ export const useFarmFromPid = (pid): Farm => {
   return farm
 }
 
-export const useFarmFromSymbol = (lpSymbol: string): Farm => {
+export const useFarmFromLpSymbol = (lpSymbol: string): Farm => {
   const farm = useSelector((state: State) => state.farms.data.find((f) => f.lpSymbol === lpSymbol))
   return farm
 }
@@ -82,13 +81,40 @@ export const useFarmUser = (pid) => {
   }
 }
 
-export const useLpTokenPrice = (symbol: string) => {
-  const farm = useFarmFromSymbol(symbol)
-  const tokenPriceInUsd = useGetApiPrice(getAddress(farm.token.address))
+export const useFarmFromTokenSymbol = (tokenSymbol: string): Farm => {
+  const farm = useSelector((state: State) => state.farms.data.find((f) => f.token.symbol === tokenSymbol))
+  return farm
+}
 
-  return farm.lpTotalSupply && farm.lpTotalInQuoteToken
-    ? new BigNumber(getBalanceNumber(farm.lpTotalSupply)).div(farm.lpTotalInQuoteToken).times(tokenPriceInUsd).times(2)
-    : BIG_ZERO
+export const useTokenPriceBusd = (pid: number): BigNumber => {
+  // POSSIBLE QUOTE TOKENS: wbnb, busd, qsd, ust, pbtc, btcb, eth,
+  // No route back to unit of account: qsd
+
+  let quoteTokenPriceBusd
+  let tokenPriceBusd
+  const farm = useFarmFromPid(pid)
+
+  if (farm.quoteToken.symbol === 'BUSD') {
+    tokenPriceBusd = farm.tokenPriceVsQuote ? new BigNumber(farm.tokenPriceVsQuote) : BIG_ZERO
+  } else if (farm.quoteToken.symbol === 'wBNB') {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    quoteTokenPriceBusd = usePriceBnbBusd()
+    tokenPriceBusd = quoteTokenPriceBusd.gt(0) && quoteTokenPriceBusd.times(farm.tokenPriceVsQuote)
+  } else {
+    debugger // eslint-disable-line
+  }
+
+  return tokenPriceBusd
+}
+
+export const useLpTokenPrice = (symbol: string) => {
+  const farm = useFarmFromLpSymbol(symbol)
+  const farmTokenPriceInUsd = useTokenPriceBusd(farm.pid)
+  const lpTokenPrice = new BigNumber(getBalanceNumber(farm.lpTotalSupply))
+    .div(farm.lpTotalInQuoteToken)
+    .times(farmTokenPriceInUsd)
+    .times(2)
+  return farm.lpTotalSupply && farm.lpTotalInQuoteToken ? lpTokenPrice : BIG_ZERO
 }
 
 // Pools
--- a/src/state/hooks.ts
+++ b/src/state/hooks.ts
@@ -91,7 +91,7 @@ export const useFarmFromTokenSymbol = (tokenSymbol: string, preferredQuoteTokens
 export const useTokenPriceBusd = (pid: number): BigNumber => {
   const farm = useFarmFromPid(pid)
   const bnbPriceBusd = usePriceBnbBusd()
-  const quoteTokenFarm = useFarmFromTokenSymbol(farm.token.symbol)
+  const quoteTokenFarm = useFarmFromTokenSymbol(farm.quoteToken.symbol)
 
   if (farm.quoteToken.symbol === 'BUSD') {
     return farm.tokenPriceVsQuote ? new BigNumber(farm.tokenPriceVsQuote) : BIG_ZERO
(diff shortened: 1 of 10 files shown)


src/state/hooks.ts
```diff
@@ -78,13 +78,54 @@ export const useFarmUser = (pid) => {
   }
 }
 
-export const useLpTokenPrice = (symbol: string) => {
-  const farm = useFarmFromSymbol(symbol)
-  const tokenPriceInUsd = useGetApiPrice(getAddress(farm.token.address))
+export const useFarmFromTokenSymbol = (tokenSymbol: string, preferredQuoteTokens?: string[]): Farm => {
+  const farms = useSelector((state: State) => state.farms.data.filter((f) => f.token.symbol === tokenSymbol))
+  const filteredFarm = filterFarmsByQuoteToken(farms, preferredQuoteTokens)
+  return filteredFarm
```
ChefHutch, line 84: Filter farm to get a 'preferred' LP, for example if we were to call the function with tokenSymbol `KUN`, two farms would match the filter:

```ts
  {
    pid: 373,
    lpSymbol: 'KUN-BUSD LP',
    token: tokens.kun,
    quoteToken: tokens.busd,
  },
  {
    pid: 372,
    lpSymbol: 'KUN-QSD LP',
    token: tokens.kun,
    quoteToken: tokens.qsd,
  },
```

We would want to return the `BUSD` farm to get the token price, as it is the most straightforward conversion (in fact in this instance, we'd have no way of getting price from `QSD` as there is no `QSD - someQuoteToken` pair).

The actual `filteredFarm` logic is within `'utils/farmsPriceHelpers'`

src/state/hooks.ts
```diff
@@ -78,13 +78,54 @@ export const useFarmUser = (pid) => {
...
+export const useTokenPriceBusd = (pid: number): BigNumber => {
+  const farm = useFarmFromPid(pid)
+  const bnbPriceBusd = usePriceBnbBusd()
+  const quoteTokenFarm = useFarmFromTokenSymbol(farm.token.symbol)
+
+  if (farm.quoteToken.symbol === 'BUSD') {
+    return farm.tokenPriceVsQuote ? new BigNumber(farm.tokenPriceVsQuote) : BIG_ZERO
+  }
+
+  if (farm.quoteToken.symbol === 'wBNB') {
```
ChefHutch, line 96: If the farm has a quoteToken of BUSD or wBNB, it is quite straightforward returning the price.

src/state/hooks.ts
```diff
@@ -78,13 +78,54 @@ export const useFarmUser = (pid) => {
...
-    ? new BigNumber(getBalanceNumber(farm.lpTotalSupply)).div(farm.lpTotalInQuoteToken).times(tokenPriceInUsd).times(2)
-    : BIG_ZERO
+  // Possible alternative farm quoteTokens:
+  // UST (MIR-UST), pBTC (PNT-pBTC), BTCB (bBADGER-BTCB), ETH (SUSHI-ETH)
+  // If the farm's quote token isn't BUSD or wBNB, we then use the quote token farm's quote token
+  // i.e. for farm PNT - pBTC
+  // we find the pBTC farm, pBTC - BNB
+  // from the BNB - pBTC BUSD price, we can get the PNT - BUSD price
+
+  if (quoteTokenFarm.quoteToken.symbol === 'wBNB') {
```
ChefHutch, line 107: If the farm doesn't have a quoteToken of BUSD or BNB, we access the quoteToken... of the quoteToken.

For example:
`bBADGER-BTCB`
Find a BTCB  BUSD or BNB farm
When we know the value of BTCB to BUSD or BNB, we can find the value of bBADGER to BUSD

src/utils/farmsPriceHelpers.ts
```diff
@@ -0,0 +1,12 @@
+import { Farm } from 'state/types'
+
+export const filterFarmsByQuoteToken = (farms: Farm[], preferredQuoteTokens: string[] = ['BUSD', 'wBNB']): Farm => {
+  const preferredFarm = farms.find((farm) => {
+    return preferredQuoteTokens.some((quoteToken) => {
+      return farm.quoteToken.symbol === quoteToken
+    })
+  })
+  return preferredFarm || farms[0]
```
ChefHutch, line 9: Farm filtering logic, to return the first BUSD or BNB farm if possible.

src/views/Pools/components/PoolCard/AprRow.tsx
```diff
@@ -35,25 +27,32 @@ const AprRow: React.FC<AprRowProps> = ({
...
   const { targetRef, tooltip, tooltipVisible } = useTooltip(tooltipContent, { placement: 'bottom-end' })
 
-  const earningTokenPrice = useGetApiPrice(earningToken.address ? getAddress(earningToken.address) : '')
+  const earningTokenFarmForPriceCalc = useFarmFromTokenSymbol(earningToken.symbol)
+  const earningTokenPrice = useTokenPriceBusd(earningTokenFarmForPriceCalc.pid)
+  const earningTokenPriceAsNumber = earningTokenPrice && earningTokenPrice.toNumber()
+
+  const stakingTokenFarmForPriceCalc = useFarmFromTokenSymbol(stakingToken.symbol)
+  const stakingTokenPrice = useTokenPriceBusd(stakingTokenFarmForPriceCalc.pid)
+  const stakingTokenPriceAsNumber = stakingTokenPrice && stakingTokenPrice.toNumber()
```
ChefHutch, line 36: This is a bit icky. I started this work by tackling the smallest pieces first, but I may have been better off tackling the biggest first (i.e. storing farm prices to state).

In the next PR I will be tackling the last of the API uses - the values stored to farms' prices state. 

I ought to then be able to tidy up price retrieval throughout the rest of the app, simplifying this sort of thing.

src/state/hooks.ts
```diff
@@ -78,13 +78,54 @@ export const useFarmUser = (pid) => {
...
+  if (quoteTokenFarm.quoteToken.symbol === 'BUSD') {
+    const quoteTokenInBusd = quoteTokenFarm.tokenPriceVsQuote
+    return quoteTokenInBusd ? new BigNumber(farm.tokenPriceVsQuote).times(quoteTokenInBusd) : BIG_ZERO
+  }
+
+  // Catch in case token does not have immediate or once-removed BUSD/wBNB quoteToken
+  return BIG_ZERO
+}
+
+export const useLpTokenPrice = (symbol: string) => {
```
ChefHutch, line 121: This is used by the ifo when displaying the LP token price. The diff is a bit confusing - it was already a function, I've just modified it.

src/views/Pools/components/PoolCard/CardActions/HarvestActions.tsx
```diff
@@ -25,9 +24,16 @@ const HarvestActions: React.FC<HarvestActionsProps> = ({
   isLoading = false,
 }) => {
   const { t } = useTranslation()
-  const earningTokenPrice = useGetApiPrice(earningToken.address ? getAddress(earningToken.address) : '')
+
+  const earningTokenFarmForPriceCalc = useFarmFromTokenSymbol(earningToken.symbol)
+  const earningTokenPrice = useTokenPriceBusd(earningTokenFarmForPriceCalc.pid)
```
ChefHutch, line 29: `MIX` pool janks out here because it doesn't have an `earningTokenFarmForPriceCalc`. Previously I've just avoided rendering the entire component when this stuff that the hook requires isn't present, but the MIX pool _does_ need some of the stuff in this component.

src/state/hooks.ts
```diff
@@ -78,13 +78,54 @@ export const useFarmUser = (pid) => {
   }
 }
 
-export const useLpTokenPrice = (symbol: string) => {
-  const farm = useFarmFromSymbol(symbol)
-  const tokenPriceInUsd = useGetApiPrice(getAddress(farm.token.address))
+export const useFarmFromTokenSymbol = (tokenSymbol: string, preferredQuoteTokens?: string[]): Farm => {
+  const farms = useSelector((state: State) => state.farms.data.filter((f) => f.token.symbol === tokenSymbol))
```
hachiojidev, line 82: Please use descriptive variable names `f `

src/state/hooks.ts
hachiojidev: Please add a comment to this effect (shortened of course).

src/state/hooks.ts
hachiojidev: Comment please.

src/utils/farmsPriceHelpers.ts
hachiojidev: Comment please.

src/views/Pools/components/PoolCard/CardActions/StakeActions.tsx
```diff
@@ -30,8 +29,12 @@ const StakeAction: React.FC<StakeActionsProps> = ({
   const { stakingToken, stakingLimit, isFinished, userData } = pool
   const { t } = useTranslation()
   const stakedTokenBalance = getBalanceNumber(stakedBalance, stakingToken.decimals)
+  const stakingTokenFarmForPriceCalc = useFarmFromTokenSymbol(stakingToken.symbol)
+  const stakingTokenPrice = useTokenPriceBusd(stakingTokenFarmForPriceCalc.pid)
+  const stakingTokenPriceAsNumber = stakingTokenPrice && stakingTokenPrice.toNumber()
+
   const stakedTokenDollarBalance = getBalanceNumber(
-    stakedBalance.multipliedBy(stakingTokenPrice),
+    stakedBalance.multipliedBy(stakingTokenPriceAsNumber),
```
hachiojidev, line 37: Any problem here if `stakingTokenPriceAsNumber` returns `false`?

Chef-Cheems (approved):

src/state/hooks.ts
```diff
@@ -82,13 +82,66 @@ export const useFarmUser = (pid) => {
   }
 }
 
-export const useLpTokenPrice = (symbol: string) => {
-  const farm = useFarmFromSymbol(symbol)
-  const tokenPriceInUsd = useGetApiPrice(getAddress(farm.token.address))
+// Return a farm for a given token symbol. The farm is filtered based on attempting to return a farm with a quote token from an array of preferred quote tokens
+export const useFarmFromTokenSymbol = (tokenSymbol: string, preferredQuoteTokens?: string[]): Farm => {
+  const farms = useSelector((state: State) => state.farms.data.filter((farm) => farm.token.symbol === tokenSymbol))
```
hachiojidev, line 87: Can you use `useFarmFromSymbol`?

ChefHutch: `useFarmFromSymbol` was just returning the first farm that it found in the constants, so the order in which the config is maintained becomes important. i.e. [with this example](https://github.com/pancakeswap/pancake-frontend/pull/1244#discussion_r634528092) if the `KUN-QSD` farm came before `KUN-BUSD` in farms.ts, it would try to retrieve the price based on the QSD pair, and it wouldn't be able to get a price.

The new filtering logic means it would use the `KUN-BUSD` farm, even if `KUN-QSD` came before it in the config.

hachiojidev (approved):

PR #1258, opened 2021-05-19
Feature/pools table

Chef-Cheems:
Pools table
Preview - https://awesome-raman-0813c6.netlify.app/


Chef-Cheems (commented):
Basically table consists of main table wrapper `<PoolsTable>` which has bunch of `<PoolRow>` inside, each row consist of cells present in `Cells` directory. The data calculation logic is pretty much the same as in cards with the exception that vault is not separated to its own "VaultRow" component but integrated into normal flow (i.e. there is bunch of `if (isAutoVault) ...` conditions here.

src/components/Balance.tsx
```diff
@@ -8,6 +8,7 @@ interface BalanceProps extends TextProps {
   unit?: string
   isDisabled?: boolean
   prefix?: string
+  onClick?: (event: React.MouseEvent<HTMLElement>) => void
```
hachiojidev, line 11: I assume this is necessary but it shouldn't be. `onClick` should be inherited from `TextProps`

src/state/pools/index.ts
```diff
@@ -15,6 +15,7 @@ import fetchVaultUser from './fetchVaultUser'
 
 const initialState: PoolsState = {
   data: [...poolsConfig],
+  userDataLoaded: false,
```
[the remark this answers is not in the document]
hachiojidev: Not a priority but we have used `isInitialized` throughout the app for this.

src/views/Pools/components/CakeVaultCard/UnstakingFeeCountdownRow.tsx
```diff
@@ -6,7 +6,11 @@ import useWithdrawalFeeTimer from 'hooks/cakeVault/useWithdrawalFeeTimer'
 import { useCakeVault } from 'state/hooks'
 import WithdrawalFeeTimer from './WithdrawalFeeTimer'
 
-const UnstakingFeeCountdownRow = () => {
+interface UnstakingFeeCountdownRowProps {
+  tableVariant?: boolean
```
hachiojidev, line 10: Please prefix this with `is` or `has`

src/views/Pools/components/HelpButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+const HelpButton = () => {
+  const { t } = useTranslation()
+  return (
+    <StyledLink external href="https://docs.pancakeswap.finance/syrup-pools/syrup-pool">
+      <Button px={['14px', null, null, null, '20px']} variant="subtle">
+        <ButtonText color="backgroundAlt" bold fontSize="16px">
+          {t('Help')}
+        </ButtonText>
+        <HelpIcon color="backgroundAlt" ml={[null, null, null, 0, '6px']} />
```
hachiojidev, line 29: You can't use `endIcon` here?

src/views/Pools/components/PoolTabButtons.tsx
```diff
@@ -1,66 +1,65 @@
...
-        </StyledLink>
-      </Flex>
+    )
+  }
+
+  return (
+    <Flex
+      alignItems="center"
+      justifyContent={['space-around', 'space-around', 'flex-start']}
+      mb={['24px', '24px', '24px', '0px']}
```
hachiojidev, line 58: Just an FYI, I see in other spots you are using `null`. Using `null` is preferable as it doesn't create a new style.

src/views/Pools/components/PoolsTable/ActionPanel/ActionPanel.tsx
```diff
@@ -0,0 +1,241 @@
...
+  }
+`
+
+type MediaBreakpoints = {
+  isXs: boolean
+  isSm: boolean
+  isMd: boolean
+  isLg: boolean
+  isXl: boolean
+}
```
hachiojidev, line 88: This is not exported from the toolkit? If not, then can you add a TODO here to do so.

src/views/Pools/components/PoolsTable/Apr.tsx
```diff
@@ -0,0 +1,120 @@
...
+        parseFloat(tokenPerBlock),
+      )
+    : 0
+
+  // special handling for tokens like tBTC or BIFI where the daily token rewards for $1000 dollars will be less than 0.001 of that token
+  const isHighValueToken = Math.round(earningTokenPriceAsNumber / 1000) > 0
+  const roundingDecimals = isHighValueToken ? 4 : 2
+
+  //   Estimate & manual for now. 288 = once every 5 mins. We can change once we have a better sense of this
+  const compoundFrequency = isAutoVault ? 288 : 1
```
hachiojidev, line 42: `288` should be in a config I think.

src/views/Pools/components/PoolsTable/Cells/EarningsCell.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+interface EarningsCellProps {
+  pool: Pool
+  account: string
+  userDataLoaded: boolean
+  isAutoVault?: boolean
+}
+
+const StyledCell = styled(BaseCell)`
+  flex: 4.5;
```
hachiojidev, line 24: Why 4.5?

src/views/Pools/components/PoolsTable/PoolsTable.tsx
```diff
@@ -0,0 +1,63 @@
...
+  cakeVault?: Pool
+  account: string
+  showFinishedPools: boolean
+}
+
+const StyledTable = styled.div`
+  box-shadow: ${({ theme }) =>
+    theme.isDark
+      ? `rgba(217, 217, 226, 0.05) 0px 1px 0px, rgba(217, 217, 226, 0.3) 0px 0px 8px;`
+      : `rgba(17, 17, 26, 0.05) 0px 1px 0px, rgba(17, 17, 26, 0.1) 0px 0px 8px;`}
```
hachiojidev, line 20: Do we need to add this to the UI Kit? If so please add a TODO

src/state/pools/index.ts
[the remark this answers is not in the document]
Chef-Cheems: Will check it out, haven't seen it before 👀

src/components/Balance.tsx
Chef-Cheems: I think I tried to pass it just like that but TS couldn't find it... will try again, maybe I did something wrong last time 🤔

src/views/Pools/components/PoolTabButtons.tsx
Chef-Cheems: Good to know 👍

src/views/Pools/components/PoolsTable/Cells/EarningsCell.tsx
Chef-Cheems: So there is bunch of weird flex ratios here and there, this is mostly fine-tuned in order to show table properly on all screen sizes. Whole numbers are nice but as you resize the window it looks awful at certain resolutions.
Even if there is solution for that that can use whole numbers it would probably require other complex stuff around it.
And non-integer flex values are supported in every browser as far as I know (maybe except IE11) so shouldn't cause any issues 🙂

src/views/Pools/components/PoolsTable/PoolsTable.tsx
Chef-Cheems: yeah those are not in UIKit, will confirm with Bugs exact shadows and put it there 👍

Chef-Cheems: actually talked with Bugs, we're ditching shadows to outlines, will rework this

src/views/Pools/components/PoolsTable/Apr.tsx
ChefHutch: Yea, and maybe it's time to increase it. We had compounding every block at first and it hasn't slowed down _that_ much. Worth seeing what @Chef-Chungus / Hops thinks about this. TBH the impact on APY between 288 and 28k daily compounds isn't that significant.

src/components/Balance.tsx
ChefHutch: I think if you didn't pass it at all within the `Balance` component Cheems, it may just work? 

`<Text>` accepts `{...props}` and then the `TextProps` should allow an onClick event?

src/views/Pools/components/PoolsTable/ActionPanel/ActionPanel.tsx
```diff
@@ -0,0 +1,241 @@
...
+    blocksRemaining || blocksUntilStart ? (
+      <Flex mb="8px" justifyContent="space-between">
+        <Text mr="4px">{t('Ends in')}:</Text>
+        <Flex>
+          <Link external href={`${BASE_BSC_SCAN_URL}/block/countdown/${endBlock}`}>
+            <Balance fontSize="16px" value={blocksToDisplay} decimals={0} color="primary" />
+            <Text ml="4px" color="primary">
+              {t('blocks')}
+            </Text>
+            <TimerIcon ml="4px" color="primary" />
```
ChefHutch, line 166: Some visual feedback on hover for this icon might be nice to indicate that it's clickable? I'm not sure the Block Balance & text is actually triggering the link? The timer icon is though!

src/views/Pools/components/PoolsTable/ActionPanel/ActionPanel.tsx
```diff
@@ -0,0 +1,241 @@
...
+  const tokenAddress = earningToken.address ? getAddress(earningToken.address) : ''
+  const imageSrc = `${BASE_URL}/images/tokens/${earningToken.symbol.toLowerCase()}.png`
+
+  const {
+    totalCakeInVault,
+    fees: { performanceFee },
+  } = useCakeVault()
+
+  const performanceFeeAsDecimal = performanceFee && performanceFee / 100
+  const isManualCakePool = sousId === 0
```
ChefHutch, line 136: Is this sort of thing (not all of it, but a lot of it) the kind of thing you were talking about in terms of having it in helpers, as it's used in a few places?

src/views/Pools/components/PoolsTable/ActionPanel/ActionPanel.tsx
```diff
@@ -0,0 +1,241 @@
...
+        </Flex>
+        {account && isMetaMaskInScope && tokenAddress && (
+          <Flex mb="8px" justifyContent={['flex-end', 'flex-end', 'flex-start']}>
+            <Button
+              variant="text"
+              p="0"
+              height="auto"
+              onClick={() => registerToken(tokenAddress, earningToken.symbol, earningToken.decimals, imageSrc)}
+            >
+              <Text color="primary" fontSize="14px">
```
ChefHutch, line 224: I think this should just be 16px?

src/views/Pools/components/PoolsTable/Apr.tsx
```diff
@@ -0,0 +1,120 @@
...
+        stakingTokenPriceAsNumber,
+        earningTokenPriceAsNumber,
+        getBalanceNumber(totalStaked, stakingToken.decimals),
+        parseFloat(tokenPerBlock),
+      )
+    : 0
+
+  // special handling for tokens like tBTC or BIFI where the daily token rewards for $1000 dollars will be less than 0.001 of that token
+  const isHighValueToken = Math.round(earningTokenPriceAsNumber / 1000) > 0
+  const roundingDecimals = isHighValueToken ? 4 : 2
```
ChefHutch, line 39: It seems a bit of a shame to have stuff like this repeated from the card `AprRow` component rather than creating something that can be shared? It adds a bit of maintenance to make these changes in multiple components going forward. It's pretty likely that these decimals will need tweaking in future.

src/views/Pools/components/PoolsTable/ActionPanel/Stake.tsx
```diff
@@ -0,0 +1,268 @@
...
+  }
+
+  const { targetRef, tooltip, tooltipVisible } = useTooltip(
+    t("You've already staked the maximum amount you can stake in this pool!"),
+    { placement: 'bottom' },
+  )
+
+  const reachStakingLimit = stakingLimit.gt(0) && userData.stakedBalance.gte(stakingLimit)
+
+  const handleApprove = useCallback(async () => {
```
ChefHutch, line 119: This should maybe be abstracted out of this component & shared with the card ui components rather than duplicating?

src/views/Pools/components/PoolsTable/ActionPanel/Harvest.tsx
```diff
@@ -0,0 +1,167 @@
...
+    pricePerFullShare,
+    fees: { performanceFee },
+  } = useCakeVault()
+  const hasAutoEarnings =
+    account && cakeAtLastUserAction && cakeAtLastUserAction.gt(0) && userShares && userShares.gt(0)
+  const { cakeAsBigNumber } = convertSharesToCake(userShares, pricePerFullShare)
+  const cakeProfit = cakeAsBigNumber.minus(cakeAtLastUserAction)
+  const cakeToDisplay = cakeProfit.gte(0) ? getBalanceNumber(cakeProfit, 18) : 0
+  const dollarValueOfCake = cakeProfit.times(earningTokenPrice)
+  const dollarValueToDisplay = dollarValueOfCake.gte(0) ? getBalanceNumber(dollarValueOfCake, 18) : 0
```
ChefHutch, line 58: Obviously we do a lot of this in the card too. I feel like all of my recent comments are on this theme - can we be smarter and share the logic between the table / card view where possible?

src/views/Pools/components/PoolsTable/Cells/EarningsCell.tsx
```diff
@@ -0,0 +1,154 @@
...
+    pricePerFullShare,
+  } = useCakeVault()
+  const hasAutoEarnings =
+    account && cakeAtLastUserAction && cakeAtLastUserAction.gt(0) && userShares && userShares.gt(0)
+  const { cakeAsBigNumber } = convertSharesToCake(userShares, pricePerFullShare)
+  const cakeProfit = cakeAsBigNumber.minus(cakeAtLastUserAction)
+  const cakeToDisplay = cakeProfit.gte(0) ? getBalanceNumber(cakeProfit, 18) : 0
+
+  const dollarValueOfCake = cakeProfit.times(earningTokenPriceAsNumber)
+  const dollarValueToDisplay = dollarValueOfCake.gte(0) ? getBalanceNumber(dollarValueOfCake, 18) : 0
```
ChefHutch, line 66: Is this the same as https://github.com/pancakeswap/pancake-frontend/pull/1258/files/b4bf66aee61da2b82999acd9b55da26ec710d8c0#diff-402e4cb9074d844667b90f43f90e042e9bc0d3fa8bdcd9bae57434dd5010373eR55-R58?

src/views/Pools/components/HelpButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+const ButtonText = styled(Text)`
+  display: none;
+  ${({ theme }) => theme.mediaQueries.lg} {
+    display: block;
+  }
+`
+
+const StyledLink = styled(Link)`
+  margin-right: 16px;
```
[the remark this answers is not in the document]
hachiojidev: You know this brings up an issue I have been wrestling with. If you are already extending the style is it better to put the stuff in here to keep the component below clean? I don't know but I have done this before. My guess it is basically the same thing.

src/views/Pools/components/PoolsTable/ActionPanel/Harvest.tsx
Chef-Cheems: Reply to this and couple of other comments about copy-paste.
Yeah, I was kinda trying not to mess with cards initially so I don't end up breaking everything. I'm now refactoring/extracting some common stuff where possible. Not everything though since for example prices are getting moved to redux state by you so there is no point making more abstractions based on hooks.

src/views/Pools/components/HelpButton.tsx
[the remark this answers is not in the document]
ChefHutch: Yea for sure. I'm pro that. Applying styles in two places on the same component will no doubt make debugging a bit trickier, even if the syntax is marginally neater 👍

Chef-Cheems:
@hachiojidev @ChefHutch Updated bunch of stuff here, commits starting from d18177b826ea618725f57115f7cd82539668f815 are new.
In short what happened:
- Using Farm/LP prices now
- Common logic is shared between cards / table
- Sorting by APR & Earned works
- Some minor bug fixes here and there

Preview also updated - https://awesome-raman-0813c6.netlify.app/

src/views/Pools/components/HelpButton.tsx
Chef-Cheems: endIcon wouldn't work here since we need to hide the text on small screens, with `endIcon` it's not possible.

src/components/Balance.tsx
Chef-Cheems: update on that - yea, TextProps doesn't allow onClick at all, needs to be explicitly defined here. TextProps inherits only from SpaceProps, TypographyProps, LayoutProps, and they don't contain onClick definition

hachiojidev (approved):

Chef-Cheems:
Rebased and force-pushed after https://github.com/pancakeswap/pancake-frontend/pull/1348 got merged

src/hooks/useApprove.ts
```diff
@@ -27,22 +31,94 @@ export const useApprove = (lpContract: Contract) => {
...
+  const handleApprove = () => {
+    cakeContract.methods
+      .approve(cakeVaultContract.options.address, ethers.constants.MaxUint256)
+      .send({ from: account })
+      .on('sending', () => {
+        setRequestedApproval(true)
+      })
+      .on('receipt', () => {
+        toastSuccess(t('Contract Enabled'), t('You can now stake in the %symbol% vault!', { symbol: 'CAKE' }))
+        setLastUpdated()
```
ChefHutch, line 88: (following our slack convos)

This `setLastUpdated` call is used to manually trigger a state change that used to cause the `checkApprovalStatus`  function to re-fire (not particularly elegant, I know).

On ln 106, you're instantiating a new `lastUpdated` & `setLastUpdated`, which I don't think will be responsive to `setLastUpdated()` being called here, on ln 88.

The impact for a user is that they have to refresh to trigger the vault approval check. Until they do that - they'll be able to endlessly submit approval transactions.

ChefHutch (changes requested):

src/hooks/useApprove.ts
Chef-Cheems: I think I've tested approving/disapproving multiple times after this change and it worked fine. Will double-check it once nodes become responsive.....

Chef-Cheems: Seems to be fine. I think all those other hooks involved in the process somehow "cache" the state 😄 
https://user-images.githubusercontent.com/81824236/119792445-5e5b7000-bede-11eb-8e7d-ec41c67bf620.mp4

ChefHutch: 👍 nice one, ah yea it could be because these are being called within hooks. yolo

ChefHutch (approved):
Looks great & seems to work well

PR #1274, opened 2021-05-20
fix: Missing more translations and add test that checks strings in json

memoyil:

### DIFF
--- a/src/__tests__/config/translations.test.ts
+++ b/src/__tests__/config/translations.test.ts
@@ -23,8 +23,7 @@ describe('Check translations available', () => {
     })
   }
 
-  throughDirectory('src/views')
-  throughDirectory('src/components')
+  throughDirectory('src/')
 
   it.each(files)('Translation key should exist in translations json', (file) => {
     const regex = /\bt\('([^']*)'/gm
--- a/src/__tests__/config/translations.test.ts
+++ b/src/__tests__/config/translations.test.ts
@@ -1,7 +1,45 @@
+import Path from 'path'
+import fs from 'fs'
 import translations from '../../config/localization/translations.json'
 
 describe('Check translations integrity', () => {
   it.each(Object.keys(translations))('Translation key value should be equal', (key) => {
     expect(key).toEqual(translations[key])
+  })
+})
+
+describe('Check translations available', () => {
+  const files = []
+  const translationKeys = Object.keys(translations)
+
+  function throughDirectory(directory) {
+    fs.readdirSync(directory).forEach((file) => {
+      const absolute = Path.join(directory, file)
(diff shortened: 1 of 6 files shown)


src/__tests__/config/translations.test.ts
```diff
@@ -1,7 +1,45 @@
...
+    const data = fs.readFileSync(file, { encoding: 'utf8', flag: 'r' })
+    let match
+    // eslint-disable-next-line no-cond-assign
+    while ((match = regex.exec(data)) !== null) {
+      if (match[1].trim()) {
+        const includes = translationKeys.includes(match[1])
+        try {
+          expect(includes).toBe(true)
+        } catch (e) {
+          throw new Error(`Something went wrong with in ${file} value ${match[1]}`)
```
Chef-Cheems, line 40: Maybe this could be more descriptive message like `Found unknown key ${match[1]} in ${file}`

memoyil: Done

Chef-Cheems (approved):
Looks good. I've talked with Bugs about different capitalization in some strings that we have, will have to go through everything to make stuff consistent. But I see that you just fixed the caps to match json file, so that's great 👍

src/__tests__/config/translations.test.ts
```diff
@@ -1,7 +1,45 @@
...
+  const translationKeys = Object.keys(translations)
+
+  function throughDirectory(directory) {
+    fs.readdirSync(directory).forEach((file) => {
+      const absolute = Path.join(directory, file)
+      if (fs.statSync(absolute).isDirectory()) return throughDirectory(absolute)
+      if (absolute.includes('.tsx')) {
+        return files.push(absolute)
+      }
+      return files.length
```
hachiojidev, line 22: Why are we returning `files.length`?

src/__tests__/config/translations.test.ts
```diff
@@ -1,7 +1,45 @@
...
+    })
+  }
+
+  throughDirectory('src/views')
+  throughDirectory('src/components')
```
hachiojidev, line 27: Is it faster to do it like this versus the entire `src` directory?

src/__tests__/config/translations.test.ts
memoyil: The return above returns new array's length so I keep the consistency although that return is not used at the end

src/__tests__/config/translations.test.ts
memoyil: Those are the 2 directories that we have translations, but I changed to "src" to traverse all directories

hachiojidev (approved):

PR #1277, opened 2021-05-20
fix(lp-price-hook): Add raw LP token totals to fetchFarms and re-calculate LP token price

ChefHutch:
Preview: https://lp-price-calc-fix.netlify.app/

LP price was coming through incorrectly from the `useLpTokenPrice` hook.

Previous price calculations were being made using `farm.lpTotalSupply` - _all_ LP tokens & `farm.lpTotalInQuoteToken` - _staked_ quote tokens in the MC contract. The staked total is about 10% less than the raw total. As we're using the raw total LP supply, we should also be calculating the LP unit price using the raw LP token amounts.

I've also made the maths in the `useLpTokenPrice` slightly more verbose and easier to understand.

### DIFF
--- a/src/state/farms/fetchFarms.ts
+++ b/src/state/farms/fetchFarms.ts
@@ -5,7 +5,6 @@ import multicall from 'utils/multicall'
 import { BIG_TEN } from 'utils/bigNumber'
 import { getAddress, getMasterChefAddress } from 'utils/addressHelpers'
 import { FarmConfig } from 'config/constants/types'
-import { DEFAULT_TOKEN_DECIMAL } from 'config'
 
 const fetchFarms = async (farmsToFetch: FarmConfig[]) => {
   const data = await Promise.all(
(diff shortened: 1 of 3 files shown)


src/state/farms/fetchFarms.ts
```diff
@@ -82,11 +80,13 @@ const fetchFarms = async (farmsToFetch: FarmConfig[]) => {
 
       return {
         ...farmConfig,
-        tokenAmount: tokenAmount.toJSON(),
-        quoteTokenAmount: quoteTokenAmount.toJSON(),
+        tokenAmountMc: tokenAmountMc.toJSON(),
+        quoteTokenAmountMc: quoteTokenAmountMc.toJSON(),
+        tokenAmountTotal: tokenAmountTotal.toJSON(),
+        quoteTokenAmountTotal: quoteTokenAmountTotal.toJSON(),
```
ChefHutch, line 86: After these changes, only `tokenAmountTotal` is actually used elsewhere in the codebase. 

These aren't, but I think it's conceivable that these values would be useful? Maybe I leave them in for now but rip them out after finishing the pricing refactor if they still aren't useful?

`tokenAmountMc` - amount of base token staked in MC
`quoteTokenAmountMc` - amount of quote token staked in MC 
`quoteTokenAmountTotal` - amount of quote token in total

src/state/farms/fetchFarms.ts
```diff
@@ -50,20 +49,19 @@ const fetchFarms = async (farmsToFetch: FarmConfig[]) => {
...
-      const tokenAmount = new BigNumber(tokenBalanceLP).div(BIG_TEN.pow(tokenDecimals)).times(lpTokenRatio)
-      const quoteTokenAmount = new BigNumber(quoteTokenBalanceLP)
-        .div(BIG_TEN.pow(quoteTokenDecimals))
-        .times(lpTokenRatio)
+      // Amount of token in the LP that are staked in the MC (i.e amount of token * lp ratio)
+      const tokenAmountMc = tokenAmountTotal.times(lpTokenRatio)
+      const quoteTokenAmountMc = quoteTokenAmountTotal.times(lpTokenRatio)
+
+      // Total staked in LP, in quote token value
+      const lpTotalInQuoteToken = quoteTokenAmountMc.times(new BigNumber(2))
```
ChefHutch, line 64: This was previously on ln 57. I've just tidied up the composition a bit and rephrased the comment.

memoyil, line 64: would be nice to have BIG_TWO 👍

hachiojidev (approved):

PR #1280, opened 2021-05-20
feat: Add translations for earn cards in home page

memoyil:
Since we have different components for apr and asset text, this is the solution that i found. It is better than nothing I think.


src/views/Home/components/EarnAssetCard.tsx
```diff
@@ -31,17 +32,21 @@ const latestPools: Pool[] = orderBy(activeNonCakePools, ['sortOrder', 'pid'], ['
 const assets = ['CAKE', ...latestPools.map((pool) => pool.earningToken.symbol)].join(', ')
 
 const EarnAssetCard = () => {
+  const { t } = useTranslation()
+  const assetText = t('Earn %assets% in Pools', { assets })
+  const assetTextParts = assetText.split(assets)
```
hachiojidev, line 37: I am...not a fan of this but lets give it a shot and roll with it.

hachiojidev (approved):

PR #1292, opened 2021-05-21
[wip] feat(farms-redux): Calculate and store token & quoteToken prices in farms redux

ChefHutch:
[wip] - storing token prices in farms redux


src/state/farms/fetchFarmsPrices.ts
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  if (quoteTokenFarm.quoteToken.symbol === 'BUSD') {
+    return new BigNumber(quoteTokenFarm.tokenPriceVsQuote)
+  }
+
+  return BIG_ZERO
+}
+
+const fetchFarmsPrices = async (farms) => {
```
ChefHutch, line 69: I can't type this shit as an array of farms `Farm[]` because of a string / BigNumber typing conflict. The `farms` its receiving are all with serialised BN values, so the types are:

```
tokenAmountMc: string;
quoteTokenAmountMc: string;
tokenAmountTotal: string;
quoteTokenAmountTotal: string;
etc.
```

Whereas in the `Farm` interface, these are all BNs

I'm not 100% why it's happy enough typing the action payload as `Farm[]` here https://github.com/pancakeswap/pancake-frontend/blob/5bfe5f074b493f3d314e310d91bd02186319ad02/src/state/farms/index.ts#L34, which is receiving the same farm object, with strings rather than BNs.

Similarly, I _think_ it's OK typing the argument as `Farm` in the .find/.map functions in the lines below this comment because the values are optional, and it isn't able to infer the types in the farm that's being passed to it.

Annoying, not sure the best solution here, I think the typing of these values as BN may just be wrong?

Chef-Cheems: I think the problem is that `Farm` itself is typed improperly, they are all strings (they can't be BigNumber cause they go to Redux), and we convert them to BigNumber in Farms table code.

So I guess the typings could be resolved by changing Farm to string? That might cause some typing issues elsewhere in the code, although I suppose not much (if at all) because improper BigNumber didn't raise any type errors

src/App.tsx
```diff
@@ -37,7 +37,7 @@ const App: React.FC = () => {
   useEagerConnect()
   useFetchPublicData()
   useFetchProfile()
-  useFetchPriceList()
+  useFetchApiPriceList()
```
ChefHutch, line 40: Renaming any of the old 'prices' state (fetched from API) to include `api` so that it's explicit where this data is coming from

src/config/constants/types.ts
```diff
@@ -10,6 +10,7 @@ export interface Token {
   address?: Address
   decimals?: number
   projectLink?: string
+  price?: string
```
ChefHutch, line 13: At the moment I'm attaching the token price to the `token` within farm state, i.e.

`farm.quoteToken.price`, `farm.token.price`.

The alternative to this would be to just sit it at the `farm` level:

`farm.quoteTokenPriceVsBusd`, `farm.tokenPriceVsBusd`

src/state/farms/fetchFarmsPrices.ts
ChefHutch: I think you're right Cheems. The issue is that we have some functions expecting these values to be BNs, so those would need to be changed too. I don't think it's too big a job to change these types from BN > string but would ideally do it in another PR IMO.

src/state/farms/index.ts
```diff
@@ -60,7 +61,8 @@ export const fetchFarmsPublicDataAsync = () => async (dispatch, getState) => {
   const fetchArchived = getState().farms.loadArchivedFarmsData
   const farmsToFetch = fetchArchived ? farmsConfig : nonArchivedFarms
   const farms = await fetchFarms(farmsToFetch)
-  dispatch(setFarmsPublicData(farms))
+  const farmsWithPrices = await fetchFarmsPrices(farms)
```
ChefHutch, line 64: Attaching prices to farms state before dispatch

src/state/hooks.ts
```diff
@@ -89,48 +89,15 @@ export const useFarmFromTokenSymbol = (tokenSymbol: string, preferredQuoteTokens
   return filteredFarm
 }
 
+// Return the base token price for a farm, from a given pid
 export const useBusdPriceFromPid = (pid: number): BigNumber => {
   const farm = useFarmFromPid(pid)
-  const bnbPriceBusd = usePriceBnbBusd()
```
ChefHutch, line 94: This function is now effectively a state selector. The logic is being done as part of the farms fetch.

src/config/constants/types.ts
```diff
@@ -10,6 +10,7 @@ export interface Token {
   address?: Address
   decimals?: number
   projectLink?: string
+  busdPrice?: string
```
ChefHutch, line 13: At the moment I'm attaching the token price to the `token` within farm state, i.e.

`farm.quoteToken. busdPrice`, `farm.token. busdPrice`.

The alternative to this would be to just sit it at the `farm` level:

`farm.quoteTokenPriceVsBusd`, `farm.tokenPriceVsBusd`

Chef-Cheems (commented):
Overall looks ok, some stuff will be further cleaned up by me when I'll integrate these prices into pools. One question about KUN - QSD:

I get that there is no QSD farm and we can't find the price because of it, but that in turn means that we can't calculate APR and Liquidity for it, so it shows broken row for it (screenshot). Should we introduce fallback for API price in case farm is not found?
Screenshot 2021-05-24 at 14 08 18

ChefHutch:
Great catch @Chef-Cheems - I've added a catch for instances where the price can't be obtained from the farm, and it'll use the API instead. As I say - this is currently only KUN-QSD.

src/state/farms/index.ts
```diff
@@ -61,7 +61,7 @@ export const fetchFarmsPublicDataAsync = () => async (dispatch, getState) => {
   const fetchArchived = getState().farms.loadArchivedFarmsData
   const farmsToFetch = fetchArchived ? farmsConfig : nonArchivedFarms
   const farms = await fetchFarms(farmsToFetch)
-  const farmsWithPrices = await fetchFarmsPrices(farms)
+  const farmsWithPrices = await fetchFarmsPrices(farms, getState)
```
Chef-Cheems, line 64: Would it be better to get prices the line before and pass them as 2nd argument instead of `getState` here?

Chef-Cheems (approved):

src/state/farms/fetchFarmsPrices.ts
```diff
@@ -0,0 +1,102 @@
...
+    const quoteTokenFarm = getFarmFromTokenSymbol(farms, farm.quoteToken.symbol)
+    let baseTokenPrice = getFarmBaseTokenPrice(farm, quoteTokenFarm, bnbPriceBusd)
+    let quoteTokenPrice = getFarmQuoteTokenPrice(farm, quoteTokenFarm, bnbPriceBusd)
+
+    // Catches to use API price if price cannot be obtained from farm. As of 24.5.21 this is only KUN-QSD
+    if (!baseTokenPrice.gt(0)) {
+      const tokenAddress = getAddress(farm.token.address)
+      baseTokenPrice = apiPriceData ? new BigNumber(apiPriceData[tokenAddress.toLowerCase()]) : BIG_ZERO
+    }
+    if (!quoteTokenPrice.gt(0)) {
```
hachiojidev, line 90: Do these always come back as numbers? Is the default 0? I would prefer we do not invert a check (readability). Can you use `eq`, `lt`, or `lte`?

hachiojidev (approved):
>FYI if you want to check a farm without a price, an example would be KUN - QSD. It currently has no way of figuring out its price, because it calculates the price relative to its quoteToken (QSD), and there is no QSD-BNB or QSD-BUSD farm. I have added a catch to use the api price redux store in this instance.

We need to keep an eye on this.

src/state/farms/fetchFarmsPrices.ts
ChefHutch: Yea `getFarmBaseTokenPrice` / `getFarmQuoteTokenPrice` will always return BigNumbers. Good shout on making the condition neater, I have added that.

ChefHutch:
> > FYI if you want to check a farm without a price, an example would be KUN - QSD. It currently has no way of figuring out its price, because it calculates the price relative to its quoteToken (QSD), and there is no QSD-BNB or QSD-BUSD farm. I have added a catch to use the api price redux store in this instance.
> 
> We need to keep an eye on this.

Yea I've let Roo know @hachiojidev - so should have some shared awareness across the team of when our fallback will kick in

ChefHutch:
Example of how this can be done without the prices API, making use of additional LPs for price https://github.com/ChefHutch/pancake-frontend/pull/16

src/config/constants/priceHelperLps.ts
```diff
@@ -0,0 +1,20 @@
...
+  {
+    pid: null,
+    lpSymbol: 'QSD-BNB LP',
+    lpAddresses: {
+      97: '',
+      56: '0x7b3ae32eE8C532016f3E31C8941D937c59e055B9',
+    },
+    token: tokens.qsd,
+    quoteToken: tokens.wbnb,
+  },
```
hachiojidev, line 17: Can you go into a bit more detail? Why are they this shape? Suppose to have the same shape as a farm?

ChefHutch: Yea sure - QSD doesn’t have a BNB or BUSD farm and so we can’t, with the methodology in fetchFarmsPrices, figure out its price.

By adding a QSD-BNB farm, it can figure out its price. In order to do that we need data about QSD’s price relative to BNB though, and so we need to effectively treat this extra LP like a normal farm and fetch details about the LP’s base & quote tokens, importantly tokenPriceVsQuote.

So once the LP helper farm has got all of the smart contract info, it’s passed to the price-fetching function and is used as part of that logic to help the KUN-QSD farm figure out its prices.

Finally, any LPs without pids are filtered out, so that only our masterchef farms (now all with prices) are actually stored to redux.

I just tidied up the flow here, might help understanding as well? https://github.com/pancakeswap/pancake-frontend/blob/655a6c9bc6c047a5fb4bccd1dbfa3a4b21feedf9/src/state/farms/index.ts#L64-L71

PR #1301, opened 2021-05-23
Feature/voting proposal

hachiojidev:
### Integration

Integrates voting directly in the app using [snapshot's API](https://docs.snapshot.org/graphql-api)

### Vote Caching & Verification

Currently voting power was calculated each time a proposal was loaded. Due to issues with our Syrup token the call to calculate a wallet's voting is very resource intensive.

To solve this we include the calculated voting power in the meta data along with a verification hash. The verification hash is checked when a user loads a proposal. This is to ensure that the voting power was calculated by our API. Any votes that are not validated by the API are filtered out (this is because anyone can vote using the API including entering any amount of voting power).


src/views/Voting/components/Proposals/TabMenu.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+const TabMenu: React.FC<TabMenuProps> = ({ proposalType, onTypeChange }) => {
+  const handleItemClick = (index: number) => {
+    onTypeChange(getTypeFromIndex(index))
+  }
+
+  return (
+    <StyledTabMenu>
+      <UIKitTabMenu activeIndex={getIndexFromType(proposalType)} onItemClick={handleItemClick}>
+        <Tab>Core</Tab>
```
memoyil, line 47: Translations should be added

package.json
```diff
@@ -59,10 +59,12 @@
     "react-dom": "^17.0.1",
     "react-feather": "^2.0.8",
     "react-helmet-async": "^1.0.9",
+    "react-markdown": "^6.0.2",
     "react-redux": "^7.2.4",
     "react-router-dom": "^5.2.0",
     "react-scripts": "^4.0.1",
     "react-transition-group": "^4.4.1",
+    "remark-gfm": "^1.0.0",
```
hachiojidev, line 67: Markdown table support

src/views/Voting/Proposal/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+        setNotFound(true)
+      }
+    }
+
+    fetchData()
+  }, [id, setNotFound, setProposal])
+
+  if (notFound) {
+    return <Redirect to="/voting" />
+  }
```
hachiojidev, line 51: Send them back to the voting index if we never loaded a proposal. Probably not ideal but will work for now.

src/components/DatePicker/DatePickerPortal.tsx
```diff
@@ -0,0 +1,112 @@
+import React from 'react'
```
hachiojidev, line 1: These styles unfortunately had to in a portal for it to work.

src/state/voting/helpers.ts
```diff
@@ -0,0 +1,134 @@
...
+          const response = await fetch(`${SNAPSHOT_VOTING_API}/verify`, {
+            method: 'post',
+            headers: {
+              'Content-Type': 'application/json',
+            },
+            body: JSON.stringify({
+              block: blockNumber,
+              votes: votesToVerify,
+            }),
+          })
```
hachiojidev, line 111: This is where we verify each vote.

src/utils/callHelpers.ts
```diff
@@ -128,54 +126,37 @@ export const soushHarvestBnb = async (sousChefContract, account) => {
...
-  try {
-    const masterContract = getMasterchefContract(web3WithArchivedNodeProvider)
-    const response = await masterContract.methods.userInfo(0, account).call(undefined, block)
-
-    return getBalanceAmount(new BigNumber(response.amount))
-  } catch (error) {
-    console.error('Error getting stake in CAKE pool', error)
-    return BIG_ZERO
-  }
+export const getActivePools = async (block?: number) => {
```
hachiojidev, line 132: We send the eligible pools to the voting API so there is a single source of truth. The FE is responsible to let the API know which pools to check for a balance given a block number.

src/components/DatePicker/DatePickerPortal.tsx
```diff
@@ -0,0 +1,112 @@
...
+    .react-datepicker__time
+    .react-datepicker__time-box
+    ul.react-datepicker__time-list
+    li.react-datepicker__time-list-item {
+    height: auto;
+    padding: 8px;
+  }
+`
+
+const DatePickerPoral = () => {
```
ChefHutch, line 108: Tiny typo pal - `DatePickerPortal`

src/components/layout/Container.tsx
```diff
@@ -1,16 +1,10 @@
...
-  padding-left: 16px;
-  padding-right: 16px;
-
-  ${({ theme }) => theme.mediaQueries.sm} {
-    padding-left: 24px;
-    padding-right: 24px;
-  }
-`
+const Container: React.FC<BoxProps> = ({ children, ...props }) => (
+  <Box px={['16px', '24px']} mx="auto" maxWidth="1200px" {...props}>
```
ChefHutch, line 5: Not sure if this might have an impact on when the container is restyled? i.e. in our `<Page>` https://github.com/pancakeswap/pancake-frontend/blob/d84936552b09c91435dcff9c9b68db100bb1f5bc/src/components/layout/Page.tsx#L10 Spreading `...props` into it probably means it should be OK?

src/state/types.ts
```diff
@@ -256,6 +256,89 @@ export interface PredictionsState {
   bets: BetData
 }
 
+// Voting
+
+/* eslint-disable camelcase */
+/**
+ * @see https://hub.snapshot.page/graphql
+ */
+export interface VoteWhere {
```
ChefHutch, line 265: At some point should we maybe consider breaking these `/state/types` into separate files, by view? Else there's no upper limit to how big this puppy could get.

src/views/Voting/CreateProposal/helpers.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  if (choices.length < MINIMUM_CHOICES) {
+    errors.choices = [t('Please create a minimum of %num% choices', { num: MINIMUM_CHOICES })]
+  }
+
+  const hasEmptyChoice = choices.some((choice) => !choice.value)
+  if (choices.length === MINIMUM_CHOICES && hasEmptyChoice) {
+    errors.choices = Array.isArray(errors.choices)
+      ? [...errors.choices, t('Choices must not be empty')]
```
ChefHutch, line 36: As this fires the moment you start filling in your first choice - maybe a better error here would be `'Your proposal must include at least 2 choices'`?

src/views/Voting/CreateProposal/helpers.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  if (!isValid(endTime)) {
+    errors.endTime = [t('Please select a valid time')]
+  }
+
+  const startDateTimestamp = combineDateAndTime(startDate, startTime)
+  const endDateTimestamp = combineDateAndTime(endDate, endTime)
+
+  if (endDateTimestamp < startDateTimestamp) {
```
ChefHutch, line 59: I think this is truthy even when the selected End Date is after the selected Start Date, if there is no End Time selected yet. Setting a non-null time default, or using 00:00 might solve this?... _and may create new issues_

ChefHutch (commented):
A few small things that tbh are at most Trello-worthy and can be solved post-release. I ran out of time today for a deeper review and there's a lot to get through 😅 . 

Apart from the voting power issue - it works well! I'll have time tomorrow morning to help with more testing & more reviewing.

package.json
```diff
@@ -73,8 +77,10 @@
     "@commitlint/config-conventional": "^12.1.1",
     "@pancakeswap/eslint-config-pancake": "^1.1.0",
     "@types/jest": "^26.0.23",
+    "@types/react-datepicker": "^3.1.8",
```
RabbitDoge, line 80: types for deps shouldn't be in the dev deps. I think if you try to build in prod mod it will fail

src/components/EasyMde/index.tsx
```diff
@@ -0,0 +1,84 @@
+import React, { TextareaHTMLAttributes, useEffect, useRef } from 'react'
```
RabbitDoge, line 1: is this generic enough to be in components ?

RabbitDoge (approved):

src/components/EasyMde/index.tsx
hachiojidev: Yes. It is a generic pancake markdown editor...

ChefHutch (approved):
As you mentioned at standup - it's hard to go through in detail when encountering the code 'cold' and reviewing PRs of this scale.

Still - the previews worked as expected, and the only points I found to raise were tiny. Nice one 👍

PR #1310, opened 2021-05-24
feat: Truncate wallet address helper

hachiojidev:

### DIFF
--- a/src/utils/truncateWalletAddress.ts
+++ b/src/utils/truncateWalletAddress.ts
@@ -1,0 +1,8 @@
+/**
+ * Truncates a wallet's address
+ */
+const truncateWalletAddress = (address: string, startLength = 4, endLength = 4) => {
+  return `${address.substring(0, startLength)}...${address.substring(address.length - endLength)}`
+}
+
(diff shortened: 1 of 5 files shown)


src/utils/truncateWalletAddress.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Truncates a wallet's address
+ */
+const truncateWalletAddress = (address: string, startLength = 4, endLength = 4) => {
+  return `${address.substring(0, startLength)}...${address.substring(address.length - endLength)}`
```
ChefHutch, line 5: Noice. If you do a search for `accountEllipsis` you'll see a few bits in the codebase that could use this. Any chance you could find & replace and remove that helper?

hachiojidev: Ha, weird...I must not have pushed my changes. I did update that.

Chef-Cheems (approved):

ChefHutch (approved):

RabbitDoge (approved):

PR #1316, opened 2021-05-24
style(apr): Fix right alignment of APR in farm table view

ChefHutch:
Tiny style tweak to fix left-alignment of APR in farm table view below `968` px

### DIFF
--- a/src/views/Farms/components/FarmTable/Apr.tsx
+++ b/src/views/Farms/components/FarmTable/Apr.tsx
@@ -26,6 +26,7 @@ const Container = styled.div`
   button {
     width: 20px;
     height: 20px;
+    margin-left: 0;
 
     svg {
       path {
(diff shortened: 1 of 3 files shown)


src/views/Farms/components/FarmTable/Apr.tsx
```diff
@@ -37,7 +38,12 @@ const Container = styled.div`
 
 const AprWrapper = styled.div`
   min-width: 60px;
-  text-align: left;
+  text-align: right;
+  margin-right: 14px;
```
hachiojidev, line 42: Margins should be in multiples of 4px.

hachiojidev (changes requested):

src/views/Farms/components/FarmTable/Apr.tsx
ChefHutch: Ah I think this is a crab cake introduction  - inherited from them. I'll fix it up in the other places it exists in farms too.

PR #1318, opened 2021-05-24
fix: (farms) Farm data and cake price cannot be seen for very long time mainly Firefox

memoyil:
Issue mainly happens on firefox but might also happen on other browsers as well. If one of the farms fetching got stuck, that causes already fetched farms data to not be pushed to redux. Therefore user cannot see farm details as well as cake/busd price for very long time. Since we are using slowRefresh there user has to wait additional one minute to renew the request (and that request might also get stuck).  So I added timeout promise alongside with fetch promise and use promise race.

Review: https://deploy-preview-1318--pancakeswap-dev.netlify.app/

To reproduce the issue:

1. Open main page in firefox
2. See cake price is not loading 
3. If it is loaded quickly refresh the page and check it again.

### DIFF
--- a/src/state/farms/fetchFarm.ts
+++ b/src/state/farms/fetchFarm.ts
@@ -2,8 +2,7 @@ import { Farm } from 'state/types'
 import fetchPublicFarmData from './fetchPublicFarmData'
 
 const fetchFarm = async (farm: Farm): Promise<Farm> => {
-  const { pid, lpAddresses, token, quoteToken } = farm
-  const farmPublicData = await fetchPublicFarmData(pid, lpAddresses, token, quoteToken)
+  const farmPublicData = await fetchPublicFarmData(farm)
 
   return { ...farm, ...farmPublicData }
 }
--- a/src/state/farms/fetchPublicFarmData.ts
+++ b/src/state/farms/fetchPublicFarmData.ts
@@ -1,96 +1,120 @@
 import BigNumber from 'bignumber.js'
-import { Address, Token } from 'config/constants/types'
 import masterchefABI from 'config/abi/masterchef.json'
 import erc20 from 'config/abi/erc20.json'
 import { getAddress, getMasterChefAddress } from 'utils/addressHelpers'
 import { BIG_TEN, BIG_ZERO } from 'utils/bigNumber'
 import multicall from 'utils/multicall'
+import { Farm, SerializedBigNumber } from '../types'
 
-const fetchFarm = async (pid: number, lpAddresses: Address, token: Token, quoteToken: Token) => {
+type PublicFarmData = {
+  tokenAmountMc?: SerializedBigNumber
+  quoteTokenAmountMc?: SerializedBigNumber
+  tokenAmountTotal?: SerializedBigNumber
+  quoteTokenAmountTotal?: SerializedBigNumber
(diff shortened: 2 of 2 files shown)


src/state/farms/fetchFarms.ts
```diff
@@ -8,88 +8,105 @@ import { FarmConfig } from 'config/constants/types'
...
+          quoteTokenAmountTotal: quoteTokenAmountTotal.toJSON(),
+          lpTotalSupply: new BigNumber(lpTotalSupply).toJSON(),
+          lpTotalInQuoteToken: lpTotalInQuoteToken.toJSON(),
+          tokenPriceVsQuote: quoteTokenAmountTotal.div(tokenAmountTotal).toJSON(),
+          poolWeight: poolWeight.toJSON(),
+          multiplier: `${allocPoint.div(100).toString()}X`,
+        }
       }
+
+      const timeout = new Promise((resolve) => {
```
ChefHutch, line 102: This seems reasonable Memoyil, might be worth adding a comment to explain why we're doing it and could you fix your prettier linting?

src/state/farms/fetchFarms.ts
```diff
@@ -8,88 +8,105 @@ import { FarmConfig } from 'config/constants/types'
...
+        const id = setTimeout(() => {
+          clearTimeout(id)
+          resolve(farmConfig)
+        }, 5000)
+      })
+
+      return Promise.race([timeout, farmFetch()])
```
Chef-Cheems, line 109: I wonder if that would be better handled by `Promise.allSettled` with proper rejection handling 🤔 
With this approach we create bunch of timers, not sure how much it affects performance, but seems a bit unnecessary

memoyil: Promise.allSettled needs promise to resolve or reject right ? In our issue it gets stuck that we get neither rejection nor resolve

I don't see any performance impact at first glance but you are right it needs to be further analyzed.

src/state/farms/fetchFarms.ts
memoyil: Logic will be changed a bit after this https://github.com/pancakeswap/pancake-frontend/pull/1332 merged, so I will wait until that to visit your comment.

What is wrong with my linting? I am using yarn prettier directly

src/state/farms/fetchFarms.ts
Chef-Cheems: aaah, in that case.... yeah, then race is better 👍

src/state/farms/fetchFarms.ts
ChefHutch: Cool cool. Some unintended prettier changes in this PR I think? Maybe not... the diff is a bit difficult to read. https://github.com/pancakeswap/pancake-frontend/pull/1318/files#diff-c124a2935612b3ec404c2824c52fc4f4ff5ff25ffe1a92bbc2f7e2b697d30c99L49-R57

memoyil: That happened because I extracted that logic into variable. All fetch logic indented.

ChefHutch:
There's been some more changes to the fetch Memoyil, be sure to pull in develop 👌

memoyil:
@hachiojidev I updated the pr rebasing current develop

hachiojidev (approved):

PR #1332, opened 2021-05-25
refactor: Break out farm fetch into own helper

hachiojidev:
Simply breaks out the fetching of a farm into a single function.


ChefHutch (approved):
Thanks for breaking these out into separate PRs 👍

memoyil:
We should still apply the timeout I think otherwise It will not proceed. Wdyt?

Chef-Cheems (approved):

src/state/farms/fetchPublicFarmData.ts
```diff
@@ -0,0 +1,92 @@
+import BigNumber from 'bignumber.js'
+import { Address, Token } from 'config/constants/types'
+import masterchefABI from 'config/abi/masterchef.json'
+import erc20 from 'config/abi/erc20.json'
+import { getAddress, getMasterChefAddress } from 'utils/addressHelpers'
+import { BIG_TEN } from 'utils/bigNumber'
+import multicall from 'utils/multicall'
+
+const fetchFarm = async (pid: number, lpAddresses: Address, token: Token, quoteToken: Token) => {
```
hachiojidev, line 9: We have an issue with typing `FarmConfig`, `Farm` and the state in redux. To mitigate this I extracted the needed arguments for more accurate typing.

hachiojidev: I also removed the merging of the config to keep this function more pure.

RabbitDoge (approved):

PR #1342, opened 2021-05-25
refactor: (TVL) Remove unused response variables

memoyil:


src/hooks/api.ts
```diff
@@ -2,12 +2,6 @@ import { useEffect, useState } from 'react'
 
 /* eslint-disable camelcase */
```
ChefHutch, line 3: Cool thanks Memoyil, you can probably remove this now too

ChefHutch (approved):

src/hooks/api.ts
memoyil: Done

hachiojidev:
@memoyil These types should stay. Defeats the purpose of typing if we only type what we use.

PR #1351, opened 2021-05-26
Feature/farm polling refactor

hachiojidev:
TODO

### DIFF
--- a/src/state/farms/getFarmsPrices.ts
+++ b/src/state/farms/getFarmsPrices.ts
@@ -72,8 +72,8 @@ const getFarmQuoteTokenPrice = (farm: Farm, quoteTokenFarm: Farm, bnbPriceBusd: 
   return BIG_ZERO
 }
 
-const fetchFarmsPrices = async (farms) => {
-  const bnbBusdFarm = farms.find((farm: Farm) => farm.pid === 252)
+const getFarmsPrices = (farms: Farm[]) => {
+  const bnbBusdFarm = farms.find((farm) => farm.pid === 252)
   const bnbPriceBusd = bnbBusdFarm.tokenPriceVsQuote ? new BigNumber(1).div(bnbBusdFarm.tokenPriceVsQuote) : BIG_ZERO
 
   const farmsWithPrices = farms.map((farm) => {
@@ -88,4 +88,4 @@ const fetchFarmsPrices = async (farms) => {
   return farmsWithPrices
 }
 
-export default fetchFarmsPrices
+export default getFarmsPrices
--- a/src/state/farms/index.ts
+++ b/src/state/farms/index.ts
@@ -1,10 +1,12 @@
 /* eslint-disable no-param-reassign */
-import { createSlice } from '@reduxjs/toolkit'
+import { createAsyncThunk, createSlice } from '@reduxjs/toolkit'
 import farmsConfig from 'config/constants/farms'
 import priceHelperLpsConfig from 'config/constants/priceHelperLps'
+import { FarmConfig } from 'config/constants/types'
 import isArchivedPid from 'utils/farmHelpers'
 import fetchFarms from './fetchFarms'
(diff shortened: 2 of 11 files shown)


src/state/farms/index.ts
```diff
@@ -27,6 +29,22 @@ const noAccountFarmConfig = farmsConfig.map((farm) => ({
 
 const initialState: FarmsState = { data: noAccountFarmConfig, loadArchivedFarmsData: false, userDataLoaded: false }
 
+// Async thunks
+
+export const fetchFarmsData = createAsyncThunk('farms/fetchFarmsData', async () => {
+  const farmsData = await fetchFarms(nonArchivedFarms)
+  return farmsData
+})
```
hachiojidev, line 37: Get non-archived farms once.

src/state/farms/index.ts
```diff
@@ -27,6 +29,22 @@ const noAccountFarmConfig = farmsConfig.map((farm) => ({
...
+
+export const fetchFarmData = createAsyncThunk<{ pid: number; data: Farm }, number>(
+  'farms/fetchFarmData',
+  async (pid) => {
+    const farm = farmsConfig.find((farmConfig) => farmConfig.pid === pid)
```
hachiojidev, line 42: I chose to get the farm data from the config instead of passing it in because we wouldn't be able to use it in any hooks as the object is mutating all the time causing re-render.

src/state/hooks.ts
```diff
@@ -55,6 +56,37 @@ export const usePollBlockNumber = () => {
...
+export const useSlowFarmPolling = (pid: number) => {
+  const { slowRefresh } = useRefresh()
+  const dispatch = useAppDispatch()
+
+  useEffect(() => {
+    if (pid) {
+      dispatch(fetchFarmData(pid))
+    }
+  }, [pid, slowRefresh, dispatch])
+}
```
hachiojidev, line 88: In the future I want to modify `useRefresh` to allow custom values. For now will compose these hooks.

src/state/hooks.ts
```diff
@@ -91,6 +122,9 @@ export const useFarmFromTokenSymbol = (tokenSymbol: string, preferredQuoteTokens
 // Return the base token price for a farm, from a given pid
 export const useBusdPriceFromPid = (pid: number): BigNumber => {
   const farm = useFarmFromPid(pid)
+
+  useSlowFarmPolling(pid)
```
hachiojidev, line 126: I think this hook covers the tokens from the LPs

src/state/hooks.ts
```diff
@@ -111,7 +145,7 @@ export const useLpTokenPrice = (symbol: string) => {
     // Double it to get overall value in LP
     const overallValueOfAllTokensInFarm = valueOfBaseTokenInFarm.times(2)
     // Divide total value of all tokens, by the number of LP tokens
-    const totalLpTokens = getBalanceAmount(farm.lpTotalSupply)
+    const totalLpTokens = getBalanceAmount(new BigNumber(farm.lpTotalSupply))
```
hachiojidev, line 148: `lpTotalSupply` was typed as `BigNumber` but was not.

src/state/farms/getFarmsPrices.ts
```diff
@@ -72,8 +72,8 @@ const getFarmQuoteTokenPrice = (farm: Farm, quoteTokenFarm: Farm, bnbPriceBusd:
   return BIG_ZERO
 }
 
-const fetchFarmsPrices = async (farms) => {
-  const bnbBusdFarm = farms.find((farm: Farm) => farm.pid === 252)
+const getFarmsPrices = (farms: Farm[]) => {
```
hachiojidev, line 75: @ChefHutch I renamed this function because it is not really fetching anything nor does it need to be asynchronous.

hachiojidev:
This PR is just no good, going to go back to the drawing board.

PR #1352, opened 2021-05-26
fix: Add missing translations in team battle

memoyil:


src/views/TradingCompetition/index.tsx
```diff
@@ -38,6 +38,7 @@ import BattleCta from './components/BattleCta'
 import PrizesInfo from './components/PrizesInfo'
 import Rules from './components/Rules'
 import TeamRanks from './components/TeamRanks'
+import { useTranslation } from '../../contexts/Localization'
```
hachiojidev, line 41: Absolute import please.

hachiojidev (changes requested):

src/views/TradingCompetition/index.tsx
memoyil: done

hachiojidev (approved):

PR #1353, opened 2021-05-26
chore(api-prices): Remove API prices state and hooks

ChefHutch:
- Removing the API prices hooks and redux store


hachiojidev (changes requested):
Please remove `PriceApiState`

ChefHutch:
Ah good catch @hachiojidev, done

hachiojidev (approved):

PR #1356, opened 2021-05-26
feat(lottery-v2): Lottery V2

ChefHutch:
### Planned merge & deploy time : 08:00 UTC / 16:00 SGT Jul 6th

**Files of particular note:**

Hooks / helpers:
- `state/lottery/fetchUnclaimedUserRewards.ts` - Lots of functions around checking if a user has winning tickets for a round, or rounds. It also formats data to be passed to the `claimTickets` contract call. `fetchUnclaimedUserRewards` is the route in
- `state/lottery/helpers.ts` - data calls and logic to help with state fetching 

Components:
- `BuyTicketsModal` - The ticket buying modal.
- `EditNumbersModal` - Editing ticket numbers (this work can be seen in isolation here - https://github.com/ChefHutch/pancake-frontend/pull/20)
- `generateTicketNumbers` + test - Logic for generating random ticket numbers for purchase
- `ClaimPrizesInner` - Some UI, but lots of logic around constructing the calls for claiming for a single, or multiple, rounds

Notes:
- **Merged**. ~@Chef-Cheems' PR with a range of styling & ui changes can be reviewed separately. It will be merged into this PR before release. https://github.com/ChefHutch/pancake-frontend/pull/23~


.env.development
```diff
@@ -12,4 +12,5 @@ REACT_APP_NODE_3 = "https://bsc-dataseed.binance.org"
 
 REACT_APP_GRAPH_API_PROFILE = "https://api.thegraph.com/subgraphs/name/pancakeswap/profile"
 REACT_APP_GRAPH_API_PREDICTION = "https://api.thegraph.com/subgraphs/name/pancakeswap/prediction"
+REACT_APP_GRAPH_API_LOTTERY = "https://api.thegraph.com/subgraphs/name/pancakeswap/lottery"
```
RabbitDoge, line 15: It should be in the env file only if the value change based on the node_env value

src/config/index.ts
```diff
@@ -29,3 +29,5 @@ export const LOTTERY_TICKET_PRICE = 1
 export const DEFAULT_TOKEN_DECIMAL = BIG_TEN.pow(18)
 export const DEFAULT_GAS_LIMIT = 200000
 export const DEFAULT_GAS_PRICE = 5
+export const TESTNET_CHAIN_ID = '97'
```
RabbitDoge, line 32: remove and uses `@pancakeswap-libs/sdk`

src/hooks/lottery/v2/useRoundEndCountdown.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  useEffect(() => {
+    const secondsRemainingCalc = endTime - startingSeconds
+    setSecondsRemaining(secondsRemainingCalc)
+
+    const tick = () => {
+      setSecondsRemaining((prevSeconds) => prevSeconds - 1)
+    }
+    const timerInterval = setInterval(() => tick(), 1000)
+    return () => clearInterval(timerInterval)
```
RabbitDoge, line 15: you also need to clear it every time the effect run since you have a non empty deps array

src/state/lottery/fetchUnclaimedUserRewards.ts
```diff
@@ -0,0 +1,136 @@
...
+  // Winning numbers are evaluated right-to-left in the smart contract, so we reverse their order for validation here:
+  // i.e. '1123456' should be evaluated as '6543211'
+  const ticketNumAsArray = ticketNumber.split('').reverse()
+  const winningNumsAsArray = finalNumber.split('').reverse()
+  const matchingNumbers = []
+
+  // The number at index 6 in all tickets is 1 and will always match, so finish at index 5
+  for (let index = 0; index < winningNumsAsArray.length - 1; index++) {
+    if (ticketNumAsArray[index] === winningNumsAsArray[index]) {
+      matchingNumbers.push(ticketNumAsArray[index])
```
dinhnhatbang, line 44: Should we handle the negative case first?
Example:
```
if (ticketNumAsArray[index] !== winningNumsAsArray[index]) {
   break;
}
// do something

File: .env.development (hunk 1 above)
Comment (ChefHutch) [reply .env.development]: Good point - I've made a PR moving all of these unchanging values out of the env variables https://github.com/pancakeswap/pancake-frontend/pull/1550

File: src/config/index.ts (hunk 2 above)
Comment (ChefHutch) [reply src/config/index.ts]: 👍 Done here https://github.com/pancakeswap/pancake-frontend/pull/1554

File: src/state/lottery/fetchUnclaimedUserRewards.ts (hunk 4 above)
Comment (ChefHutch) [reply src/state/lottery/fetchUnclaimedUserRewards.ts:53]: Sure, why not 👍

File: src/hooks/lottery/v2/useRoundEndCountdown.ts (hunk 3 above)
Comment (ChefHutch) [reply src/hooks/lottery/v2/useRoundEndCountdown.ts]: I think I've caught where this could be a problem, when the `nextEventTime` (formerly `endTime`) changes, the effect will re-run and a new timer will be created, [fixed here by clearing the timeout when the timer runs out](https://github.com/pancakeswap/pancake-frontend/blob/463c232aa80570bb09de2f55240d64e0985a0d8b/src/hooks/lottery/v2/useNextEventCountdown.ts#L16). Other than that, the useEffect won't re-run.

File: .env.development (hunk 1 above)
Comment (ChefHutch) [reply .env.development]: Ah, this does get changed at the env level sometimes within Netlify - we've had a few occasions where we needed to switch to using the BSC graph. I'm gonna leave them in.

File: src/views/LotteryV2/helpers.tsx [hunk 5]
```diff
@@ -0,0 +1,81 @@
...
+}
+
+/**
+ * Remove the '1' and reverse the digits in a lottery number retreived from the smart contract
+ */
+export const parseRetreivedNumber = (number: string): string => {
+  const numberAsArray = number.split('')
+  // remove the '1' from the number
+  numberAsArray.splice(0, 1)
+  // reverse it
```
hachiojidev, line 49: Not sure this comment is necessary :)

src/views/LotteryV2/helpers.tsx
```diff
@@ -0,0 +1,81 @@
...
+  numberAsArray.reverse()
+  return numberAsArray.join('')
```
hachiojidev, line 51: Or complete the trifecta and add

```
// and return it
```

src/views/LotteryV2/hooks/useGetNextLotteryEvent.ts
```diff
@@ -0,0 +1,48 @@
...
+    }
+    // Current lottery claimable. Next lottery has not yet started
+    if (status === LotteryStatus.CLAIMABLE) {
+      setNextEvent({
+        nextEventTime: endTime + blockBuffer + secondsBetweenRounds,
+        preCountdownText: t('Tickets on sale in'),
+        postCountdownText: null,
+      })
+    }
+  }, [status, endTime, t])
```
hachiojidev, line 43: Just checking but no infinite calls here having `t` as a dependency?

src/views/LotteryV2/components/BuyTicketsModal.tsx
```diff
@@ -0,0 +1,299 @@
...
+import { useAppDispatch } from 'state'
+import { usePriceCakeBusd, useLottery } from 'state/hooks'
+import { fetchUserTicketsAndLotteries } from 'state/lottery'
+import useTheme from 'hooks/useTheme'
+import useTokenBalance, { FetchStatus } from 'hooks/useTokenBalance'
+import useApproveConfirmTransaction from 'hooks/useApproveConfirmTransaction'
+import { useCake, useLotteryV2Contract } from 'hooks/useContract'
+import useToast from 'hooks/useToast'
+import UnlockButton from 'components/UnlockButton'
+import ApproveConfirmButtons, { ButtonArrangement } from 'views/Profile/components/ApproveConfirmButtons'
```
Chef-Cheems, line 20: Not that I suggest to fix it right away, but this should probably be moved out of Profile if we reuse it in other places

src/views/LotteryV2/components/BuyTicketsModal.tsx
```diff
@@ -0,0 +1,299 @@
...
+    if (cakeValueOfInput.gt(userCake)) {
+      setUserNotEnoughCake(true)
+    } else if (inputAsBN.gt(maxPossibleTicketPurchase)) {
+      setMaxTicketPurchaseExceeded(true)
+    } else {
+      setUserNotEnoughCake(false)
+      setMaxTicketPurchaseExceeded(false)
+    }
+    // Prevent decimals in input
+    const inputAsInt = parseInt(input)
```
Chef-Cheems, line 137: `parseInt(input, 10)`

src/hooks/lottery/v2/useRoundEndCountdown.ts
Chef-Cheems: @ChefHutch I'm not sure if it works as it looks in the snippet you shared. At least it reads a bit weird:
First time it runs `const timerInterval = setInterval(() => tick(), 1000)` the `timerInterval` inside `tick()` would be undefined I presume (which is fine). Next time what would be `timerInterval`? I think it will be undefined again since it is not declared before `tick()` is initialized. I don't think it will somehow remember `timerInterval` from previous useEffect run 🤔  
I think usually you track the timer id with refs or just another state variable.

src/views/LotteryV2/hooks/useGetNextLotteryEvent.ts
Chef-Cheems: @hachiojidev `t` is [wrapped](https://github.com/pancakeswap/pancake-frontend/blob/develop/src/contexts/Localization/Provider.tsx#L78) in `useCallback` and therefore should remain it's reference unless you change language

src/views/LotteryV2/svgs/Balls.tsx
```diff
@@ -0,0 +1,244 @@
...
+          clipRule="evenodd"
+          d="M2.10087 9.51443C3.77283 5.93689 6.78541 3.11142 10.4922 1.68435C10.5461 1.73247 10.5988 1.78231 10.6504 1.83387C12.4839 3.6674 10.912 5.74432 8.66157 7.99477C6.41112 10.2452 4.33429 11.817 2.50076 9.98347C2.3535 9.83621 2.22025 9.67943 2.10087 9.51443Z"
+          fill="white"
+        />
+      </g>
+    </Svg>
+  )
+}
+
+export default PinkBall
```
Chef-Cheems, line 244: Unless PinkBall is your favourite I don't think it needs separate default export 😄

Chef-Cheems: I will fix it on my branch though....

ChefHutch: But I love PinkBall...

src/hooks/lottery/v2/useRoundEndCountdown.ts
ChefHutch: Yea you could be right Cheems - am I right in thinking that the impact of the issue you're describing is just that when it comes to `clearInterval(timerInterval)` - it will be calling it on undefined, thus clearing no interval?

It's possible that it is in scope and isn't a problem - but you're right managing the timer via a ref is way nicer,  I've updated that.

src/state/hooks.ts
```diff
@@ -498,3 +504,113 @@ export const useGetProposalLoadingStatus = () => {
...
+    }
+  }, [dispatch, currentLotteryId, account])
+}
+
+export const useLottery = () => {
+  const currentRound = useSelector((state: State) => state.lottery.currentRound)
+  const maxNumberTicketsPerBuyOrClaimAsString = useSelector(
+    (state: State) => state.lottery.maxNumberTicketsPerBuyOrClaim,
+  )
+  const isTransitioning = useSelector((state: State) => state.lottery.isTransitioning)
```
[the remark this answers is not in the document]
ChefHutch: So `isTransitioning` is dispatched within the countdown - it's basically a boolean that I can use to have the UI react to the countdown being zero, and then respond again to a new lottery countdown starting (which means there has been a status change (i.e. 'open' > 'closed')). https://github.com/pancakeswap/pancake-frontend/blob/af3644c4e8c4cd46f7cdf9a10f53fe0013b6c9db/src/views/LotteryV2/hooks/useNextEventCountdown.ts#L25

The lottery `endTime` is fixed - but the timestamp when it _actually_ ends isn't totally predictable. It takes about 10 blocks from the `endTime` for the smart contract transaction to execute & broadcast. So I can get pretty close by counting down to `endTime` + 30s, but the lottery state sometimes hasn't actually changed by then - in that instance, I use `isTransitioning` to freeze some bits of the UI and handle these weird few seconds where the countdown thinks we should be moving onto the next phase of the lottery, but the smart contract doesn't.

Jesus TMI.

src/state/lottery/fetchUnclaimedUserRewards.ts
```diff
@@ -0,0 +1,190 @@
...
+const fetchUserTicketsForMultipleRounds = async (roundsToCheck: UserRound[], account: string) => {
+  // Build calls with data to help with merging multicall responses
+  const callsWithRoundData = roundsToCheck.map((round) => {
+    const totalTickets = parseInt(round.totalTickets, 10)
+    const calls = getViewUserTicketInfoCalls(totalTickets, account, round.lotteryId)
+    return { calls, lotteryId: round.lotteryId, count: calls.length }
+  })
+
+  // Batch all calls across all rounds
+  const multicalls = [].concat(...callsWithRoundData.map((callWithRoundData) => callWithRoundData.calls))
```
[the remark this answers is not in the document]
ChefHutch: So `callsWithRoundData` is an array of different objects, I map over them to extract just the calls... but I don't want these returned in an array of arrays (I want one array with all the calls - so they can be batched as a single call). 

Spreading the return of the map into `[].concat(...` was one way to achieve this. Maybe there's a better way?

src/state/lottery/helpers.ts
```diff
@@ -0,0 +1,256 @@
...
+    return null
+  }
+}
+
+export const getGraphLotteries = async (): Promise<LotteryRoundGraphEntity[]> => {
+  const response = await request(
+    GRAPH_API_LOTTERY,
+    gql`
+      query getLotteries {
+        lotteries(first: 100, orderDirection: desc, orderBy: block) {
```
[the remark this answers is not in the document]
ChefHutch: Yea there are quite a few piece of pagination/scalability that will need to be implemented following the launch, this being one of them! I'll make a note of follow-up tasks

src/views/LotteryV2/hooks/useGetNextLotteryEvent.ts
ChefHutch: Yea nice one both - good query, and I think Cheems is spot on.

The effect fires on language change, but that's it 👍

src/views/LotteryV2/components/BuyTicketsModal.tsx
ChefHutch: 100%. I'll add this to follow-up tasks, I think it got moved into Profile in some of the recent component refactoring work.

src/views/LotteryV2/svgs/Balls.tsx
ChefHutch: I won't make this change here to save you having conflicts with your BallHandling work :D

src/state/lottery/helpers.ts
```diff
@@ -0,0 +1,256 @@
...
+    const statusKey = Object.keys(LotteryStatus)[status]
+    const serializedCakePerBracket = cakePerBracket.map((cakeInBracket) => {
+      return new BigNumber(cakeInBracket.toString()).toJSON()
+    })
+    const serializedCountWinnersPerBracket = countWinnersPerBracket.map((winnersInBracket) => {
+      return new BigNumber(winnersInBracket.toString()).toJSON()
+    })
+    const serializedRewardsBreakdown = rewardsBreakdown.map((reward) => {
+      return new BigNumber(reward.toString()).toJSON()
+    })
```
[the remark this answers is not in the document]
ChefHutch: Yea man good point. I've added two here - https://github.com/pancakeswap/pancake-frontend/blob/d68eed053148fc6dd45f733c67395f1ae7b98883/src/utils/bigNumber.ts#L10-L13

I would break theses into a separate PR but we're a bit tight on time now to be merging multiple different pieces.

src/views/LotteryV2/components/BuyTicketsModal.tsx
```diff
@@ -0,0 +1,299 @@
...
+    setUserNotEnoughCake(false)
+    setMaxTicketPurchaseExceeded(false)
+  }
+
+  const { isApproving, isApproved, isConfirmed, isConfirming, handleApprove, handleConfirm } =
+    useApproveConfirmTransaction({
+      onRequiresApproval: async () => {
+        try {
+          const response = await cakeContract.allowance(account, lotteryContract.address)
+          const currentAllowance = new BigNumber(response.toString())
```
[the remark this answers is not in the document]
ChefHutch: Have done!

src/__tests__/views/lottery/generateTicketNumbers.test.ts
```diff
@@ -0,0 +1,93 @@
...
+    startingNumber: 1156791,
+    endingNumber: 1161790,
+    numbersToGenerate: 5000,
+  },
+]
+
+const withExistingTickets = [
+  {
+    ...withoutExistingNumbers[0],
+    numbersToGenerate: 99,
```
ChefHutch, line 24: Interestingly the tests flagged that if a user were to buy... 999,901 tickets (at a cost of ~$5m in CAKE plus a massive amount in gas), and they tried to buy 100 more (exceeding the max ticket range by 1) - the function would crash. 

It's a fairly absurd edge case. I can write something to cater for this if we really think it necessary.

src/config/constants/types.ts
```diff
@@ -153,3 +154,26 @@ export type PageMeta = {
...
+  PENDING = 'pending',
+  OPEN = 'open',
+  CLOSE = 'close',
+  CLAIMABLE = 'claimable',
+}
+
+export interface LotteryTicket {
+  id: string
+  number: string
+  status?: boolean
```
hachiojidev, line 168: Doesn't a lottery ticket always have/need a status?

src/config/constants/types.ts
```diff
@@ -153,3 +154,26 @@ export type PageMeta = {
...
+  rewardBracket?: number
+  roundId?: string
+  cakeReward?: SerializedBigNumber
```
hachiojidev, line 171: Same here, are these really not required?

src/state/hooks.ts
```diff
@@ -498,3 +500,80 @@ export const useGetProposalLoadingStatus = () => {
...
+export const useGetCurrentLotteryId = () => {
+  return useSelector((state: State) => state.lottery.currentLotteryId)
+}
+
+export const useGetUserLotteriesGraphData = () => {
+  return useSelector((state: State) => state.lottery.userLotteryData)
+}
+
+export const useGetUserLotteryGraphRoundById = (lotteryId: string) => {
+  const userLotteriesData = useSelector((state: State) => state.lottery.userLotteryData)
```
hachiojidev, line 514: Use `useGetUserLotteriesGraphData`

src/state/hooks.ts
```diff
@@ -498,3 +500,80 @@ export const useGetProposalLoadingStatus = () => {
...
+  return userLotteriesData.rounds.find((userRound) => userRound.lotteryId === lotteryId)
+}
+
+export const useGetLotteriesGraphData = () => {
+  return useSelector((state: State) => state.lottery.lotteriesData)
+}
```
hachiojidev, line 520: Duplicate of line 509?

src/state/lottery/helpers.ts
```diff
@@ -0,0 +1,295 @@
...
+      currentLotteryId: null,
+      maxNumberTicketsPerBuyOrClaim: null,
+    }
+  }
+}
+
+export const processRawTicketsResponse = (ticketsResponse: UserTicketsResponse): LotteryTicket[] => {
+  const ticketIds = ticketsResponse[0]
+  const ticketNumbers = ticketsResponse[1]
+  const ticketStatuses = ticketsResponse[2]
```
hachiojidev, line 112: `const [ticketIds, ticketNumbers, ticketStatuses] = ticketResponse`

src/state/lottery/helpers.ts
```diff
@@ -0,0 +1,295 @@
...
+  if (ticketIds.length > 0) {
+    return ticketIds.map((ticketId, index) => {
+      return {
+        id: ticketId.toString(),
+        number: ticketNumbers[index].toString(),
+        status: ticketStatuses[index],
+      }
+    })
+  }
+  return []
```
hachiojidev, line 123: I don't think you need this clause. `map` will return an empty array.

src/utils/bigNumber.ts
```diff
@@ -1,6 +1,13 @@
...
+import { ethers } from 'ethers'
+import { SerializedBigNumber } from 'state/types'
 
 export const BIG_ZERO = new BigNumber(0)
 export const BIG_ONE = new BigNumber(1)
 export const BIG_NINE = new BigNumber(9)
 export const BIG_TEN = new BigNumber(10)
+
+export const ethersToSerializedBigNumber = (ethersBn: ethers.BigNumber): SerializedBigNumber =>
+  ethersToBigNumber(ethersBn).toJSON()
```
hachiojidev, line 11: ethers' BigNumber has `toJSON()`.

src/views/LotteryV2/components/BuyTicketsModal/TicketInput.tsx
```diff
@@ -0,0 +1,278 @@
...
+      const nextDigitId = digitId + 1
+      // if we're not on the last digit - auto-tab
+      const nextInput = digitRefs[nextDigitId]
+      if (nextDigitId !== 6 && nextInput.current) {
+        nextInput.current.focus()
+      }
+    }
+
+    // Backsapce key
+    if (event.key === 'Backspace') {
```
hachiojidev, line 129: Maybe these comments are a bit redundant since you are using `Backspace` instead of the key code.

src/views/LotteryV2/components/BuyTicketsModal/generateTicketNumbers.tsx
```diff
@@ -0,0 +1,50 @@
+import { LotteryTicket } from 'config/constants/types'
+
+/**
+ * Return a random number between (and inclusive of) minNumber & maxNumber.
+ */
+function generateRandomNumber(minNumber: number, maxNumber: number): number {
+  const minInt = Math.ceil(minNumber)
+  const maxInt = Math.floor(maxNumber)
+  return Math.floor(Math.random() * (maxInt - minInt + 1) + minInt) // The maximum is inclusive and the minimum is inclusive
+}
```
hachiojidev, line 10: https://lodash.com/docs/4.17.15#random

src/hooks/lottery/v2/useRoundEndCountdown.ts
Chef-Cheems: Yeah there is no real issue I guess other than memory clogging + taking CPU time.

src/App.tsx
```diff
@@ -60,7 +60,7 @@ const App: React.FC = () => {
               <Pools />
             </Route>
             <Route path="/lottery">
-              <Lottery />
+              <LotteryV2 />
```
Chef-Cheems, line 63: Should be renamed to <Lottery /> but I guess it can wait until we remove old one from repo completely

src/hooks/useApproveConfirmTransaction.ts
```diff
@@ -106,25 +106,29 @@ const useApproveConfirmTransaction = ({
...
-        onApproveSuccess(state)
-      } else {
+      try {
+        const tx = await onApprove()
+        dispatch({ type: 'approve_sending' })
+        const receipt = await tx.wait()
+        if (receipt.status) {
+          dispatch({ type: 'approve_receipt' })
+          onApproveSuccess(state)
+        }
```
Chef-Cheems, line 116: Are you sure this doesn't need `dispatch({ type: 'approve_error' })` and toast in case `receipt.status` is `undefined`? I agree that logically if there is no status then something wrong happened and it will throw, but not sure if there is some other edge case

src/state/hooks.ts
Chef-Cheems: 509 is `userLotteryData`
this is `lotteriesData`

src/state/hooks.ts
```diff
@@ -498,3 +500,80 @@ export const useGetProposalLoadingStatus = () => {
...
+
+export const useGetLotteryGraphDataById = (lotteryId: string) => {
+  const lotteriesData = useSelector((state: State) => state.lottery.lotteriesData)
```
Chef-Cheems, line 523: `useGetLotteriesGraphData` can be used

src/state/hooks.ts
```diff
@@ -498,3 +500,80 @@ export const useGetProposalLoadingStatus = () => {
...
+}
+
+export const useFetchLottery = () => {
+  const { account } = useWeb3React()
+  const { fastRefresh } = useRefresh()
+  const dispatch = useAppDispatch()
+  const currentLotteryId = useGetCurrentLotteryId()
+
+  useEffect(() => {
+    // get current lottery ID, max tickets and historical lottery subgraph data
```
Chef-Cheems, line 534: max tickets?

Chef-Cheems: ah, I see it is configurable in the contract 👍

src/state/lottery/fetchUnclaimedUserRewards.ts
```diff
@@ -0,0 +1,209 @@
...
+  // The number at index 6 in all tickets is 1 and will always match, so finish at index 5
+  for (let index = 0; index < winningNumsAsArray.length - 1; index++) {
+    if (ticketNumAsArray[index] !== winningNumsAsArray[index]) {
+      break
+    }
+    matchingNumbers.push(ticketNumAsArray[index])
+  }
+
+  // Reward brackets refer to indexes, 0 = 1 match, 5 = 6 matches. Deduct 1 from matchingNumbers' length to get the reward bracket
+  const rewardBracket = matchingNumbers.length - 1
```
Chef-Cheems, line 64: Ignore for now: seems like a good candidate for Enum instead of array with indices. Not worth spending time refactoring this now though.

src/state/lottery/fetchUnclaimedUserRewards.ts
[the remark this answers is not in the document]
Chef-Cheems: could be `callsWithRoundData.flatMap((callWithRoundData) => callWithRoundData.calls)`
That would be transpiled into something like you have though cause .flatMap is only available since ES2019 and project target is ES5.
Could be changed but not critical

src/state/lottery/fetchUnclaimedUserRewards.ts
```diff
@@ -0,0 +1,209 @@
...
+  console.log(`all tickets round #${roundId}`, ticketsWithRewardBrackets)
+
+  // A rewardBracket of -1 means no matches. 0 and above means there has been a match
+  const allWinningTickets = ticketsWithRewardBrackets.filter((ticket) => {
+    return ticket.rewardBracket >= 0
+  })
+
+  // If ticket.status is true, the ticket has already been claimed
+  const unclaimedWinningTickets = allWinningTickets.filter((ticket) => {
+    return !ticket.status
```
Chef-Cheems, line 92: ignore for now: I guess `status` can be renamed to `claimed`? no need to refactor now if it takes long time I guess, but could be on todo list

src/state/pools/index.ts
```diff
@@ -151,7 +151,7 @@ export const updateUserPendingReward =
     dispatch(updatePoolsUserData({ sousId, field: 'pendingReward', value: pendingRewards[sousId] }))
   }
 
-export const fetchCakeVaultPublicData = createAsyncThunk<CakeVault>('cakeVault/fetchPublicData', async () => {
+export const fetchCakeVaultPublicData = createAsyncThunk<CakeVault>('cakeVault/fetchCurrentLotteryId', async () => {
```
Chef-Cheems, line 154: em... auto-replaced by accident? 🤔

ChefHutch: Lol great spot. There's a function in the lottery `fetchPublicData`, will have auto-replaced. thanks!

src/views/Ifos/components/IfoFoldableCard/IfoPoolCard/ContributeModal.tsx
```diff
@@ -65,6 +65,7 @@ const ContributeModal: React.FC<Props> = ({
       onApprove: () => {
         return raisingTokenContract.approve(contract.address, ethers.constants.MaxUint256, { gasPrice })
       },
+
```
Chef-Cheems, line 68: super nitpick but this adds one extra file to total changed 😄

src/hooks/useApproveConfirmTransaction.ts
ChefHutch: Picking up from our DMs - the previous if / else condition wasn't working as intended. The 'else' was never hit. So `dispatch({ type: 'approve_error' }) ` was never firing.

Wrapping it in a try/catch does catch & successfully trigger errors states.

src/views/LotteryV2/components/BuyTicketsModal/EditNumbersModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+  allComplete: boolean
+  onConfirm: () => void
+  isConfirming: boolean
+  onDismiss?: () => void
+}> = ({ totalCost, updateTicket, randomize, tickets, allComplete, onConfirm, isConfirming, onDismiss }) => {
+  const { theme } = useTheme()
+  const { t } = useTranslation()
+  return (
+    <StyledModal
+      title="Edit numbers"
```
Chef-Cheems, line 40: My bad - missing translation string.

src/views/LotteryV2/components/BuyTicketsModal/TicketInput.tsx
```diff
@@ -0,0 +1,278 @@
...
+
+  &:placeholder-shown {
+    text-align: left;
+  }
+
+  &:focus {
+    outline: none;
+  }
+
+  /* TODO check other browsers */
```
Chef-Cheems, line 55: Forgot to delete this comment. It works in FF, Chrome and Safari.
Not sure if webkit prefixes are needed though, but with stuff like spin-buttons and such I found that sometimes they don't get compiled correctly automatically, so maybe we should leave them in? Can test it later on different browsers without prefixes on prod built version.

src/views/LotteryV2/components/BuyTicketsModal/TicketInput.tsx
Chef-Cheems: True. Originally I was using `event.keyCode === NN` but `keyCode` seems to become deprecated so I changed to `key` but forgot to remove comments.

src/views/LotteryV2/components/PreviousRoundCard/Body.tsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react'
+import styled from 'styled-components'
+import { CardBody, Heading, Flex, Skeleton, Text, Box, Button, useModal } from '@pancakeswap/uikit'
+import { LotteryRound } from 'state/types'
+import { useGetUserLotteriesGraphData } from 'state/hooks'
+import { useTranslation } from 'contexts/Localization'
+import WinningNumbers from '../WinningNumbers'
+import ViewTicketsModal from '../ViewTicketsModal'
+
+const StyledCardBody = styled(CardBody)``
```
Chef-Cheems, line 10: Unnecessary

src/App.tsx
ChefHutch: Yup. I'll prep the OG Lottery removal PR later today

src/state/lottery/fetchUnclaimedUserRewards.ts
ChefHutch: Good point. They're used in a few different places - this could be a neat improvement

src/views/LotteryV2/components/BuyTicketsModal/TicketInput.tsx
ChefHutch: Yea I'd say leave them in for now Cheems. I'll add a note to the Trello refactors ticket to check & remove later

src/state/lottery/fetchUnclaimedUserRewards.ts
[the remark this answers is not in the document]
ChefHutch: Added a note to the refactors trello, as you say - not sure it's critical or even necessary.

src/config/constants/types.ts
ChefHutch: So `status` _is_ actually always there if I want to use it... good point. But some of the components that use the interface, don't use the `status` prop (https://github.com/pancakeswap/pancake-frontend/blob/039587a9e1e87b38ab544df3880120d448d601ee/src/views/LotteryV2/components/TicketNumber.tsx#L33).

I thought it was cleaner just to make it optional rather than make a new type for components making use of only _some_ of the `LotteryTicket`'s props?

src/config/constants/types.ts
ChefHutch: These are attached as part of rewards-fetching, not all `LotteryTicket`s have them and they aren't required

src/state/hooks.ts
ChefHutch: Yea it's the two different subgraph data points - one is user-specific (`userLotteryData`), the other is 'all lotteries' (`lotteriesData`)

src/state/lottery/helpers.ts
ChefHutch: Good point. 

I think it'll crash if it calls `map` on `undefined`/`null`... but it'll crash on the `if` anyway. I've tested this pretty extensively but just in case I'll improve the `if` to `if (ticketIds?.length > 0) {`. 

The `return []` then becomes a catch for when `ticketsResponse` is undefined, rather than an empty array.

src/utils/bigNumber.ts
ChefHutch: I think the pattern is still to convert to a `BigNumber.js` BN first, and then call `.toJSON()` though. I spoke with @RabbitDoge about this briefly a week or so ago - calling `.toJSON()` directly on the ethers BN can return a different value to the `BigNumber.js` equivalent.

src/views/LotteryV2/components/BuyTicketsModal/generateTicketNumbers.tsx
ChefHutch: Well damn ain't that a lot neater. Updated

src/state/hooks.ts
ChefHutch: You and Hachioji are IN SYNC!

hachiojidev (approved):

PR #1357, opened 2021-05-26
fix: Change Cakebusd price checks against 0 equality to greater than

memoyil:
Root cause: usePriceCakeBusd() returns NaN that also not equals to 0 (which makes our condition always true from the beginning in that component) which causes in firefox browser to not rerender the component when price is available

To reproduce: Check Farms & Staking section on homepage with Production build on firefox

Issue:

image

### DIFF
--- a/src/views/Home/components/CakeHarvestBalance.tsx
+++ b/src/views/Home/components/CakeHarvestBalance.tsx
@@ -39,7 +39,7 @@ const CakeHarvestBalance = () => {
   return (
     <Block>
       <CardValue value={earningsSum} lineHeight="1.5" />
-      {!cakePriceBusd.eq(0) && <CardBusdValue value={earningsBusd} />}
+      {cakePriceBusd.gt(0) && <CardBusdValue value={earningsBusd} />}
     </Block>
   )
 }
--- a/src/views/Pools/components/CakeVaultCard/VaultStakeModal.tsx
+++ b/src/views/Pools/components/CakeVaultCard/VaultStakeModal.tsx
@@ -46,7 +46,8 @@ const VaultStakeModal: React.FC<VaultStakeModalProps> = ({ pool, stakingMax, isR
   const [percent, setPercent] = useState(0)
   const { hasUnstakingFee } = useWithdrawalFeeTimer(parseInt(lastDepositedTime, 10), userShares)
   const cakePriceBusd = usePriceCakeBusd()
-  const usdValueStaked = stakeAmount && formatNumber(new BigNumber(stakeAmount).times(cakePriceBusd).toNumber())
+  const usdValueStaked =
+    cakePriceBusd.gt(0) && stakeAmount ? formatNumber(new BigNumber(stakeAmount).times(cakePriceBusd).toNumber()) : ''
 
   const handleStakeInputChange = (input: string) => {
     if (input) {
@@ -174,7 +175,7 @@ const VaultStakeModal: React.FC<VaultStakeModalProps> = ({ pool, stakingMax, isR
       <BalanceInput
         value={stakeAmount}
         onUserInput={handleStakeInputChange}
-        currencyValue={`~${usdValueStaked || 0} USD`}
+        currencyValue={cakePriceBusd.gt(0) && `~${usdValueStaked || 0} USD`}
       />
       <Text mt="8px" ml="auto" color="textSubtle" fontSize="12px" mb="8px">
         {t('Balance: %balance%', { balance: getFullDisplayBalance(stakingMax, stakingToken.decimals) })}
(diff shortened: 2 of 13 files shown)


src/views/Pools/components/CakeVaultCard/VaultStakeModal.tsx
```diff
@@ -174,7 +175,7 @@ const VaultStakeModal: React.FC<VaultStakeModalProps> = ({ pool, stakingMax, isR
       <BalanceInput
         value={stakeAmount}
         onUserInput={handleStakeInputChange}
-        currencyValue={`~${usdValueStaked || 0} USD`}
+        currencyValue={cakePriceBusd.gt(0) && `~${usdValueStaked || 0} USD`}
```
Chef-Cheems, line 178: I think this is unnecessary since it is checked above already, it would show 0 USD even without this condition

memoyil: If we remove that check and If cakePriceBusd is NaN (when price is being loaded or something wrong when loaded) inside the model it will always show ~0 usd below the stake amount which can confuse the user.

Chef-Cheems: hm... makes sense 👍

Chef-Cheems (approved):

src/views/Home/components/CakeHarvestBalance.tsx
```diff
@@ -39,7 +39,7 @@ const CakeHarvestBalance = () => {
   return (
     <Block>
       <CardValue value={earningsSum} lineHeight="1.5" />
-      {!cakePriceBusd.eq(0) && <CardBusdValue value={earningsBusd} />}
+      {cakePriceBusd.gt(0) && <CardBusdValue value={earningsBusd} />}
```
ChefHutch, line 42: I understand why these `!....eq(0)` > `.gt(0)` conditions are an improvement and directly address the small issue that the PR is highlighting - but curious as to whether the other conditional changes are fixing _actual_ problems?

Particularly when tackling issues that appear momentarily in specific circumstances, it would be good if you could include a short video/instructions on how to recreate the issue - it eases the burden on us figuring what's wrong/how your work fixes it.

ChefHutch (approved):
LGTM Memoyil. Please see comment about being very specific and deliberate about 

1. What you're fixing
2. How you're fixing it

We're a small team and this will help us review quickly.

src/views/Home/components/CakeHarvestBalance.tsx
ChefHutch: > That fix actually fixes the issue that shown in the image I put in description.

Yea I realise that. My comment is more pointed towards the other changes.

> It shows ~$0 when cake price is being loaded and in Firefox (strangely) after the cake price is loaded. And more strangely it only happens in production build not the development one. 

All very important information - browser & env specific instructions on how to recreate it. Please include this in the PR description.

> happens because now we have NaN instead of 0 at first

Yea it'll be undefined now right? Maybe an alternative solution to look at at some point would be initialising the `busdPrice` as a zero BN.

memoyil: Don't want to touch the new usePriceCakeBusd logic, but if we add there to return BIG_ZERO when output is NaN we don't need to do all these changes. What do you prefer?

ChefHutch: I think it would make sense to ensure we're always going to get a BN returned by those hooks (`usePriceBnbBusd` & `useBusdPriceFromPid` too).

Is the root cause that the farm isn't yet in state? i.e. `cakeBnbFarm` is undefined?
or is the issue that the `token` doesn't yet have a `busdPrice`, so `token.busdPrice` is undefined? 

If it's the latter I think it'll be as simple as `return new BigNumber(cakeBnbFarm.token.busdPrice || 0)`?

I think you're still right to modify the conditions that I mentioned in my first comment, from  `!....eq(0)` > `.gt(0)`

With regards to whether the other changes will still be necessary - I'll leave that up to you.

memoyil: Is the root cause that the farm isn't yet in state? i.e. cakeBnbFarm is undefined?
or is the issue that the token doesn't yet have a busdPrice, so token.busdPrice is undefined?

`cakeBnbFarm is defined but doesn't yet have its public farm data or token.busdprice (due to both farms (cake- bnb or bnb-busd) promises might get stuck which i try to fix it here https://github.com/pancakeswap/pancake-frontend/pull/1318) `

memoyil:
@ChefHutch  I will keep it as it is and let cake price be NaN when farm token price is loading, I think it is more intuitive than being 0 when loading.

hachiojidev (approved):
Next time lets break PR's like this up into smaller chunks

PR #1358, opened 2021-05-26
feat(prediction): Default gasPrice on BetBull/BetBear

ChefKai:
To remove when new smart contract is live, or to update when BSC network fees are updated.

### DIFF
--- a/src/views/Predictions/components/RoundCard/SetPositionCard.tsx
+++ b/src/views/Predictions/components/RoundCard/SetPositionCard.tsx
@@ -25,6 +25,7 @@ import useToast from 'hooks/useToast'
 import { BetPosition } from 'state/types'
 import { getDecimalAmount } from 'utils/formatBalance'
 import UnlockButton from 'components/UnlockButton'
+import { BIG_NINE, BIG_TEN } from 'utils/bigNumber'
 import PositionTag from '../PositionTag'
 import { getBnbAmount } from '../../helpers'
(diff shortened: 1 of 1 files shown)


Chef-Chungus (approved):

src/views/Predictions/components/RoundCard/SetPositionCard.tsx
```diff
@@ -38,6 +39,10 @@ interface SetPositionCardProps {
   onSuccess: (decimalValue: BigNumber, hash: string) => Promise<void>
 }
 
+// /!\ TEMPORARY /!\
+// Set default gasPrice (6 gwei) when calling BetBull/BetBear before new contract is released fixing this 'issue'.
+const gasPrice = new BigNumber(6).times(BIG_TEN.pow(BIG_NINE)).toString()
```
ChefHutch, line 44: I realise you'll probably be on top of this Kai when the change can be made, but we often add the text 'TODO' in the comment so that temporary code / work to be done can be easily searched for and found - can you add?

ChefHutch (approved):

src/views/Predictions/components/RoundCard/SetPositionCard.tsx
ChefKai: Done in cd3d4c4b9c3cdcb0b1e85e180652ffb56d965651

PR #1367, opened 2021-05-27
fix: (Team Battle) Add missing translations in team battle and fix test to check translations with carriage return

memoyil:
Review: https://deploy-preview-1367--pancakeswap-dev.netlify.app/

To reproduce the issue: 

1. Change language to French or Magyar for example (the text that is missing is translated there)
2. Go to Team Battle
3. Go to FAQ
4. Click details of "How do I claim my prize(s)?" (the bottom one)
5. See the answer is not translated

I also added a rule to unit test that catches these translation keys

### DIFF
--- a/src/views/TradingCompetition/components/Rules/FAQs.tsx
+++ b/src/views/TradingCompetition/components/Rules/FAQs.tsx
@@ -103,7 +103,7 @@ const FAQ = () => {
           <FoldableText title={t('How do I claim my prize(s)?')} mt="24px">
             <Text fontSize="14px" color="textSubtle">
               {t(
-                'After the battle ends, visit the event page and click the “Claim Prizes” button in the top section or in the “Your Score” section. Once you claim your prizes successfully, the button text will change to “Prizes Claimed”.',
+                'After the battle ends, visit the event page and click the “Claim Prizes” button in the top section or in “Your Score” section. Once you claim your prizes successfully, the button text will change to “Prizes Claimed”.',
               )}
             </Text>
           </FoldableText>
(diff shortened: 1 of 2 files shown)


src/views/TradingCompetition/components/Rules/FAQs.tsx
```diff
@@ -103,7 +103,7 @@ const FAQ = () => {
           <FoldableText title={t('How do I claim my prize(s)?')} mt="24px">
             <Text fontSize="14px" color="textSubtle">
               {t(
-                'After the battle ends, visit the event page and click the “Claim Prizes” button in the top section or in the “Your Score” section. Once you claim your prizes successfully, the button text will change to “Prizes Claimed”.',
+                'After the battle ends, visit the event page and click the “Claim Prizes” button in the top section or in “Your Score” section. Once you claim your prizes successfully, the button text will change to “Prizes Claimed”.',
```
hachiojidev, line 106: Why remove `the`? Seems correct to me.

memoyil: It is removed by Crowdin,

English translations updated here 

https://github.com/pancakeswap/pancake-frontend/commit/fa8f24b24098d2f00f3ab2782df87bad22ec4213#diff-fdd36d293ae5980f1c3962fba7dc86c9672b8b6278bc4e6fd5657775d91ba1ff

Then I decided to update our keys based on that

https://github.com/pancakeswap/pancake-frontend/commit/8e916570418ea36de948fc11881a1447296c4ee5#diff-f0ac50d3e3553110bae1a843a44f7a32629ecff7e31a122d27560688f662f1fe

But I forgot to change this part, but our translation test couldn't detect the missing key because of carriage return and spaces. (which I added an additional regex rule for that in this pr as well)

hachiojidev (approved):

PR #1373, opened 2021-05-27
fix: (prediction) Add missing timer label translation

memoyil:
Review: https://deploy-preview-1373--pancakeswap-dev.netlify.app/

To reproduce the issue:

1. Go to prediction
2. Select greek language for example
3. Check timer label in top right
4. See the minute unit is not translated

Before: 

image

After: 

Screenshot 2021-05-27 at 14 21 23

### DIFF
--- a/src/views/Predictions/components/Label.tsx
+++ b/src/views/Predictions/components/Label.tsx
@@ -101,9 +101,10 @@ export const PricePairLabel: React.FC = () => {
 
 interface TimerLabelProps {
   interval: string
+  unit: 'm' | 'h' | 'd'
 }
 
-export const TimerLabel: React.FC<TimerLabelProps> = ({ interval }) => {
+export const TimerLabel: React.FC<TimerLabelProps> = ({ interval, unit }) => {
   const seconds = useRoundCountdown()
   const countdown = formatRoundTime(seconds)
   const { t } = useTranslation()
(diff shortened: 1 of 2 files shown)


Chef-Cheems:
I'm not sure this is the right way to do it. People who will be translating this will just see `m` without context. We need to discuss how to approach this with Crowdin,  maybe there is a way to provide a comment next to string like "`m` as in 10m referring to 10 minutes". 🤔
@hachiojidev any ideas?

memoyil:
There is a "%num%d" key in translations.json but never translated so i decided to translate only the unit

memoyil:
> I'm not sure this is the right way to do it. People who will be translating this will just see `m` without context. We need to discuss how to approach this with Crowdin, maybe there is a way to provide a comment next to string like "`m` as in 10m referring to 10 minutes". 🤔
> @hachiojidev any ideas?

we can add tooltip :)

hachiojidev (approved):
@Chef-Cheems @memoyil 

The best solution is as Cheems mentioned, be able to provide context per translation string. Unfortunately Crowdin does not support context when your file format is JSON
https://support.crowdin.com/file-formats/json/

I am not so sure this is worth changing the system now so I think we will have to accept a less ideal solution of translating units. I quickly looked into formatting options with date-fns but that too is probably overkill for this situation.

Note - We can add context via Crowdin's interface. Maybe in the future when we get these kinds of strings we make a note to update the context in Crowdin.

PR #1376, opened 2021-05-27
fix: (select) Selected value is not translated in select component

memoyil:
Review: https://deploy-preview-1376--pancakeswap-dev.netlify.app/

To produce the issue:

1. Go to farms
2. Change sort by value from select component or keep it as it is
3. Try different languages
4. See selected value is not translated

Before: 

https://user-images.githubusercontent.com/2213635/119902869-1ec96e00-bf48-11eb-9138-f93266e86479.mov

After: 

https://user-images.githubusercontent.com/2213635/119902883-225cf500-bf48-11eb-9726-f9ffb50435ac.mov

### DIFF
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -105,17 +105,17 @@ const Select: React.FunctionComponent<SelectProps> = ({ options, onChange }) => 
   const containerRef = useRef(null)
   const dropdownRef = useRef(null)
   const [isOpen, setIsOpen] = useState(false)
-  const [selectedOption, setSelectedOption] = useState(options[0])
+  const [selectedOptionIndex, setSelectedOptionIndex] = useState(0)
   const [containerSize, setContainerSize] = useState({ width: 0, height: 0 })
 
   const toggling = () => setIsOpen(!isOpen)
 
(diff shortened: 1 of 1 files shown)


src/components/Select/Select.tsx
```diff
@@ -119,6 +119,15 @@ const Select: React.FunctionComponent<SelectProps> = ({ options, onChange }) =>
...
 
+  useEffect(() => {
+    const updatedSelection = options.find(
+      (option) => selectedOption.value === option.value && selectedOption.label !== option.label,
+    )
+    if (updatedSelection) {
+      setSelectedOption(updatedSelection)
+    }
+  }, [options, selectedOption])
+
```
hachiojidev, line 130: I understand that this fixes the problem but it is not an optimal solution. This is essentially a hook specifically for the translation function.

I am not sure if this would work but I think it would be better for `selectedOption` to be the index of the options instead. That way we are using the latest item in options.

Can you test that theory?

hachiojidev (changes requested):

src/components/Select/Select.tsx
memoyil: Thanks, I applied and it works. 👍

hachiojidev (approved):

PR #1382, opened 2021-05-28
feat: Reduce amount of requests when tab is hidden

memoyil:
This will reduce amount of requests made when tab is hidden (not focus)

For example in pools when tab or window is hidden, Amount of requests is reduced from ~600 to ~30 per minute

Check it from

https://deploy-preview-1382--pancakeswap-dev.netlify.app/

### DIFF
--- a/src/hooks/useTabVisibility.ts
+++ b/src/hooks/useTabVisibility.ts
@@ -1,0 +1,21 @@
+import { useEffect, useRef } from 'react'
+
+const useTabVisibility = () => {
+  const tabVisibleRef = useRef(true)
+
+  useEffect(() => {
+    const onVisibilityChange = () => {
+      tabVisibleRef.current = !document.hidden
+    }
+
+    window.addEventListener('visibilitychange', onVisibilityChange)
+
+    return () => {
+      window.removeEventListener('visibilitychange', onVisibilityChange)
+    }
+  }, [])
+
+  return { tabVisibleRef }
+}
+
+export default useTabVisibility
--- a/src/contexts/RefreshContext.tsx
+++ b/src/contexts/RefreshContext.tsx
@@ -1,4 +1,5 @@
 import React, { useState, useEffect } from 'react'
+import useTabVisibility from 'hooks/useTabVisibility'
 
 const FAST_INTERVAL = 10000
 const SLOW_INTERVAL = 60000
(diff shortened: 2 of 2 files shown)


hachiojidev (commented):
I want to take these changes in steps. First, `useRefresh` should pause when the tab is not active.

src/utils/throttled.ts
```diff
@@ -0,0 +1,14 @@
...
+      func.apply(...args)
+      inThrottle = true
+      setTimeout(() => {
+        inThrottle = false
+      }, limit)
+    }
+  }
+}
+
+export default throttled
```
hachiojidev, line 14: Use the one from lodash.

src/components/IdleModal.tsx
```diff
@@ -0,0 +1,55 @@
...
+        <Text as="p" mb="16px">
+          {t('You will get timed out. You want to stay?')}
+        </Text>
+        <Button width="100%" onClick={handleConfirm} mb="16px">
+          {t('Continue')}
+        </Button>
+      </ModalBody>
+    </ModalContainer>
+  )
+}
```
hachiojidev, line 53: I don't think this modal is necessary. We can pause the fetching when the user is idle or tab inactive. As I mentioned in my other comment let's start with just pausing `useRefresh` when the tab is inactive.

src/utils/throttled.ts
memoyil: done

src/components/IdleModal.tsx
memoyil: I removed idle modal and make tab active set to to true in tabActivity hook.

memoyil: > I don't think this modal is necessary. We can pause the fetching when the user is idle or tab inactive. As I mentioned in my other comment let's start with just pausing `useRefresh` when the tab is inactive.

If we put that check in useRefresh, we might have consequences for example cake price in page title showing an old value when tab is hidden or waiting a bit to get updated state in prediction (which might be frustrating for the user).

memoyil:
@hachiojidev With this approach we are not indicating user that they are idle and want it to continue. Is it ok ?

src/hooks/useTabVisibility.ts
```diff
@@ -0,0 +1,20 @@
+import { useEffect, useState } from 'react'
+
+const useTabVisibility = () => {
```
memoyil, line 3: this is not used and logic moved to useTabAcitivity hook but for now I keep it because it might be used if this feature implementation changes

src/hooks/useTabVisibility.ts
```diff
@@ -0,0 +1,37 @@
...
+  const [isTabVisible, setIsTabVisible] = useState(true)
+
+  useEffect(() => {
+    const onFocus = () => {
+      onVisibilityChange()
+    }
+
+    const onVisibilityChange = () => {
```
memoyil, line 11: I put cooldown period of 60sec to publish visibility state from true to false, to not make so much request if user switches tab or window frequently.

hachiojidev (commented):
A couple things.

Why are we not using the page visibility api?
 https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API
 
 As we discussed I want to move this logic higher. I still think the best place for this is in the global `useRefresh` hook.

ChefHutch (commented):
This is neat @memoyil but I agree with @hachiojidev - incorporating this into the useRefresh hook would be preferred to passing a new argument into every hook we want to make responsive to tab-visibility (which isn't bad, it just adds quite a bit of overhead).

> As we discussed I want to move this logic higher. I still think the best place for this is in the global useRefresh hook.

memoyil:
> A couple things.
> 
> Why are we not using the page visibility api?
> https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API
> 
> As we discussed I want to move this logic higher. I still think the best place for this is in the global `useRefresh` hook.

I moved logic to refreshcontext and changed to listen visibilitychange event as you suggested. Since we cannot use onblur or onfocus (not reliable when multiple monitors or when popup window appears) when user switches between programs or windows (not tabs) it does not fire visibility change event neither the document.hidden is false. To cover them we should apply idle check in future.

So the good news is if tab is hidden (meaning if browser windows is minimized or tab is in background), we only request usePollBlockNumber which is once in 6 seconds

So in pools it reduces amount of request from ~600 to 10 per minute.

hachiojidev (approved):
Nice work. We can test this out and build upon it by adding more strategies to reduce unnecessary calls.

PR #1384, opened 2021-05-28
chore(useTokenBalance): Add hasFetchedBalance state to useTokenBalance

ChefHutch:
At the moment the state returned by `useTokenBalance` defaults to `BIG_ZERO`

Meaning there's no difference in state between these two scenarios:
- ‘finished fetching, user has none of that token’
- ‘not yet fetched the user balance’

In the new Lottery, we may want to conditionally send the users to purchase more CAKE if they have a zero balance. We wouldn't want to show that while the user balance is still being fetched. The new `fetchStatus` enum allows us to do this.

### DIFF
--- a/src/hooks/useTokenBalance.ts
+++ b/src/hooks/useTokenBalance.ts
@@ -7,8 +7,13 @@ import useWeb3 from './useWeb3'
 import useRefresh from './useRefresh'
 import useLastUpdated from './useLastUpdated'
 
+export type FetchStatus = 'notFetched' | 'fetching' | 'success' | 'failed'
+
 const useTokenBalance = (tokenAddress: string) => {
-  const [balanceState, setBalanceState] = useState({ balance: BIG_ZERO, hasFetchedBalance: false })
+  const [balanceState, setBalanceState] = useState({
+    balance: BIG_ZERO,
+    fetchStatus: 'notFetched' as FetchStatus,
+  })
   const { account } = useWeb3React()
   const web3 = useWeb3()
   const { fastRefresh } = useRefresh()
@@ -16,11 +21,19 @@ const useTokenBalance = (tokenAddress: string) => {
   useEffect(() => {
     const fetchBalance = async () => {
       const contract = getBep20Contract(tokenAddress, web3)
+      setBalanceState((prev) => ({
+        ...prev,
+        fetchStatus: 'fetching',
+      }))
       try {
         const res = await contract.methods.balanceOf(account).call()
-        setBalanceState({ balance: new BigNumber(res), hasFetchedBalance: true })
+        setBalanceState({ balance: new BigNumber(res), fetchStatus: 'success' })
       } catch (e) {
         console.error(e)
+        setBalanceState((prev) => ({
+          ...prev,
+          fetchStatus: 'failed',
+        }))
       }
     }
 
@@ -29,7 +42,7 @@ const useTokenBalance = (tokenAddress: string) => {
     }
   }, [account, tokenAddress, web3, fastRefresh])
 
-  return { ...balanceState }
+  return balanceState
 }
 
 export const useTotalSupply = () => {
--- a/src/hooks/useTokenBalance.ts
+++ b/src/hooks/useTokenBalance.ts
@@ -7,32 +7,38 @@ import useWeb3 from './useWeb3'
 import useRefresh from './useRefresh'
 import useLastUpdated from './useLastUpdated'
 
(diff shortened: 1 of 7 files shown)


src/hooks/useTokenBalance.ts
```diff
@@ -8,24 +8,28 @@ import useRefresh from './useRefresh'
 import useLastUpdated from './useLastUpdated'
 
 const useTokenBalance = (tokenAddress: string) => {
-  const [balance, setBalance] = useState(BIG_ZERO)
+  const [balanceState, setBalanceState] = useState({ balance: BIG_ZERO, hasFetchedBalance: false })
```
hachiojidev, line 11: This is fine. However, I have found that it is more scalable and flexible if you add a status. The IFO for example uses
```
export type IfoStatus = 'idle' | 'coming_soon' | 'live' | 'finished'
```

src/hooks/useTokenBalance.ts
```diff
@@ -8,24 +8,28 @@ import useRefresh from './useRefresh'
...
+      }
     }
 
     if (account) {
       fetchBalance()
     }
   }, [account, tokenAddress, web3, fastRefresh])
 
-  return balance
+  return { ...balanceState }
```
hachiojidev, line 32: Why the spread?

src/hooks/useTokenBalance.ts
Chef-Cheems: Agree. That implies that there will be proper handling for `notFetched`, `isFetching`, `fetched` and `failed` statuses. Seems like we've managed so far to reasonably handle most of these states without explicit statuses so I'm not sure how necessary it is, but certainly this is how I'd expect it to be in "perfect" codebase.

src/hooks/useTokenBalance.ts
ChefHutch: Mystery. Removed 👍

src/hooks/useTokenBalance.ts
ChefHutch: Yea as, as Cheems said - we don't have an immediate use for these but it's much nicer.

Added!

src/hooks/useTokenBalance.ts
```diff
@@ -7,25 +7,42 @@ import useWeb3 from './useWeb3'
 import useRefresh from './useRefresh'
 import useLastUpdated from './useLastUpdated'
 
+export type FetchStatus = 'notFetched' | 'fetching' | 'success' | 'failed'
+
 const useTokenBalance = (tokenAddress: string) => {
-  const [balance, setBalance] = useState(BIG_ZERO)
+  const [balanceState, setBalanceState] = useState({
+    balance: BIG_ZERO,
+    fetchStatus: 'notFetched' as FetchStatus,
```
hachiojidev, line 15: If you are going to do this then just type the state

```
interface State {
  balance: BigNumber
  fetchStatus: FetchStatus
}
```
Then

```
useState<State>(...
```

hachiojidev (changes requested):

src/hooks/useTokenBalance.ts
```diff
@@ -7,25 +7,42 @@ import useWeb3 from './useWeb3'
 import useRefresh from './useRefresh'
 import useLastUpdated from './useLastUpdated'
 
+export type FetchStatus = 'notFetched' | 'fetching' | 'success' | 'failed'
```
hachiojidev, line 10: A gentle reminder - what I said about this pattern was just a _suggestion_. :)

(and, recently I have been using enums)

ChefHutch: I liked it! I think introducing these patterns is pretty light. I stripped out the 'fetching' state - as it has no explicit use atm and effectively re-triggers a 'loading' state every 6s for the users.

Enums ✅ 
Typed state ✅

src/hooks/useTokenBalance.ts
```diff
@@ -7,40 +7,46 @@ import useWeb3 from './useWeb3'
...
+
+export enum FetchStatus {
+  NOT_FETCHED = 'not-fetched',
+  SUCCESS = 'success',
+  FAILED = 'failed',
+}
 
 const useTokenBalance = (tokenAddress: string) => {
-  const [balanceState, setBalanceState] = useState({
+  const { NOT_FETCHED, SUCCESS, FAILED } = FetchStatus
```
hachiojidev, line 22: Not sure de structuring this is necessary inside the component but you do you.

hachiojidev (approved):

PR #1385, opened 2021-05-28
(styles) Add some initial new lottery styles

ChefHutch:
Changes that impact live pages:
- Tidy up `PageSection` component styles - used by the Trading Competition. All tested locally.

Changes that do not impact live pages:
- Initial SVGs/ style consts for lottery
- Tiny `TicketsNumberButton` comp

### DIFF
--- a/src/components/PageSection/index.tsx
+++ b/src/components/PageSection/index.tsx
@@ -15,6 +15,7 @@ interface BackgroundColorProps extends FlexProps {
   index: number
   background?: string
   hasZeroMargin?: boolean
+  padding?: string
 }
 
 const BackgroundColor = styled(Flex)<BackgroundColorProps>`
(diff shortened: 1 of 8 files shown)


src/components/PageSection/index.tsx
```diff
@@ -15,20 +15,21 @@ interface BackgroundColorProps extends FlexProps {
   index: number
   background?: string
   hasZeroMargin?: boolean
+  padding?: string
```
hachiojidev, line 18: Can we extend this component with `<Box />`?

src/components/PageSection/index.tsx
```diff
@@ -70,7 +89,13 @@ const PageSection: React.FC<PageSectionProps> = ({
           dividerComponent={dividerComponent}
         />
       )}
-      <BackgroundColor background={background} index={index} hasZeroMargin={hasZeroMargin} {...props}>
+      <BackgroundColor
+        background={background}
+        index={index}
+        padding={getPadding()}
+        hasZeroMargin={hasZeroMargin}
```
hachiojidev, line 96: Just a thought. This prop is oddly specific. Would it be cleaner to pass in the margin and calculate whether it needs it in the parent component?

src/components/PageSection/index.tsx
```diff
@@ -60,6 +61,24 @@ const PageSection: React.FC<PageSectionProps> = ({
...
   const hasZeroMargin = !hasCurvedDivider || curvePosition === 'top'
 
+  const getPadding = () => {
+    // No curved divider
+    if (!hasCurvedDivider) {
+      return '48px 0'
+    }
+    // Bottom curved divider
+    // Less bottom padding, as the divider is present there
+    if (curvePosition === 'bottom') {
```
Chef-Cheems, line 71: I suppose it is this one? If yes do we need the top version? Don't see similar thing on top in Figma. Unless top version is just in case.
Screen Shot 2021-05-31 at 9 57 00 AM

ChefHutch: Yea the component is shared with the trading competition, which has a top version:

Screenshot 2021-05-31 at 08 34 43

Also soon to be used by the new home page - although no top version in those designs.

src/components/PageSection/index.tsx
ChefHutch: Hmmn this was a useful pause for me to reflect on exactly why this negative top & bottom margins are needed. It's basically 'pulling' two page sections into one another, so that they overlap at the 'curved' SVG, giving the layered effect. The curved section's height changes as the page scales, hence why the margin has a few breakpoints- getting larger as the screen gets bigger. 

I think giving the parent control of the margin will massively clean up the components' margins, which are frankly.. mental. 

Let me make the change here and see what you think.

src/components/PageSection/index.tsx
ChefHutch: Yea no need for padding being passed to it as a prop - passed `p = getPadding()` to the component instead. It needs some of the flex properties though, so extended using that.

Chef-Cheems (approved):

hachiojidev (approved):

PR #1397, opened 2021-05-31
fix: (pools) Add missing translations in pools table view

memoyil:
Issues: 

1. Total staked column name in pools row
2. Sort by items in dropdown menu
3. Placeholder text in search input

### DIFF
--- a/src/views/Pools/index.tsx
+++ b/src/views/Pools/index.tsx
@@ -243,19 +243,19 @@ const Pools: React.FC = () => {
                 <Select
                   options={[
                     {
-                      label: 'Hot',
+                      label: t('Hot'),
                       value: 'hot',
                     },
                     {
-                      label: 'APR',
+                      label: t('APR'),
                       value: 'apr',
                     },
                     {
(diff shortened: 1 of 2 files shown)


src/views/Pools/index.tsx
```diff
@@ -268,7 +268,7 @@ const Pools: React.FC = () => {
                 {t('Search')}
               </Text>
               <ControlStretch>
-                <SearchInput onChange={handleChangeSearchQuery} placeholder="Search pools" />
+                <SearchInput onChange={handleChangeSearchQuery} placeholder={t('Search Pools')} />
```
Chef-Cheems, line 271: This is actually translated inside SearchInput. There is `t(placeholder)` in there. I'm not sure which way is better though 😄 
If you want you can refactor SearchInput to accept already translated string + also change SearchInput in Farms.tsx. But I think it's not necessary.

memoyil: I removed translation only the translation key is fixed

hachiojidev (approved):

PR #1400, opened 2021-05-31
fix(fetch-public-farm-data): Increase timeout on farmFetch to 9s

ChefHutch:
- Increase time before dropping farm fetch requests from 5s to 9s
- Add multiplier to return type

### DIFF
--- a/src/state/farms/fetchPublicFarmData.ts
+++ b/src/state/farms/fetchPublicFarmData.ts
@@ -111,7 +111,7 @@ const fetchFarm = async (farm: Farm): Promise<PublicFarmData> => {
     const id = setTimeout(() => {
       clearTimeout(id)
       resolve({})
-    }, 5000)
+    }, 9000)
   })
 
   return Promise.race([timeout, farmFetch()])
--- a/src/state/farms/fetchPublicFarmData.ts
+++ b/src/state/farms/fetchPublicFarmData.ts
@@ -15,6 +15,7 @@ type PublicFarmData = {
   lpTotalSupply?: SerializedBigNumber
   tokenPriceVsQuote?: SerializedBigNumber
   poolWeight?: SerializedBigNumber
+  multiplier?: string
 }
 
 const fetchFarm = async (farm: Farm): Promise<PublicFarmData> => {


Chef-Cheems (approved):

memoyil:
Will user have to wait 9 seconds to get the already fetched farm details? If we keep it at 5 (since we are using fast refresh which is 10sec interval) if one farms gets stuck user will only wait 5 seconds and stuck one will be shown in the next cycle which is after 5 seconds.

Chef-Cheems:
@memoyil the farm fetching interval was reduced to slowRefresh (60s) last week (except CAKE - BNB and BUSD - BNB farms). Seems that on weekend there was some issue with farms loading, presumably since most farms took more than 5 sec to load. Hence this PR 🙂

memoyil:
> @memoyil the farm fetching interval was reduced to slowRefresh (60s) last week (except CAKE - BNB and BUSD - BNB farms). Seems that on weekend there was some issue with farms loading, presumably since most farms took more than 5 sec to load. Hence this PR 🙂

What will happen if it takes more than 9 sec? Should we add kind of dynamic variable there? I mean if we remove the timeout we have to wait until the last farm data is fetched which is also not good.

ChefHutch:
So yea Memoyil, just to reiterate what @Chef-Cheems said. I've done some digging and I think our initial diagnosis is right: the timeout of 5s was just too short for many users to complete the request and so the timeout was resolving the promise instead. It would then only attempt the fetch the data again after a slowRefresh.

> What will happen if it takes more than 9 sec? Should we add kind of dynamic variable there? I mean if we remove the timeout we have to wait until the last farm data is fetched which is also not good.

At the moment if 9s passes and the farm hasn't successfully fetched the data - it'll resolve without any of the data being fetched by the smart contract. Let's see what impact this increase has. As you say - the point of the timeout is to prevent the entire farms promise being held up by 1 farm that has failed to fetch. But, we can't have the timeout be so tight that it's _forcing_ farm requests to skip that would otherwise succeed.

If we need to (as is what I think you're suggesting), we could pass a 'timeout' variable into the function - which could be higher for the few farms on a `fastRefresh`, and longer for the majority of farms on a `slowRefresh`.

src/state/farms/fetchPublicFarmData.ts
```diff
@@ -15,6 +15,7 @@ type PublicFarmData = {
   lpTotalSupply?: SerializedBigNumber
   tokenPriceVsQuote?: SerializedBigNumber
   poolWeight?: SerializedBigNumber
+  multiplier?: string
```
memoyil, line 18: what is this :D

Chef-Cheems: it is returned here but was missing in type, I guess @ChefHutch just fixed it while he's changing this file 🙂

ChefKai (approved):

src/state/farms/fetchPublicFarmData.ts
memoyil: Ah yes true, I thought it is related to timeout thing

PR #1415, opened 2021-06-02
Feature/gas price setting

hachiojidev:


src/utils/contractHelpers.ts
```diff
@@ -50,10 +50,14 @@ import easterNftAbi from 'config/abi/easterNft.json'
 import cakeVaultAbi from 'config/abi/cakeVault.json'
 import predictionsAbi from 'config/abi/predictions.json'
 import chainlinkOracleAbi from 'config/abi/chainlinkOracle.json'
+import { DEFAULT_GAS_PRICE } from 'config'
+import { getSettings } from './settings'
 
-const getContract = (abi: any, address: string, web3?: Web3) => {
+const getContract = (abi: any, address: string, web3?: Web3, account?: string) => {
```
hachiojidev, line 56: I opted for passing in the account because getting the user settings from the web3 object is

1. Not guaranteed to be there
2. Would require (as far as I know) for the function to be async. We would then have to update the entire code base

src/utils/settings.ts
```diff
@@ -0,0 +1,39 @@
...
+export interface UserSettings {
+  gasPrice: string
+}
+
+export const VERSION = 1
+
+export const GAS_SETTINGS = {
+  default: DEFAULT_GAS_PRICE, // 5
+  fast: '10000000000', // 10
+  reallyfast: '15000000000', // 15
```
hachiojidev, line 12: These are just initial values. I am sure they will be modified once we flesh out the requirements.

src/utils/settings.ts
```diff
@@ -0,0 +1,39 @@
...
+  gasPrice: GAS_SETTINGS.default,
+})
+
+export const getStorageKey = (account: string) => {
+  return `pancakeswap_settings_${account}_${VERSION}`
+}
+
+export const getSettings = (account: string): UserSettings => {
+  try {
+    const settingsFromLs = localStorage.getItem(getStorageKey(account))
```
hachiojidev, line 25: Settings won't change often so I see no reason to put them in another state. This also allows for the interface to be framework agnostic.

src/config/index.ts
```diff
@@ -28,3 +28,4 @@ export const LOTTERY_MAX_NUMBER_OF_TICKETS = 50
 export const LOTTERY_TICKET_PRICE = 1
 export const DEFAULT_TOKEN_DECIMAL = BIG_TEN.pow(18)
 export const DEFAULT_GAS_LIMIT = 200000
+export const DEFAULT_GAS_PRICE = '5000000000'
```
ChefKai, line 31:
```suggestion
export const DEFAULT_GAS_PRICE = '5000000000'

export const DEFAULT_GAS_PRICE = parseUnits('5', 'gwei')
```

Readability would be enhanced using utils from `ethers` to format gasPrice directly

src/utils/settings.ts
```diff
@@ -0,0 +1,39 @@
...
+    return settingsFromLs ? JSON.parse(settingsFromLs) : getDefaultSettings()
+  } catch (error) {
+    return getDefaultSettings()
+  }
+}
+
+export const setSettings = (account: string, newSettings: UserSettings) => {
+  localStorage.setItem(getStorageKey(account), JSON.stringify(newSettings))
```
ChefHutch, line 33: The marginal naming difference between `setSetting` & `setSettings` rustles my jimmies a bit.

Does this need to be broken into two functions? If yes maybe this one could be renamed `setSettings` > `storeSettingToLs`?

ChefHutch (approved):

src/config/index.ts
hachiojidev: A note on this - I tried using `parseUnits` and Typescript started reporting a bunch of errors. I wonder if there is a version mismatch or ethers does not support a certain version of Bignumber.

src/utils/settings.ts
hachiojidev: You know, I agree with your jimmies in this case. I removed `setSetting` all together. `setSettings` can easily do the work of both.

RabbitDoge (approved):

PR #1417, opened 2021-06-02
chore(fetch-public-farm-data): Remove race promise timeout from fetchFarm function

ChefHutch:
The race condition for the farm fetch was causing more problems than it was solving. The impact for a lot of users was that the farms never fetched within the timeout, and this has the knock-on impact of meaning price data (used for APRs throughout the site) was missing.

This PR removes the race condition between the farmFetch and the timeout

https://lucid-haibt-de8af4.netlify.app/

### DIFF
--- a/src/state/farms/fetchPublicFarmData.ts
+++ b/src/state/farms/fetchPublicFarmData.ts
@@ -7,115 +7,103 @@ import multicall from 'utils/multicall'
 import { Farm, SerializedBigNumber } from '../types'
 
 type PublicFarmData = {
-  tokenAmountMc?: SerializedBigNumber
-  quoteTokenAmountMc?: SerializedBigNumber
-  tokenAmountTotal?: SerializedBigNumber
-  quoteTokenAmountTotal?: SerializedBigNumber
-  lpTotalInQuoteToken?: SerializedBigNumber
-  lpTotalSupply?: SerializedBigNumber
-  tokenPriceVsQuote?: SerializedBigNumber
-  poolWeight?: SerializedBigNumber
-  multiplier?: string
+  tokenAmountMc: SerializedBigNumber
+  quoteTokenAmountMc: SerializedBigNumber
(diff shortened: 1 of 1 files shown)


src/state/farms/fetchPublicFarmData.ts
```diff
@@ -7,115 +7,103 @@ import multicall from 'utils/multicall'
...
+  lpTotalSupply: SerializedBigNumber
+  tokenPriceVsQuote: SerializedBigNumber
+  poolWeight: SerializedBigNumber
+  multiplier: string
 }
 
 const fetchFarm = async (farm: Farm): Promise<PublicFarmData> => {
   const { pid, lpAddresses, token, quoteToken } = farm
   const lpAddress = getAddress(lpAddresses)
-  const farmFetch = async () => {
```
ChefHutch, line 24: Hyperactive diff because spacing within the function has changed. 

THIS line declaring the async `farmFetch` was deleted

src/state/farms/fetchPublicFarmData.ts
```diff
@@ -7,115 +7,103 @@ import multicall from 'utils/multicall'
...
-
-  // In some browsers promise above gets stuck that causes fetchFarms to not proceed.
-  const timeout = new Promise((resolve) => {
-    const id = setTimeout(() => {
-      clearTimeout(id)
-      resolve({})
-    }, 9000)
-  })
-
-  return Promise.race([timeout, farmFetch()])
```
ChefHutch, line 118: ...and these lines were deleted

hachiojidev:
> Can you check, is there a stuck issues in firefox after removing this race? That browser where i saw most

@memoyil 

We had a chat about it and decided since FF is low usage (less than 5%) we will remove this optimization for now. We want to go back to the drawing board on this one as we are getting quite a few reports from non-FF users that the APR is not loading for them.

hachiojidev (approved):

PR #1420, opened 2021-06-02
chore: Introduce big two

memoyil:
To review: 

https://deploy-preview-1420--pancakeswap-dev.netlify.app/

### DIFF
--- a/src/state/farms/fetchPublicFarmData.ts
+++ b/src/state/farms/fetchPublicFarmData.ts
@@ -2,7 +2,7 @@ import BigNumber from 'bignumber.js'
 import masterchefABI from 'config/abi/masterchef.json'
 import erc20 from 'config/abi/erc20.json'
 import { getAddress, getMasterChefAddress } from 'utils/addressHelpers'
-import { BIG_TEN, BIG_ZERO } from 'utils/bigNumber'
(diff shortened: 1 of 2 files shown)


ChefHutch:
I don't think we need to wrap every BN num value in a constant @memoyil, particularly not those that are barely used in the code base. 

Not sure what the end point is if we do decide to assign BigNumber(2) to a constant. Well.. I do have one vision for how this all ends, and it ain't pretty... `BIG_TWO_THOUSAND_TWO_HUNDRED_AND_EIGHTY_FOUR`

memoyil:
> I don't think we need to wrap every BN num value in a constant @memoyil, particularly not those that are barely used in the code base?
> 
> Not sure what the end point is if we do decide to assign BigNumber(2) to a constant. Well.. I do have one vision for how this all ends, and it ain't pretty... `BIG_TWO_THOUSAND_TWO_HUNDRED_AND_EIGHTY_FOUR` 😅

In java for example bigdecimal object has predefined cache from 0 to 10. I add them one by one when we have a usage. But we should not make constant for all bignumber we use of course

hachiojidev:
@memoyil I am not sure we need any more set constants like this. I see the use case for 0, 1, and 10 but I can see no benefit for other numbers.

PR #1424, opened 2021-06-02
feat(approve-confirm-buttons): Allow different visual arrangement of ApproveConfirmButtons. Add onApproveSuccess func

ChefHutch:
Two changes:

*One*
- Enable the `ApproveConfirmButtons` to receive a `buttonArrangement` prop. This determines whether the approve & confirm buttons should be laid out in a row (the current style, set as the component default), or 'sequentially' (i.e. one after the other), as is a common pattern in other areas of the site (and will be used immediately within the Lottery).

```
<ApproveConfirmButtons
 isApproveDisabled={isApproved}
 ...
 buttonArrangement={ButtonArrangement.SEQUENTIAL}
/>
```

`isApproved` is false, therefore the approval button is rendered:
Screenshot 2021-06-02 at 16 22 23

`isApproved` is true, therefore the confirm button is rendered:
Screenshot 2021-06-02 at 16 22 42

*Two*
- Call an `onApproveSuccess()` function when there is a transaction receipt from the `handleApprove` function within the `useApproveConfirmTransaction` hook. This allows the triggering of toasts when a user has successfully confirmed a transaction. Again, this will be used immediately within the new Lottery.

### DIFF
--- a/src/hooks/useApproveConfirmTransaction.ts
+++ b/src/hooks/useApproveConfirmTransaction.ts
@@ -88,6 +88,7 @@ interface ApproveConfirmTransaction {
   onConfirm: ContractHandler
   onRequiresApproval?: () => Promise<boolean>
   onSuccess: (state: State) => void
+  onApproveSuccess?: (state: State) => void
 }
 
 const useApproveConfirmTransaction = ({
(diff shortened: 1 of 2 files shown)


src/hooks/useApproveConfirmTransaction.ts
```diff
@@ -129,6 +131,7 @@ const useApproveConfirmTransaction = ({
         })
         .on('receipt', (payload: Web3Payload) => {
           dispatch({ type: 'approve_receipt', payload })
+          onApproveSuccess(state)
```
ChefHutch, line 134: Note sure if `state` would ever be useful within the function tbh, I just mirrored the existing `onSuccess` func.

hachiojidev (approved):

PR #1439, opened 2021-06-04
fix: (pools) Bounty modal shows obsolete values when cake vault data changes

memoyil:
To review: 

https://deploy-preview-1439--pancakeswap-dev.netlify.app/

To reproduce the issue:

1. Check the difference between data shown in pools page and bounty modal when it is open for a while

Before:

https://user-images.githubusercontent.com/2213635/120809187-1c6ba300-c54a-11eb-9354-10551b7db37b.mov

After:

https://user-images.githubusercontent.com/2213635/120809175-1aa1df80-c54a-11eb-9bb0-ba77b73b1a69.mov

### DIFF
--- a/src/views/Pools/components/BountyModal.tsx
+++ b/src/views/Pools/components/BountyModal.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react'
+import React, { useMemo, useState } from 'react'
 import BigNumber from 'bignumber.js'
 import { useWeb3React } from '@web3-react/core'
 import { DEFAULT_GAS_LIMIT } from 'config'
@@ -11,12 +11,9 @@ import useToast from 'hooks/useToast'
 import { useTranslation } from 'contexts/Localization'
 import UnlockButton from 'components/UnlockButton'
 import Balance from 'components/Balance'
+import { useCakeVault, usePriceCakeBusd } from 'state/hooks'
 
 interface BountyModalProps {
-  cakeBountyToDisplay: number
-  dollarBountyToDisplay: number
-  totalPendingCakeHarvest: BigNumber
-  callFee: number
   onDismiss?: () => void
   TooltipComponent: React.ElementType
 }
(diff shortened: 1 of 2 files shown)


src/views/Pools/components/BountyModal.tsx
```diff
@@ -28,23 +25,32 @@ const Divider = styled.div`
...
+  const estimatedDollarBountyReward = useMemo(() => {
+    return new BigNumber(estimatedCakeBountyReward).multipliedBy(cakePriceBusd)
+  }, [cakePriceBusd, estimatedCakeBountyReward])
+
+  const hasFetchedDollarBounty = estimatedDollarBountyReward.gte(0)
+  const hasFetchedCakeBounty = estimatedCakeBountyReward ? estimatedCakeBountyReward.gte(0) : false
+  const dollarBountyToDisplay = hasFetchedDollarBounty ? getBalanceNumber(estimatedDollarBountyReward, 18) : 0
+  const cakeBountyToDisplay = hasFetchedCakeBounty ? getBalanceNumber(estimatedCakeBountyReward, 18) : 0
+
+  const { targetRef, tooltip, tooltipVisible } = useTooltip(<TooltipComponent fee={1000} />, {
```
ChefHutch, line 53: `fee={1000}` should probably be using `callFee`?

src/views/Pools/components/BountyModal.tsx
```diff
@@ -28,23 +25,32 @@ const Divider = styled.div`
...
-  TooltipComponent,
-}) => {
+const BountyModal: React.FC<BountyModalProps> = ({ onDismiss, TooltipComponent }) => {
   const { t } = useTranslation()
   const { account } = useWeb3React()
   const { theme } = useTheme()
   const { toastError, toastSuccess } = useToast()
   const cakeVaultContract = useCakeVaultContract()
-  const [pendingTx, setPendingTx] = useState(false)
+  const [pendingTx, setPendingTx] = useState(true)
```
ChefHutch, line 34: Should this now be defaulting to true?

ChefHutch (commented):
Thanks Memoyil. Few things to address in comments.

Slightly inelegant to have so much copied & pasted code between a parent and child component, but with the modal's current architecture there will need to be some duplication, even if some code could be abstracted to helpers. Did you take a look at a modal-wide solution for this?

src/views/Pools/components/BountyModal.tsx
memoyil: Sorry it was a left over from testing, fixed now :)

src/views/Pools/components/BountyModal.tsx
memoyil: Sorry it was a left over from testing, fixed now :)

memoyil:
> Thanks Memoyil. Few things to address in comments.
> 
> Slightly inelegant to have so much copied & pasted code between a parent and child component, but with the modal's current architecture there will need to be some duplication, even if some code could be abstracted to helpers. Did you take a look at a modal-wide solution for this?

I applied all the possible fixes that came in to my mind (such as using reference instead of direct value), but it didn't work. :(

ChefHutch (approved):
Thanks for addressing the comments Memoyil, and exploring alternatives for ways to get dynamic state into the modal. I'd be interested to see if @hachiojidev has any thoughts on how we might be able to have modals receive dynamic prop values and re-render accordingly, but this is a reasonable short-term fix for the bounty card issue.

hachiojidev (approved):

PR #1446, opened 2021-06-06
fix: (farms) Error when put amount with decimals more than 18

memoyil:
To review:

https://deploy-preview-1446--pancakeswap-dev.netlify.app/

To Reproduce the issue:

1. Go to farms
2. Click withdraw on any farm that you staked
3. Put amount with more than 20 decimals
4. Check error on console

### DIFF
--- a/src/components/ModalInput/ModalInput.tsx
+++ b/src/components/ModalInput/ModalInput.tsx
@@ -12,6 +12,7 @@ interface ModalInputProps {
   value: string
   addLiquidityUrl?: string
   inputTitle?: string
+  decimals?: number
 }
 
 const getBoxShadow = ({ isWarning = false, theme }) => {
@@ -64,6 +65,7 @@ const ModalInput: React.FC<ModalInputProps> = ({
   value,
   addLiquidityUrl,
   inputTitle,
+  decimals = 18,
 }) => {
   const { t } = useTranslation()
   const isBalanceZero = max === '0' || !max
@@ -88,7 +90,7 @@ const ModalInput: React.FC<ModalInputProps> = ({
         </Flex>
         <Flex alignItems="flex-end" justifyContent="space-around">
           <StyledInput
-            pattern="^[0-9]*[.,]?[0-9]*$"
+            pattern={`^[0-9]*[.,]?[0-9]{0,${decimals}}$`}
             inputMode="decimal"
             step="any"
             min="0"


src/components/ModalInput/ModalInput.tsx
```diff
@@ -64,6 +65,7 @@ const ModalInput: React.FC<ModalInputProps> = ({
   value,
   addLiquidityUrl,
   inputTitle,
+  decimals = 20,
```
hachiojidev, line 68: Should be 18

src/components/ModalInput/ModalInput.tsx
```diff
@@ -88,7 +90,7 @@ const ModalInput: React.FC<ModalInputProps> = ({
         </Flex>
         <Flex alignItems="flex-end" justifyContent="space-around">
           <StyledInput
-            pattern="^[0-9]*[.,]?[0-9]*$"
+            pattern={`^[0-9]*[.,]?[0-9]{0,${decimals}}$`}
```
hachiojidev, line 93: This pattern (and decimal prop) should be updated on the UI Kit and removed from here.

hachiojidev (changes requested):

src/components/ModalInput/ModalInput.tsx
memoyil: done

src/components/ModalInput/ModalInput.tsx
memoyil: I opened a pr for BalanceInput in UI Kit

https://github.com/pancakeswap/pancake-toolkit/pull/142

memoyil: Input component is used in many places for different reasons in frontend (such as search input for text, lottery search for integer and transfer nft for wallet address), if we move this pattern to uikit we need to refactor all of the usages and write rules for each of them.

hachiojidev: Yes, good catch.

hachiojidev: Wait, you added it to the BalanceInput which is fine.

memoyil: I think it is ready to merge wdyt?

hachiojidev (approved):

PR #1450, opened 2021-06-07
fix(ifo): Decimal formatting on contribute modal

hachiojidev:


src/views/Ifos/components/IfoFoldableCard/IfoPoolCard/ContributeModal.tsx
```diff
@@ -93,7 +93,7 @@ const ContributeModal: React.FC<Props> = ({
         {limitPerUserInLP.isGreaterThan(0) && (
           <Flex justifyContent="space-between" mb="16px">
             <Text>{t('Max. LP token entry')}</Text>
-            <Text>{getBalanceNumber(limitPerUserInLP, currency.decimals)}</Text>
+            <Text>{getBalanceAmount(limitPerUserInLP, currency.decimals).toString()}</Text>
```
hachiojidev, line 96: `getBalanceNumber` is calling `toNumber()` which results in rounding errors.

hachiojidev: We should get rid of `getBalanceNumber` in fact. It doesn't really do anything but call `toNumber()` which we should avoid in most cases.

RabbitDoge (approved):

PR #1454, opened 2021-06-07
fix: (predictions) Use bignumber in set position instead of number to avoid precision errors

memoyil:
To review: 

https://deploy-preview-1454--pancakeswap-dev.netlify.app/

To Reproduce the issue:

1. Go to prediction
2. Set position from card
3. Click max 
4. Check diff between your bnb balance and max (should be 0.01)

Before: 

Screenshot 2021-06-07 at 11 46 15

After:

Screenshot 2021-06-07 at 11 45 08

### DIFF
--- a/src/views/Predictions/components/RoundCard/SetPositionCard.tsx
+++ b/src/views/Predictions/components/RoundCard/SetPositionCard.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react'
+import React, { useEffect, useMemo, useState } from 'react'
 import {
   ArrowBackIcon,
   CardBody,
@@ -47,24 +47,15 @@ const gasPrice = new BigNumber(6).times(BIG_TEN.pow(BIG_NINE)).toString()
 const dust = new BigNumber(0.01).times(DEFAULT_TOKEN_DECIMAL)
 const percentShortcuts = [10, 25, 50, 75]
 
-const getPercentDisplay = (percentage: number) => {
-  if (Number.isNaN(percentage)) {
-    return ''
+const getButtonProps = (value: BigNumber, bnbBalance: BigNumber, minBetAmountBalance: BigNumber) => {
+  const hasSufficientBalance = () => {
+    if (value.gt(0)) {
+      return value.lte(bnbBalance)
+    }
+    return bnbBalance.gt(0)
   }
 
-  if (percentage > 100) {
-    return ''
-  }
-
-  if (percentage < 0) {
-    return ''
-  }
-
-  return `${percentage.toLocaleString(undefined, { maximumFractionDigits: 1 })}%`
-}
-
-const getButtonProps = (value: BigNumber, bnbBalance: BigNumber, minBetAmountBalance: number) => {
-  if (bnbBalance.eq(0)) {
+  if (!hasSufficientBalance()) {
     return { key: 'Insufficient BNB balance', disabled: true }
   }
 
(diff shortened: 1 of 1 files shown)


src/views/Predictions/components/RoundCard/SetPositionCard.tsx
```diff
@@ -86,30 +72,46 @@ const SetPositionCard: React.FC<SetPositionCardProps> = ({ position, togglePosit
...
+  const balanceDisplay = useMemo(() => {
+    return getBnbAmount(bnbBalance).toString()
+  }, [bnbBalance])
+  const maxBalance = useMemo(() => {
+    return getBnbAmount(bnbBalance.gt(dust) ? bnbBalance.minus(dust) : bnbBalance)
+  }, [bnbBalance])
+  const minBetAmountBalance = useMemo(() => {
+    return getBnbAmount(minBetAmount)
+  }, [minBetAmount])
+
```
hachiojidev, line 84: I don't see any reason why these need to be memoized. They are simply calling a pure function to format BNB.

src/views/Predictions/components/RoundCard/SetPositionCard.tsx
```diff
@@ -86,30 +72,46 @@ const SetPositionCard: React.FC<SetPositionCardProps> = ({ position, togglePosit
...
 
-  const percentageOfMaxBalance = valueAsBn.div(maxBalance).times(100).toNumber()
-  const percentageDisplay = getPercentDisplay(percentageOfMaxBalance)
   const showFieldWarning = account && valueAsBn.gt(0) && errorMessage !== null
-  const minBetAmountBalance = getBnbAmount(minBetAmount).toNumber()
 
-  const handleChange = (input) => {
-    setValue(input)
-  }
+  const [percent, setPercent] = useState(0)
```
hachiojidev, line 89: Why keep percent in state? It is easily derived from from the current value and the user's balance.

src/views/Predictions/components/RoundCard/SetPositionCard.tsx
```diff
@@ -86,30 +72,46 @@ const SetPositionCard: React.FC<SetPositionCardProps> = ({ position, togglePosit
...
 
-  const handleSliderChange = (newValue: number) => {
-    setValue(newValue.toString())
+  const handleInputChange = (input: string) => {
+    if (input) {
+      const percentage = Math.floor(new BigNumber(input).dividedBy(maxBalance).multipliedBy(100).toNumber())
+      setPercent(Math.min(percentage, 100))
```
hachiojidev, line 94: Why `Math.floor` why `Math.min`?

hachiojidev (changes requested):
I have to be honest I am a little confused with this PR. Quite a lot of changes to fix the precision errors.

On line 90 we have
```
const maxBalance = getBnbAmount(bnbBalance.gt(dust) ? bnbBalance.minus(dust) : bnbBalance).toNumber()
```
Isn't it as simple as removing `toNumber()`? `maxBalance` is not used anywhere as a number and never needed to be converted.

Similarly, `toNumber()` on line 96 can be removed.

I could be missing something but can you confirm.

src/views/Predictions/components/RoundCard/SetPositionCard.tsx
memoyil: Those i added not to calculate every time when slider is moving

src/views/Predictions/components/RoundCard/SetPositionCard.tsx
memoyil: Min is for when user enters value that exceeds max balance, floor is for not to show too many decimals in percentage

src/views/Predictions/components/RoundCard/SetPositionCard.tsx
```diff
@@ -47,24 +47,10 @@ const gasPrice = new BigNumber(6).times(BIG_TEN.pow(BIG_NINE)).toString()
...
-  }
-
-  return `${percentage.toLocaleString(undefined, { maximumFractionDigits: 1 })}%`
-}
-
-const getButtonProps = (value: BigNumber, bnbBalance: BigNumber, minBetAmountBalance: number) => {
-  if (bnbBalance.eq(0)) {
+const getButtonProps = (value: BigNumber, bnbBalance: BigNumber, minBetAmountBalance: BigNumber) => {
+  const hasSufficientBalance =
+    (value.gt(0) && value.lte(bnbBalance)) || ((value.eq(0) || value.isNaN()) && bnbBalance.gt(0))
```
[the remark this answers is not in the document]
memoyil: done

memoyil:
> I have to be honest I am a little confused with this PR. Quite a lot of changes to fix the precision errors.
> 
> On line 90 we have
> 
> ```
> const maxBalance = getBnbAmount(bnbBalance.gt(dust) ? bnbBalance.minus(dust) : bnbBalance).toNumber()
> ```
> 
> Isn't it as simple as removing `toNumber()`? `maxBalance` is not used anywhere as a number and never needed to be converted.
> 
> Similarly, `toNumber()` on line 96 can be removed.
> 
> I could be missing something but can you confirm.

Main approach on this pr is (as you remember there is a bug in input type range when there is too much decimals in value where we had to limit step with 15 decimals) changing the range that we use it in slider to 0 to 100. 

With this pr we will not have 15 decimals when user uses slider, we will have more precise output (bet value). That's why slider starts from 0 and ends 100 instead of bnb max value (is needed in number according to slider props btw if we stick with the old approach). Then we will calculate the bet value what we get from slider multiplying with max balance amount.

src/views/Predictions/components/RoundCard/SetPositionCard.tsx
memoyil: see my comment https://github.com/pancakeswap/pancake-frontend/pull/1454#issuecomment-857971554

hachiojidev (approved):

PR #1476, opened 2021-06-09
fix: (translations) Use locale upper/lower case

memoyil:
Although https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform css covers locale upper/lower case, It only does that when OS language(browser language too) and the translation language are the same.

Reproduce one of the issue of many:

1. Go to team battle
2. Change language to turkish
3. Check comptation phases top right (Entry - Live - End in english)
4. First text should be GİRİŞ instead of GIRIŞ

Next one
1. Go to prediction
2. Change language to turkish
3. Check Last Price translation in cards 
4. Text should be SON FİYAT instead of SON FIYAT (see the difference in I character)

Pr is mainly for Turkish language 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLocaleLowerCase

but In other languages we might end up with a different word when we use upper case with localization.

### DIFF
--- a/src/components/ApyCalculatorModal/index.tsx
+++ b/src/components/ApyCalculatorModal/index.tsx
@@ -40,7 +40,7 @@ const ApyCalculatorModal: React.FC<ApyCalculatorModalProps> = ({
   compoundFrequency = 1,
   performanceFee = 0,
 }) => {
-  const { t } = useTranslation()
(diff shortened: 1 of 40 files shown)


memoyil:
It contains lots of changes with very subtle results

PR #1478, opened 2021-06-09
feat: Close WalletConnect session on logout

Chef-Cheems:


src/hooks/useAuth.ts
```diff
@@ -61,6 +61,9 @@ const useAuth = () => {
   const logout = useCallback(() => {
     dispatch(profileClear())
     deactivate()
+    if (connectorsByName.walletconnect) {
```
memoyil, line 64: I think this returns you a constant string not the current connector

memoyil:
I have one pending pr in uikit https://github.com/pancakeswap/pancake-toolkit/pull/78
But your solution is much better and concise.

src/hooks/useAuth.ts
memoyil: One thing you can do is 

if (window.localStorage.getItem(ConnectorNames.WalletConnect)) {

Chef-Cheems: yeah it does, this is more of a check if its initialized or not.
I'm thinking of getting current connector but haven't looked yet how to do it without modifying UIKit. For now mostly made PR for preview link :)

Chef-Cheems:
@memoyil oh so you did make a PR for that, I thought I've seen something but couldn't find it 😄  Will check it out. I see that you manually remove the localStorage item for wallet connect, `close()` does this but also I think signals TrustWallet that connection was closed

memoyil:
> @memoyil oh so you did make a PR for that, I thought I've seen something but couldn't find it 😄 Will check it out. I see that you manually remove the localStorage item for wallet connect, `close()` does this but also I think signals TrustWallet that connection was closed

Yes, and it also closes the websocket connection, I was doing with restarting the webpage :). Very good!

hachiojidev (approved):

PR #1486, opened 2021-06-10
fix(contract-helpers): Add getGasPrice helper and chainid consts to set gas based on chainid

ChefHutch:
TestNet gas price should be 10 gwei, rather than the default 5.
- Add a getDefaultGasPrice helper to manage this.
- Add chainID consts to `/config`

### DIFF
--- a/src/utils/contractHelpers.ts
+++ b/src/utils/contractHelpers.ts
@@ -53,22 +53,20 @@ import chainlinkOracleAbi from 'config/abi/chainlinkOracle.json'
 import { DEFAULT_GAS_PRICE, TESTNET_CHAIN_ID } from 'config'
 import { getSettings, getGasPriceInWei } from './settings'
 
-export const getGasPrice = (account?: string) => {
+export const getDefaultGasPrice = () => {
   const chainId = process.env.REACT_APP_CHAIN_ID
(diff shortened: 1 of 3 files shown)


src/utils/contractHelpers.ts
```diff
@@ -50,15 +50,25 @@ import easterNftAbi from 'config/abi/easterNft.json'
...
 import { getSettings, getGasPriceInWei } from './settings'
 
+export const getGasPrice = (account?: string) => {
+  const chainId = process.env.REACT_APP_CHAIN_ID
+  if (chainId === TESTNET_CHAIN_ID) {
+    return 10
+  }
+  if (account) {
+    return getSettings(account).gasPrice
+  }
```
hachiojidev, line 63: I think you should keep this function pure (insofar as the account has nothing to do with the "config" of gas price). Adding the account context locks it into the front end app.

hachiojidev: Okay, maybe not "locks" but you get the idea.

ChefHutch: Yea sure I think I get you, I've changed the function to be explicitly getting chain defaults, although these will only be used in the absence of user settings

hachiojidev (approved):

PR #1511, opened 2021-06-15
feat: (earnaprcard) Lazy load farms data to reduce amount of requests

memoyil:
This feature is mainly for mobiles (where earn apr card is not visible at first sight), if user does not scroll down after opening the page, requests drops to 27 from around 600 at opening.

### DIFF
--- a/src/hooks/useIntersectionObserver.ts
+++ b/src/hooks/useIntersectionObserver.ts
@@ -1,0 +1,26 @@
+import { useEffect, useRef, useState } from 'react'
+
+const useIntersectionObserver = () => {
+  const observerRef = useRef<HTMLDivElement>(null)
+  const [observerIsSet, setObserverIsSet] = useState(false)
+  const [isIntersecting, setIsIntersecting] = useState(false)
+
+  useEffect(() => {
+    const checkObserverIsIntersecting = ([entry]: IntersectionObserverEntry[]) => {
+      setIsIntersecting(entry.isIntersecting)
+    }
+
+    if (!observerIsSet) {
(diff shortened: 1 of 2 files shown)


src/views/Home/components/EarnAPRCard.tsx
```diff
@@ -80,7 +84,10 @@ const EarnAPRCard = () => {
             {highestApr && !isFetchingFarmData ? (
               `${highestApr}%`
             ) : (
-              <Skeleton animation="pulse" variant="rect" height="44px" />
+              <>
+                <Skeleton animation="pulse" variant="rect" height="44px" />
+                <div ref={observerRef} />
```
memoyil, line 89: After first intersection, this div will not be rendered thus we will not get isIntersecting state updates therefore we will not make request over and over again

src/hooks/useIntersectionObserver.ts
```diff
@@ -0,0 +1,26 @@
...
+      })
+      intersectionObserver.observe(observerRef.current)
+      setObserverIsSet(true)
+    }
+  }, [observerIsSet])
+
+  return { observerRef, isIntersecting }
+}
+
+export default useIntersectionObserver
```
hachiojidev, line 26: `ref` never gets set. I think you need to pass it in.

memoyil: It is not set in here, we pass ref object to the callers of that hook and they should add this to a DOM node, For example;

https://github.com/pancakeswap/pancake-frontend/blob/0a2c6d7a74a9842514451933ae7298389908f760/src/views/Home/components/EarnAPRCard.tsx#L89

hachiojidev (approved):

PR #1525, opened 2021-06-17
chore: Remove batchRequest

RabbitDoge:
Replace `makeBatchRequest` by multicallv2


src/utils/multicall.ts
```diff
@@ -46,12 +46,10 @@ export const multicallv2 = async (abi: any[], calls: Call[], options: MulticallO
     .call(undefined, options.blockNumber)
   const res = returnData.map((call, i) => {
     const [result, data] = call
-    return {
-      result,
-      data: itf.decodeFunctionResult(calls[i].name, data),
-    }
+    return result ? itf.decodeFunctionResult(calls[i].name, data) : null
```
hachiojidev, line 49: https://www.npmjs.com/package/ethereum-multicall
>It will still be in the same order as you expect but you have a success boolean to check if it passed or failed.

I think we should check against the `success` variable. Have you tested a failure?

hachiojidev (approved):

PR #1533, opened 2021-06-18
feat: (home) Replace lottery card with prediction promotion card

memoyil:
To review:

https://deploy-preview-1533--pancakeswap-dev.netlify.app/

Also brings back WinCard and naming change on that


src/state/predictions/helpers.ts
```diff
@@ -228,6 +236,25 @@ export const getMarketData = async (): Promise<{
...
 
+export const getStats = async (): Promise<{
+  stats: Stats
+}> => {
+  const response = (await request(
+    GRAPH_API_PREDICTION,
+    gql`
+      query getMarketData {
+        market(id: 1) {
+          totalBNB
```
ChefKai, line 247: In order to have a more accurate number to display, we should add `totalBNBTreasury` to the query, and subtract it from `totalBNB` to account for the treasury fee (used for the burn) + the `House` rounds

memoyil: I added totalBNBTreasury to the calculation, what are the House rounds? (failed ones?)

memoyil: @ChefKai Done, Can you review it again?

src/state/predictions/helpers.ts
```diff
@@ -228,6 +244,25 @@ export const getMarketData = async (): Promise<{
...
+    GRAPH_API_PREDICTION,
+    gql`
+      query getTotalWonData {
+        market(id: 1) {
+          totalBNB
+          totalBNBTreasury
+        }
+        rounds(where: { position: ${BetPosition.HOUSE} }) {
+          totalAmount
+        }
```
hachiojidev, line 258: Pass in the variables like this with the second argument. You can check `getRound` as an example.

hachiojidev (changes requested):

src/state/predictions/helpers.ts
memoyil: Done

hachiojidev (approved):

hachiojidev (changes requested):
@memoyil After looking at the preview can you do 2 things:

1. Make sure the skeleton is the same size as the final number size so it doesn't jump

2. Use the count up component.

memoyil:
@hachiojidev Done

hachiojidev (approved):
Looks good.

PR #1541, opened 2021-06-21
Feature/ethers migration

RabbitDoge:
Migrate from web3js to ethers


src/hooks/useAllowance.ts
```diff
@@ -16,8 +16,8 @@ export const useLotteryAllowance = () => {
 
   useEffect(() => {
     const fetchAllowance = async () => {
-      const res = await cakeContract.methods.allowance(account, getLotteryAddress()).call()
-      setAllowance(new BigNumber(res))
+      const res = await cakeContract.allowance(account, getLotteryAddress())
+      setAllowance(new BigNumber(res.toString()))
```
ChefHutch, line 20: Hey Doge - just checking, as this is relevant for the multicall implementation I'm finishing up in the lottery. Is it OK to go from an ethers.BigNumber (i.e. what I think will be returned here in `res`), and then construct a new BigNumber.js BN, by passing in the ethers.BigNumber & calling `.toString()` on it? Will we look to migrate this in future too?

RabbitDoge: yes, the string representation can be passed in the contructor

ChefHutch: Sure - I more mean in terms of working with multiple BigNumber libraries - is that going to be the strategy?

Ethers response uses BN.js
The response is passed into a BigNumber.js constructor
Thereafter in the FE for calculations etc, BigNumber.js is used?

RabbitDoge: this is the short term strategy, until we get rid of BigNumber.js

hachiojidev (approved):

PR #1548, opened 2021-06-22
fix: (menu) Add missing translation in menu

memoyil:
To reproduce the issue

1. Check sidebar menu
2. Change language
3. Check translation of 'New' tag


src/config/localization/translations.json
```diff
@@ -820,9 +820,9 @@
   "Max. stake per user": "Max. stake per user",
   "You have claimed your rewards!": "You have claimed your rewards!",
   "Already Collected": "Already Collected",
-  "Up or Down?": "Up or Down?",
```
memoyil, line 823: It is not using anymore

hachiojidev (approved):

PR #1550, opened 2021-06-22
chore(endpoints): Move subgraph URLs out of env variables

ChefHutch:
- The subgraph URLs don't change based on env, so move them into the /endpoints file

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -24,7 +24,6 @@ export const BASE_EXCHANGE_URL = 'https://exchange.pancakeswap.finance'
 export const BASE_ADD_LIQUIDITY_URL = `${BASE_EXCHANGE_URL}/#/add`
 export const BASE_LIQUIDITY_POOL_URL = `${BASE_EXCHANGE_URL}/#/pool`
 export const BASE_BSC_SCAN_URL = 'https://bscscan.com'
-export const BASE_SUBGRAPH_URL = 'https://api.thegraph.com/subgraphs/name/pancakeswap'
 export const LOTTERY_MAX_NUMBER_OF_TICKETS = 50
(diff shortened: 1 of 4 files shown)


hachiojidev (commented):
I am okay with this but we do lose the ability to change the URL during development by simply updating a `.env.development.local` or `.env.local`

ChefHutch:
Yea also a good point @hachiojidev - easy enough to change whatever is assigned in `export const GRAPH_API_PROFILE = `${BASE_SUBGRAPH_URL}/profile` though? I guess this change is making the assumption that these values wouldn't be env-specific.

PR is following some feedback from Doge here https://github.com/pancakeswap/pancake-frontend/pull/1356#discussion_r651518444

src/config/index.ts
```diff
@@ -24,6 +24,7 @@ export const BASE_EXCHANGE_URL = 'https://exchange.pancakeswap.finance'
 export const BASE_ADD_LIQUIDITY_URL = `${BASE_EXCHANGE_URL}/#/add`
 export const BASE_LIQUIDITY_POOL_URL = `${BASE_EXCHANGE_URL}/#/pool`
 export const BASE_BSC_SCAN_URL = 'https://bscscan.com'
+export const BASE_SUBGRAPH_URL = 'https://api.thegraph.com/subgraphs/name/pancakeswap'
```
hachiojidev, line 27: I think this should stay in `endpoints.ts`

ChefHutch: Sure thing man, have moved it 👍

ChefHutch:
Ah, this does get changed at the env level sometimes within Netlify - we've had a few occasions where we needed to switch to using the BSC graph. Closing

PR #1551, opened 2021-06-22
fix: (select) Dropdown is not closed when clicking outside of the element

memoyil:
To review:

https://deploy-preview-1551--pancakeswap-dev.netlify.app/

To reproduce the issue

1. Go to pools
2. Click sort by dropdown
3. Then click outside of the element to close it intuitively
4. Dropdown should be collapsed

### DIFF
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -108,7 +108,10 @@ const Select: React.FunctionComponent<SelectProps> = ({ options, onChange }) => 
   const [selectedOptionIndex, setSelectedOptionIndex] = useState(0)
   const [containerSize, setContainerSize] = useState({ width: 0, height: 0 })
 
-  const toggling = () => setIsOpen(!isOpen)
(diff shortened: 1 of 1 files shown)


src/components/Select/Select.tsx
```diff
@@ -108,7 +108,10 @@ const Select: React.FunctionComponent<SelectProps> = ({ options, onChange }) =>
   const [selectedOptionIndex, setSelectedOptionIndex] = useState(0)
   const [containerSize, setContainerSize] = useState({ width: 0, height: 0 })
 
-  const toggling = () => setIsOpen(!isOpen)
+  const toggling = (event: React.MouseEvent<HTMLDivElement>) => {
+    setIsOpen(!isOpen)
+    event.stopPropagation()
```
hachiojidev, line 113: Why `stopPropagation`?

memoyil: It propogates through to document that we also listen click event which causes close the dropdown.

hachiojidev (approved):

PR #1554, opened 2021-06-22
chore(contractHelpers): Use SDK for chainIds

ChefHutch:
- Remove chain Ids from config
- Replace places where they are used with SDK values


hachiojidev (approved):
Nice.

PR #1555, opened 2021-06-22
feat: (prediction) Don't render trading view chart and websockets when it is not open

memoyil:
To review:

https://deploy-preview-1555--pancakeswap-dev.netlify.app/

To reproduce the issue

1. Go to predictions
2. Check websockets when chart is not open
3. See trading view websockets are instantiated
4. It should not be instantiated until chart is open


hachiojidev:
This is by design. Originally we instantiated it when you opened it but the UI was really jumpy and laggy.

PR #1556, opened 2021-06-22
fix: (poolcard) Earning token dolar balance shows usd text only when price is loading

memoyil:
To review:

https://deploy-preview-1556--pancakeswap-dev.netlify.app/

To reproduce the issue

1. Go to pools
2. See one of the pools (except cakes) earning token price loading
3. Check usd  (see below)

Screenshot 2021-06-22 at 18 22 13

Screenshot 2021-06-22 at 21 04 34

Screenshot 2021-06-22 at 20 57 09

With this solution it will not show anything regarding to usd price (like in farm cards) until usd price is loaded


hachiojidev (changes requested):
Looks good, just need to remove the unused variable.

memoyil:
> Looks good, just need to remove the unused variable.

Done

hachiojidev (approved):

PR #1561, opened 2021-06-23
feat: Add lpAprs json file and github action to update it

Chef-Cheems:
This PR adds `src/config/constants/lpAPRs.json` that stores LP rewards APR estimation.
GitHub Action is added to update this file every day at 23:15 UTC (GitHub suggests not to set time at exact hour to prevent delays due to load. Also scheduled runs are usually delayed in any case, but should run somewhere between 23:15-00:00)

GitHub action will create a PR for updated values that we can review for potential errors (e.g. huge APRs, zeros, etc.)

PR also includes initial fetch for lpAprs.json


scripts/updateLPsAPR/index.ts
```diff
@@ -0,0 +1,136 @@
...
+  return `pairs(
+      first: 30,
+      where: { id_in: ${addressesString} }
+      ${blockString}
+    ) {
+      id
+      volumeUSD
+      reserveUSD
+    }`
+}
```
hachiojidev, line 31: This is kind of dirty. I think we should use `graphql-request` with `gql`.

scripts/updateLPsAPR/index.ts
```diff
@@ -0,0 +1,136 @@
...
+
+const FARMS_BULK = (blockWeekAgo: number, farms: string[]) => {
```
hachiojidev, line 33: Can we rename this function to something more descriptive?

scripts/updateLPsAPR/index.ts
```diff
@@ -0,0 +1,136 @@
...
+}
+
+interface AprMap {
+  [key: string]: BigNumber
+}
+
+const getWeekAgoTimestamp = () => {
+  const weekAgoMs = new Date().getTime() - 7 * 24 * 60 * 60 * 1000
+  const weekAgoUnix = new Date(weekAgoMs).getTime() / 1000
+  return Math.floor(weekAgoUnix).toString()
```
hachiojidev, line 66: This is totally fine but consider using `date-fns`, we have it available.

scripts/updateLPsAPR/index.ts
```diff
@@ -0,0 +1,136 @@
...
+}
+
+const LP_HOLDERS_FEE = 0.0017
+const WEEKS_IN_A_YEAR = 52.1429
+
+const get7dBlock = async () => {
```
hachiojidev, line 72: I think we should avoid methods like this. Something more generic would be easier to maintain.

scripts/updateLPsAPR/index.ts
```diff
@@ -0,0 +1,136 @@
...
+}
+
+interface SingleFarmResponse {
+  id: string
+  reserveUSD: string
+  volumeUSD: string
+}
+
+interface FarmsResponse {
+  now: SingleFarmResponse[]
```
hachiojidev, line 55: now?

scripts/updateLPsAPR/index.ts
```diff
@@ -0,0 +1,136 @@
...
+    }, {})
+    return aprs
+  } catch (error) {
+    throw new Error(`Failed to fetch LP APR data: ${error}`)
+  }
+}
+
+const fetchAndUpdateLPsAPR = async () => {
+  const lowerCaseAddresses = farmsConfig
+    .filter((farm) => farm.pid > 250)
```
hachiojidev, line 109: At the very least this needs a comment explaining what we are filtering here.

hachiojidev (changes requested):

scripts/updateLPsAPR/index.ts
Chef-Cheems: now as in "at latest available block right now"

scripts/updateLPsAPR/index.ts
Chef-Cheems: This is mostly to keep stuff DRY, but I guess in this case as we don't request much I can just verbosely put everything into query below without this helper function 👍

scripts/updateLPsAPR/index.ts
Chef-Cheems: renamed to `farmsAtLatestBlock`

scripts/updateLPsAPR/index.ts
Chef-Cheems: refactored to `getBlockAtTimestamp = async (timestamp: string) => {...`

scripts/updateLPsAPR/index.ts
Chef-Cheems: For some reason I forgot we had it as dependency to be honest 😅

scripts/updateLPsAPR/index.ts
Chef-Cheems: Removed it completely query is small enough to just put it directly as argument to graphql-request

hachiojidev (approved):

PR #1565, opened 2021-06-23
feat: (home) Do not send tvl and stats requests when it is not visible

memoyil:
After first intersection, the load data status won't be changed to not send flood of requests when user scrolls up and down.

3 requests dropped if user doesn't scroll down (both applicable mobile and desktop since they are at the bottom)

### DIFF
--- a/src/hooks/api.ts
+++ b/src/hooks/api.ts
@@ -11,7 +11,7 @@ export interface DeBankTvlResponse {
   tvl: number
 }
 
-export const useGetStats = () => {
+export const useGetStats = (loadData = true) => {
   const [data, setData] = useState<DeBankTvlResponse | null>(null)
 
   useEffect(() => {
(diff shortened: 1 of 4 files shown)


hachiojidev (commented):
Something doesn't sit right with this. I get that it works but we should come up with a more generic solution.

1. `useGetStats` is only used in one place, I think we can move it to the `TotalValueLockedCard`

2. I think maybe hooks like these should not use slow/fast refresh. It would be better for an instance to have its own timer so it can be paused.

I would like to have a uniform way of writing these kind of fetch hooks where:

1. An option to not run on initial load render
2. Return a helpers to stop/start the timer

I think further discussion is needed. @ChefHutch @Chef-Cheems @RabbitDoge , what do you guys think?

I wonder if we should consider helper libraries like [SWR](https://swr.vercel.app/)

memoyil:
Closed until the decision

PR #1574, opened 2021-06-25
fix: (Aprcalculatormodal) No space between bullet and text in bullet list

memoyil:
To review:

https://deploy-preview-1574--pancakeswap-dev.netlify.app/

To reproduce the issue (in Firefox, Safari mobile)

1. Go to pools or farms
2. Click apr calculator button on any farm or pool

Check the gap issue has not been reintroduced in Chrome

Before:

Screenshot 2021-06-25 at 14 15 43

After:

Screenshot 2021-06-25 at 14 15 20

### DIFF
--- a/src/components/ApyCalculatorModal/index.tsx
+++ b/src/components/ApyCalculatorModal/index.tsx
@@ -31,6 +31,20 @@ const GridHeaderItem = styled.div`
 `
 
 const BulletList = styled.ul`
+  list-style-type: none;
+  margin: 0;
+  padding: 0;
+
+  li {
+    margin: 0;
+    padding: 0;
+  }
+
+  li::before {
+    content: '•';
+    margin-right: 4px;
+  }
+
   li::marker {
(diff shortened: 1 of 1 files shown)


Chef-Cheems (approved):
Huh... weird.... Seems to be Firefox specific issue 🤔 
I initially added the margin cause on chrome gap looked a bit too big. Would be cool to make it look the same on all browsers but I guess slightly bigger gap on Chrome is not critical 😅

memoyil:
> Huh... weird.... Seems to be Firefox specific issue 🤔
> I initially added the margin cause on chrome gap looked a bit too big. Would be cool to make it look the same on all browsers but I guess slightly bigger gap on Chrome is not critical 😅

Same behaviour is also visible in safari mobile, I changed the logic a bit, now it fixes both chrome gap and firefox safari no space.

hachiojidev (approved):

PR #1586, opened 2021-06-28
chore: Restructure component folder

RabbitDoge:
Remove unused components.
Keep only generic components inside the component folder.


hachiojidev (approved):

ChefHutch (commented):
It ain't perfect - but the `components/Timer` component that's being moved into the Trading Comp folder actually is intended as a generic. I'm also using it in the Lottery - https://github.com/pancakeswap/pancake-frontend/pull/1356/files#diff-d23f439ccb41d71563c27fa94210a3ab2f618ea5d160155df3ca5fceb6a03fbdR4

RabbitDoge:
> It ain't perfect - but the `components/Timer` component that's being moved into the Trading Comp folder actually is intended as a generic. I'm also using it in the Lottery - https://github.com/pancakeswap/pancake-frontend/pull/1356/files#diff-d23f439ccb41d71563c27fa94210a3ab2f618ea5d160155df3ca5fceb6a03fbdR4

A generic timer system should be a hook, and not a component. The logic is the same, but the layout and design is unlikely to be the same

ChefHutch:
> A generic timer system should be a hook, and not a component. The logic is the same, but the layout and design is unlikely to be the same

OK cool - happy with that. They _are_ visually very similar, but not identical and it probably makes sense to maintain separate visual components. 👍 

Screenshot 2021-06-28 at 09 43 09
Screenshot 2021-06-28 at 09 43 38

ChefHutch (approved):

PR #1590, opened 2021-06-28
refactor: (multiplier) Remove redundant div in multiplier tooltip

memoyil:
To review:

https://deploy-preview-1590--pancakeswap-dev.netlify.app/

There should not be no difference in view when reviewing

To review:

1. Farms
2. Hover in desktop on question mark next to the multiplier
3. See no difference in tooltip


hachiojidev (approved):
Thanks for checking. It could be possible it is a child of a flex container.

PR #1614, opened 2021-06-30
chore: Remove unused hooks

RabbitDoge:
[ ] Explain what your PR does


src/contexts/RefreshContext.tsx
```diff
@@ -1,34 +1,52 @@
...
+    window.addEventListener('visibilitychange', onVisibilityChange)
+
+    return () => {
+      window.removeEventListener('visibilitychange', onVisibilityChange)
+    }
+  }, [])
+
+  return isBrowserTabActiveRef
+}
+
```
hachiojidev, line 26: I think this should be its own hook. You can use this other places where you want to detect if the browser tab is hidden or not.

src/utils/calls/estimateGas.ts
```diff
@@ -0,0 +1,39 @@
...
+export const estimateGas = async (contract: Contract, methodName: string, methodArgs: any[]) => {
+  if (!contract[methodName]) {
+    throw new Error(`Method ${methodName} doesn't exist on ${contract.address}`)
+  }
+  const rawGasEstimation = await contract.estimateGas[methodName](...methodArgs)
+  // By convention, ethers.BigNumber values are multiplied by 1000 to avoid dealing with real numbers
+  const gasEstimation = rawGasEstimation
+    .mul(ethers.BigNumber.from(10000).add(ethers.BigNumber.from(1000)))
+    .div(ethers.BigNumber.from(10000))
+  return gasEstimation
```
hachiojidev, line 19: Didn't you do this in another PR?

src/views/Farms/hooks/useApprove.ts
```diff
@@ -0,0 +1,20 @@
+import { useCallback } from 'react'
+import { ethers, Contract } from 'ethers'
+import { useMasterchef } from 'hooks/useContract'
+
+const useApprove = (lpContract: Contract) => {
```
hachiojidev, line 5: We should consider renaming this to something like `useApproveLpContract`

src/views/Farms/hooks/useHarvest.ts
```diff
@@ -0,0 +1,15 @@
+import { useCallback } from 'react'
+import { harvest } from 'utils/calls'
+import { useMasterchef } from 'hooks/useContract'
+
+const useHarvest = (farmPid: number) => {
```
hachiojidev, line 5: Same here, `useHarvestFarm` or something.

src/views/Farms/hooks/useStake.ts
```diff
@@ -0,0 +1,19 @@
+import { useCallback } from 'react'
+import { stake } from 'utils/calls'
+import { useMasterchef } from 'hooks/useContract'
+
+const useStake = (pid: number) => {
```
hachiojidev, line 5: And here...maybe `useStakeInPool`

src/views/Farms/hooks/useUnstake.ts
```diff
@@ -0,0 +1,18 @@
+import { useCallback } from 'react'
+import { unstake } from 'utils/calls'
+import { useMasterchef } from 'hooks/useContract'
+
+const useUnstake = (pid: number) => {
```
hachiojidev, line 5: and...here

src/contexts/RefreshContext.tsx
RabbitDoge: Yes but we don't. And I don't see a single use case where we want to use it. If one day we need it elsewhere, we will move it.

src/utils/calls/estimateGas.ts
RabbitDoge: just moving to s separate file

RabbitDoge: image

hachiojidev (approved):

PR #1615, opened 2021-06-30
fix(profile): IFO achievement status check

hachiojidev:
`checkClaimStatuses` was hitting the max limit causing it to throw an error. Migrated to multicall.

### DIFF
--- a/src/utils/achievements.ts
+++ b/src/utils/achievements.ts
@@ -1,6 +1,5 @@
 import { ethers } from 'ethers'
 import { Campaign } from 'config/constants/types'
-import { getPointCenterIfoContract } from 'utils/contractHelpers'
 import ifosList from 'config/constants/ifo'
 import { campaignMap } from 'config/constants/campaigns'
 import { Achievement, TranslatableText } from 'state/types'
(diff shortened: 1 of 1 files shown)


src/utils/achievements.ts
```diff
@@ -47,24 +46,34 @@ export const getAchievementDescription = (campaign: Campaign): TranslatableText
...
-  const pointCenterContract = getPointCenterIfoContract()
 
   // Returns the claim status of every IFO with a campaign ID
-  let claimStatuses = []
-  try {
-    claimStatuses = (await pointCenterContract.checkClaimStatuses(account, ifoCampaignAddresses)) as boolean[]
-  } catch (error) {
-    console.error(error)
-  }
+  const claimStatusCalls = ifoCampaigns.map(({ address }) => {
```
RabbitDoge, line 51: you can use a shortcut without the return

RabbitDoge (approved):

src/utils/achievements.ts
hachiojidev: Yeah...I have been avoiding doing that. I like the explicit return recently.

PR #1623, opened 2021-07-01
feat: Support for BSC Wallet signing

hachiojidev:
Adds a sign message method that uses `bnbSign` if wallet is BSC.

### DIFF
--- a/src/react-app-env.d.ts
+++ b/src/react-app-env.d.ts
@@ -1,8 +1,11 @@
 /// <reference types="react-scripts" />
 
-interface WindowChain {
+interface Window {
   ethereum?: {
     isMetaMask?: true
     request?: (...args: any[]) => void
+  }
+  BinanceChain?: {
+    bnbSign?: (address: string, message: string) => Promise<{ publicKey: string; signature: string }>
   }
 }
--- a/src/react-app-env.d.ts
+++ b/src/react-app-env.d.ts
@@ -3,7 +3,7 @@
 interface Window {
   ethereum?: {
     isMetaMask?: true
-    request?: (...args: any[]) => void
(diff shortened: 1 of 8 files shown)


src/react-app-env.d.ts
```diff
@@ -1,8 +1,11 @@
...
 
-interface WindowChain {
+interface Window {
   ethereum?: {
     isMetaMask?: true
-    request?: (...args: any[]) => void
+    request?: (...args: any[]) => Promise<void>
+  }
+  BinanceChain?: {
+    bnbSign?: (address: string, message: string) => Promise<{ publicKey: string; signature: string }>
```
hachiojidev, line 9: We can simply extend the current window object

src/utils/web3React.ts
```diff
@@ -31,3 +31,22 @@ export const getLibrary = (provider): ethers.providers.Web3Provider => {
...
+  provider: ethers.providers.JsonRpcProvider | ethers.providers.Web3Provider
+  message: string
+  account: string
+}
+
+/**
+ * BSC Wallet requires a different sign method
+ * @see https://docs.binance.org/smart-chain/wallet/wallet_api.html#binancechainbnbsignaddress-string-message-string-promisepublickey-string-signature-string
+ */
+export const signMessage = async ({ provider, account, message }: SignMessage): Promise<string> => {
```
RabbitDoge, line 45: any reason why you pass an object instead of a list of params ?

hachiojidev: Yeah, I see what you mean. I originally did this because my first iteration had account optional. Since they are all required I think 3 separate arguments are better. Let me update.

RabbitDoge (approved):

PR #1632, opened 2021-07-02
Separate CAKE and LP APR

Chef-Cheems:
Show base APR and combined APR separately
Screenshot 2021-07-02 at 10 47 51

### DIFF
--- a/src/components/ApyCalculatorModal/index.tsx
+++ b/src/components/ApyCalculatorModal/index.tsx
@@ -8,7 +8,7 @@ interface ApyCalculatorModalProps {
   onDismiss?: () => void
   tokenPrice: number
   apr: number
-  combinedApr?: string
+  displayApr?: string
   linkLabel: string
   linkHref: string
   earningTokenSymbol?: string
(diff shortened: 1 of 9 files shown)


src/components/ApyCalculatorModal/index.tsx
```diff
@@ -8,6 +8,7 @@ interface ApyCalculatorModalProps {
   onDismiss?: () => void
   tokenPrice: number
   apr: number
+  combinedApr?: string
```
hachiojidev, line 11: A small thing but maybe we should change the name of this. In other places your variable names ending `Apr` are numbers.

Alternatively, make this a number.

hachiojidev: Maybe add the word `display` like we do in other places.

Chef-Cheems: Changed to `displayApr`, what do you think?

ChefHutch (approved):
👍

hachiojidev (approved):

PR #1634, opened 2021-07-02
feat: Helper function to format display of ethers' formatUnits

hachiojidev:


src/utils/formatBalance.ts
```diff
@@ -30,3 +32,25 @@ export const formatNumber = (number: number, minPrecision = 2, maxPrecision = 2)
...
+  if (leftSide.length > 0) {
+    charCount += leftSide.length
+  }
+
+  // Add 1 for the decimal point
+  if (rightSide.length > 0) {
+    charCount += 1
+  }
+
+  return balance.substr(0, charCount <= balance.length ? charCount : balance.length)
```
hachiojidev, line 55: We have two routes, given a 4 decimals display of `2.1`

1. Like `toFixed` where it pads the decimals e.g. `2.1` --> `2.1000`

2. Like `toLocaleString` where it uses the max amount of decimals e.g. `2.1` --> `2.1`

I opted for #2.

src/__tests__/utils/formatBalance.test.ts
```diff
@@ -0,0 +1,22 @@
+import { BigNumber } from 'ethers'
+import { formatBigNumber } from 'utils/formatBalance'
+
+describe('getBscScanAddressUrl', () => {
```
Chef-Cheems, line 4: `formatBalance`

src/utils/formatBalance.ts
```diff
@@ -30,3 +32,25 @@ export const formatNumber = (number: number, minPrecision = 2, maxPrecision = 2)
...
+  let charCount = displayDecimals
+
+  if (leftSide.length > 0) {
+    charCount += leftSide.length
+  }
+
+  // Add 1 for the decimal point
+  if (rightSide.length > 0) {
+    charCount += 1
+  }
```
Chef-Cheems, line 53: I'm a bit confused by all this conditions:
- `leftSide.length` will always be more than 0, no? Seems that no need to check for that
- `rightSide.length` also will be always > 0 unless token `decimals` is 0. In that case whole `rightSide` will be `undefined` and `undefined.length` will throw error.

src/utils/formatBalance.ts
Chef-Cheems: Should we also plug in `.toLocaleString()` right here to add `,` between number units if its meant to be used for UI only?  e.g.`1,250,000.00` instead of `1250000.00`
It will need to be converted to number beforehand though, `.toLocaleString()` on string representation does nothing

src/utils/formatBalance.ts
```diff
@@ -1,4 +1,6 @@
 import BigNumber from 'bignumber.js'
+import { ethers, FixedNumber } from 'ethers'
```
Chef-Cheems, line 2: FixedNumber is not used, but otherwise good to go 👍

RabbitDoge (approved):

Chef-Cheems (approved):

PR #1643, opened 2021-07-05
feat: Add structure to type contract calls

hachiojidev:
#### Add typing for multicall. You can now specify the type like so

```
const booleans = await multicall<boolean[]>(abi, calls)
```
`multicallv2` is a little different as it can return null (although doesn't change the typescript message).

#### Add example on how to type contract functions.

We can now type contract functions to make it easier to work with.

### DIFF
--- a/src/utils/contractHelpers.ts
+++ b/src/utils/contractHelpers.ts
@@ -55,6 +55,7 @@ import chainlinkOracleAbi from 'config/abi/chainlinkOracle.json'
 import MultiCallAbi from 'config/abi/Multicall.json'
 import bunnySpecialCakeVaultAbi from 'config/abi/bunnySpecialCakeVault.json'
 import bunnySpecialPredictionAbi from 'config/abi/bunnySpecialPrediction.json'
+import { PredictionsContract } from './types'
 
 const getContract = (abi: any, address: string, signer?: ethers.Signer | ethers.providers.Provider) => {
   const signerOrProvider = signer ?? simpleRpcProvider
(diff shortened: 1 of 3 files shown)


src/utils/contractHelpers.ts
```diff
@@ -127,9 +128,11 @@ export const getEasterNftContract = (signer?: ethers.Signer | ethers.providers.P
 export const getCakeVaultContract = (signer?: ethers.Signer | ethers.providers.Provider) => {
   return getContract(cakeVaultAbi, getCakeVaultAddress(), signer)
 }
+
 export const getPredictionsContract = (signer?: ethers.Signer | ethers.providers.Provider) => {
-  return getContract(predictionsAbi, getPredictionsAddress(), signer)
+  return getContract(predictionsAbi, getPredictionsAddress(), signer) as PredictionsContract
```
hachiojidev, line 133: Tried to use a generic on `useContract` but it threw an error that I could not figure out how to resolve.

ChefHutch (approved):
👍

src/utils/types.ts
```diff
@@ -0,0 +1,22 @@
...
+  endBlock: ethers.BigNumber
+  lockPrice: ethers.BigNumber
+  closePrice: ethers.BigNumber
+  totalAmount: ethers.BigNumber
+  bullAmount: ethers.BigNumber
+  bearAmount: ethers.BigNumber
+  rewardBaseCalAmount: ethers.BigNumber
+  rewardAmount: ethers.BigNumber
+  oracleCalled: boolean
+}
```
hachiojidev, line 18: I think exporting these function returns separately is a good idea so we can use the same type with `multicall`

```
const rounds = await multicall<PredictionsRounds[]>(abi, calls)
```

Chef-Cheems (approved):

RabbitDoge (approved):

PR #1649, opened 2021-07-06
DNM - just checking diff chore(lottery-v1): Remove all lottery v1 code from codebase

ChefHutch:
- Remove all v1 code
- Rename `LotteryV2` view components to `Lottery` (note - contracts still distinguished as v2)


src/views/Lottery/components/Hero.tsx
```diff
@@ -1,104 +1,279 @@
 import React from 'react'
-import styled from 'styled-components'
-import { Heading, Text, LinkExternal } from '@pancakeswap/uikit'
+import styled, { keyframes } from 'styled-components'
```
ChefHutch, line 2: This is a straight delete of v1 and replace with v2, but the file name is the same, so Git is parsing it as a diff

`views/Lottery/components/Hero` > deleted
`views/LotteryV2/components/Hero` >  renamed `views/Lottery/components/Hero`

Chef-Cheems (changes requested):
Messaged in slack about additional removals

Chef-Cheems (approved):

package.json
```diff
@@ -55,7 +55,6 @@
     "lodash": "^4.17.21",
     "node-fetch": "^2.6.1",
     "react": "^17.0.1",
-    "react-chartjs-2": "2.11.1",
```
RabbitDoge, line 58: and `chart.js` ?

ChefHutch: Good point, removed

RabbitDoge (approved):

PR #1660, opened 2021-07-06
fix: (lotterybanner) Adapt lottery banner and promotion to different languages

memoyil:
To review:

https://deploy-preview-1660--pancakeswap-dev.netlify.app/

@ChefHutch I was about to write this to the comment of previous pr but then it is merged, So i decided to implement the logic by myself.

### DIFF
--- a/src/views/Home/components/LotteryBanner.tsx
+++ b/src/views/Home/components/LotteryBanner.tsx
@@ -38,7 +38,10 @@ const LeftWrapper = styled(Flex)`
   flex-direction: column;
   flex: 1;
   padding-bottom: 40px;
-  padding-top: 40px;
+  padding-top: 24px;
+  ${({ theme }) => theme.mediaQueries.sm} {
+    padding-top: 40px;
+  }
 `
 
 const RightWrapper = styled.div`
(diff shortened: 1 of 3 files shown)


src/views/Home/components/LotteryBanner.tsx
```diff
@@ -38,7 +38,9 @@ const LeftWrapper = styled(Flex)`
   flex-direction: column;
   flex: 1;
   padding-bottom: 40px;
-  padding-top: 40px;
```
ChefHutch, line 41: Reinstate this please memoyil - it can be changed to 24px though

Not related to your change - but could you give `RightWrapper` a margin top of 24px only below `theme.mediaQueries.sm` too please?

ChefHutch (approved):
We were actually talking about this in slack, memoyil mind reader.

Small styling change requested but otherwise lgtm

src/views/Home/components/LotteryBanner.tsx
memoyil: Done

hachiojidev (approved):

PR #1664, opened 2021-07-07
Feature/predictions refactor

hachiojidev:
### Overview

To increase reliability this PR migrates data calls from a subgraph to the nodes directly. The bet history panel remains on the subgraph.

### Challenges

- Most of the code was reliant on the subgraph's shape (round id, etc...). Needed to migrate that to use the `epoch`
- Data format and shape from the nodes is quite different than the subgraph. A lot of transitional code is required to keep compatibility.

### Notes

- `Bet` has been renamed to `Ledger` for node data
- There is still some leftover duplicate code that will be resolved in the future.


src/views/Predictions/components/History/helpers.ts
```diff
@@ -0,0 +1,40 @@
+import { Bet, BetPosition } from 'state/types'
```
hachiojidev, line 1: Had to move these helpers into this folder because they are only used by the history pane now and deal with different kind of numbers.

monstro805 (approved):

RabbitDoge (approved):

PR #1669, opened 2021-07-07
fix: Auto scrolling ticket input to view, iOS numeric keypad

Chef-Cheems:
- The input you're editing should not be covered by the keyboard or other elements anymore (it scrolls into the view when you click on it)
- iOS should show numeric keypad instead of full keyboard

### DIFF
--- a/src/views/Lottery/components/BuyTicketsModal/TicketInput.tsx
+++ b/src/views/Lottery/components/BuyTicketsModal/TicketInput.tsx
@@ -216,7 +216,7 @@ const TicketContaier: React.FC<{
           value={ticket.numbers[0]}
           onKeyDown={(e: React.KeyboardEvent) => onChangeHandler(e, 0)}
           placeholder="_"
-          onChange={() => null}
+          onChange={(e) => e.preventDefault()}
           disabled={disabled}
           onFocus={onFocusHandler}
           onBlur={onBlurHandler}
@@ -229,7 +229,7 @@ const TicketContaier: React.FC<{
           value={ticket.numbers[1]}
           onKeyDown={(e: React.KeyboardEvent) => onChangeHandler(e, 1)}
           placeholder="_"
-          onChange={() => null}
+          onChange={(e) => e.preventDefault()}
           disabled={disabled}
           onFocus={onFocusHandler}
           onBlur={onBlurHandler}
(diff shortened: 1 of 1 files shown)


src/views/LotteryV2/components/BuyTicketsModal/TicketInput.tsx
```diff
@@ -207,6 +220,7 @@ const TicketContaier: React.FC<{
           onFocus={onFocusHandler}
           onBlur={onBlurHandler}
           onPaste={onPasteHandler}
+          inputMode="numeric"
```
memoyil, line 223: Pattern validation is needed too?

Chef-Cheems: Not really, pattern validation is done above in `onChangeHandler` (it has quite convoluted way of checking whether input is allowed or not, other things are ignored)

src/views/LotteryV2/components/BuyTicketsModal/TicketInput.tsx
```diff
@@ -106,8 +114,9 @@ const TicketContaier: React.FC<{
   const onChangeHandler = (event: React.KeyboardEvent, digitId: number) => {
     const currentKey = parseInt(event.key, 10)
 
-    if (event.key === 'e' || event.key === 'E' || event.key === '.') {
+    if (['e', 'E', '.', ',', '-', 'Unidentified'].includes(event.key)) {
```
Chef-Cheems, line 117: "Unidentified" is the `even.key` value emitted by mobile keyboard when you press on `.`, `-` or `,`
Android doesn't have the way to hide those buttons and for some reason `.` and `-` slips through if you click multiple times

Chef-Cheems: ugh, actually just tested it on preview in TW - this doesn't solve the problem...

src/views/LotteryV2/components/BuyTicketsModal/TicketInput.tsx
Chef-Cheems: I can add it just for double protection though 😄

ChefHutch (commented):
Nice one man, I think the behaviour is better on desktop too. Works well on my device but I will await confirmation from testing that it fixes the reported bug.

hachiojidev (approved):

PR #1672, opened 2021-07-07
feat(lottery-v2): Use node calls to populate recent User and Public Lottery data

ChefHutch:
### Testing approved 👍 good to merge on Mon/Tues

Preview https://new-node-calls.netlify.app/lottery

**What does it do**
- Fundamentally change two of the base data-fetching queries: `getUserLotteryData` & `getLotteriesData`
- Previously, for all non-current round data, we were reliant on the subgraph. This was problematic particularly when rounds ended, as the subgraph would drift out of sync and return stale data.
- These two queries **now make node calls as well as subgraph calls, for all essential user & public lottery data for the past 4 rounds.**
- The two data sources are blended - in all cases, essential data is being handled by the nodes, and some supplementary UI data comes from the subgraph
- For rounds older than 4 rounds ago - the subgraph is used as normal.
- This PR includes a fallback for where there is no subgraph data (i.e. in the case that thegraph goes down). In that instance, the 4 rounds worth of node responses will be used, and any graph data will be null.

**Notes**
- There are some follow-up TODOs marked in the PR, but we're keen to get this out to solve the transition lag.

### DIFF
--- a/src/state/lottery/getLotteriesData.ts
+++ b/src/state/lottery/getLotteriesData.ts
@@ -51,7 +51,6 @@ const applyNodeDataToLotteriesGraphResponse = (
 }
 
 const getGraphLotteries = async (): Promise<LotteryRoundGraphEntity[]> => {
-  let lotteries
   try {
     const response = await request(
       GRAPH_API_LOTTERY,
@@ -71,12 +70,11 @@ const getGraphLotteries = async (): Promise<LotteryRoundGraphEntity[]> => {
         }
       `,
     )
-    ;({ lotteries } = response)
+    return response.lotteries
   } catch (error) {
-    lotteries = []
     console.error(error)
+    return []
   }
-  return lotteries
 }
 
 const getLotteriesData = async (currentLotteryId: string): Promise<LotteryRoundGraphEntity[]> => {
--- a/src/state/lottery/getLotteriesData.ts
+++ b/src/state/lottery/getLotteriesData.ts
@@ -1,0 +1,66 @@
+import { request, gql } from 'graphql-request'
+import { GRAPH_API_LOTTERY } from 'config/constants/endpoints'
+import { LotteryRoundGraphEntity, LotteryResponse } from 'state/types'
+import { getRoundIdsArray, fetchMultipleLotteries } from './helpers'
+
+const applyNodeDataToLotteriesGraphResponse = (
+  nodeData: LotteryResponse[],
+  graphRespose: LotteryRoundGraphEntity[],
+): LotteryRoundGraphEntity[] => {
+  const mergedResponse = graphRespose.map((graphRound, index) => {
+    const nodeRound = nodeData[index]
+    if (nodeRound) {
+      // if isLoading === true, there has been an error - return graphRound
+      if (!nodeRound.isLoading) {
+        return {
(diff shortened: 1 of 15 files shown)


src/state/lottery/getLotteriesData.ts
```diff
@@ -0,0 +1,91 @@
...
+        }
+      }
+      return graphRound
+    }
+    return graphRound
+  })
+  return mergedResponse
+}
+
+const getLotteriesData = async (currentLotteryId: string): Promise<LotteryRoundGraphEntity[]> => {
```
ChefHutch, line 54: New public data fetch - called by `fetchPublicLotteries`. Requested on a fastRefresh & when the lottery status transitions

src/state/lottery/getUserLotteryData.ts
```diff
@@ -0,0 +1,147 @@
...
+        }
+      }
+      return graphRound
+    }
+    return graphRound
+  })
+  return mergedResponse
+}
+
+const getUserLotteryData = async (account: string, currentLotteryId: string): Promise<LotteryUserGraphEntity> => {
```
ChefHutch, line 50: New user data fetch - called by `fetchUserTicketsAndLotteries`, fired after ticket purchases and on load/account change

& `fetchUserLotteries`, fires as rounds transition and after successful user reward claims

src/state/lottery/getLotteriesData.ts
```diff
@@ -0,0 +1,90 @@
...
+            status
+            finalNumber
+            startTime
+            endTime
+            ticketPrice
+          }
+        }
+      `,
+    )
+    ;({ lotteries } = response)
```
Chef-Cheems, line 74: that's a bit ugly
could `return response.lotteries` perhaps. Albeit it would mean +1 more line in catch to return `[]` but idk

src/state/lottery/getUserLotteryData.ts
```diff
@@ -0,0 +1,145 @@
...
+      `,
+      { account: account.toLowerCase() },
+    )
+    const userRes = response.user
+
+    // If no user returned - return blank user
+    if (!userRes) {
+      user = blankUser
+    } else {
+      const formattedUser = userRes && {
```
Chef-Cheems, line 89: unnecessary check for `userRes`
can also directly assign `user = ` to it

ChefHutch: Bien!

Chef-Cheems (approved):

src/state/lottery/getLotteriesData.ts
```diff
@@ -0,0 +1,88 @@
...
+          totalUsers: graphRound.totalTickets,
+          winningTickets: graphRound.totalTickets,
+        }
+      }
+      return graphRound
+    }
+    return graphRound
+  })
+  return mergedResponse
+}
```
hachiojidev, line 51: If we are using node data do we need the subgraph data?

hachiojidev (approved):

src/state/lottery/getLotteriesData.ts
Chef-Cheems: Subgraph data is fallback in case node call fails.

PR #1677, opened 2021-07-08
Feature/voting verification

hachiojidev:
- Verifies the votes after they have loaded from snapshot so the page loads much quicker
- Sort proposals by `end` so the newest ones are first

### DIFF
--- a/src/state/voting/helpers.ts
+++ b/src/state/voting/helpers.ts
@@ -83,8 +83,43 @@ export const getVotes = async (first: number, skip: number, where: VoteWhere): P
   return response.votes
 }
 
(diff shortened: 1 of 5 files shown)


src/state/voting/helpers.ts
```diff
@@ -93,33 +134,7 @@ export const getAllVotes = async (proposalId: string, block?: number, votesPerCh
...
-          if (!response.ok) {
-            throw new Error(response.statusText)
-          }
-
-          const data = await response.json()
-          const verifiedVotes = votes.filter((vote) => {
-            return data.data[vote.voter.toLowerCase()]?.isValid === true
-          })
-
-          resolve(verifiedVotes)
```
hachiojidev, line 122: This code was moved out of this function and happens after the votes load.

RabbitDoge (approved):

PR #1679, opened 2021-07-08
(feature) Farm Auctions static page

ChefHutch:
- Needs application form url


src/components/PageSection/svg/CurvedSvg.tsx
```diff
@@ -56,17 +61,18 @@ export const CurvedSvgBottom = styled(CurvedSvg)<StyledSvgProps>`
...
+const ConcaveContainer = styled(Box)<{
+  clipPath: string
+  concaveBackgroundLight?: string
+  concaveBackgroundDark?: string
+}>`
   width: 100%;
   height: 20px;
-  background-color: ${({ clipPath, theme }) => {
-    if (clipPath === '#topConcaveCurve') {
-      return '#7645d9'
```
ChefHutch, line 64: Removing hard coded colors - using props instead @Chef-Cheems 

Case could probably be made for the `concaveColors` being an object (maybe also being amalgamated with `concave`) rather than separate into two strings.

src/views/FarmAuction/components/AuctionDetail.tsx
```diff
@@ -0,0 +1,67 @@
...
+            <Flex justifyContent="space-between">
+              <Text color="textSubtle">{t('End')}</Text>
+              <Text> {t('To be announced')}</Text>
+            </Flex>
+          </ScheduleInner>
+        </CardBody>
+      </LeftCard>
+      <RightCard>
+        <CardHeader variant="bubblegum">
+          <Heading color="#280D5F">{t('Auction Leaderboard')}</Heading>
```
RabbitDoge, line 53: this is the text color from the theme

RabbitDoge (approved):

src/views/FarmAuction/components/AuctionDetail.tsx
ChefHutch: `CardHeader` needs changes to the way variants are handled to make them responsive to themes - so I needed to hard code the heading color to be fixed regardless of theme until I fix that.

PR #1692, opened 2021-07-10
add Sdoge logo

DOUSENCAI:
Sdoge
It is a pure exercise and fluidity agreement, and it is also a non-exercise, which can be continuous exercise and exercise continuity


ChefNina:
This is not proper process to add logo.
Please refer: https://docs.pancakeswap.finance/contact-us/business-partnerships

PR #1693, opened 2021-07-10
fix: Add fixed width option to balance component

memoyil:
To review

https://deploy-preview-1693--pancakeswap-dev.netlify.app/

Open webpage in iphone 12 pro, can also be reproduced with iphone 5/SE device simulator in browsers

Check lottery banner blinking while price counting up

@ChefHutch I have applied two different approach, please let me know which one suits you most.

First one converts all 1’s to 9 to get the maximum width but has a cornercase that if number has too many 1’s it ends up with long space at the end of it.

Downside of second one is different font  ( fallback one) ,which supports same width for digits by tabular-nums variant, than the default one.

### DIFF
--- a/src/components/Balance.tsx
+++ b/src/components/Balance.tsx
@@ -9,6 +9,7 @@ interface BalanceProps extends TextProps {
   isDisabled?: boolean
   prefix?: string
   onClick?: (event: React.MouseEvent<HTMLElement>) => void
+  fixedWidth?: boolean
 }
 
 const Balance: React.FC<BalanceProps> = ({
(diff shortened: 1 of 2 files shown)


src/components/Balance.tsx
```diff
@@ -19,19 +20,42 @@ const Balance: React.FC<BalanceProps> = ({
...
   onClick,
+  fixedWidth,
   ...props
 }) => {
   const previousValue = useRef(0)
+  const containerStyles = () => {
+    if (fixedWidth) {
+      return {
+        fontVariantNumeric: 'tabular-nums',
+        fontFamily: 'sans-serif',
```
hachiojidev, line 31: Need to run this by Chef Bugs.

src/components/Balance.tsx
```diff
@@ -19,19 +20,42 @@ const Balance: React.FC<BalanceProps> = ({
   unit,
   prefix,
   onClick,
+  fixedWidth,
```
hachiojidev, line 23: I want to avoid specific props like this where possible. It would be better if we had a `containerProps` prop and created a new component that composed this one.

src/components/Balance.tsx
ChefBugs: It's a no from me, as the mix of fonts is visually jarring and there are other potential workarounds for this specific case.
Long term, I'm looking at other fonts that cover all our languages and include monospace numerals. Will return to that discussion at a later date.

memoyil: https://github.com/pancakeswap/pancake-frontend/pull/1718

You can review this

memoyil:
Closed and applied temporary solution in https://github.com/pancakeswap/pancake-frontend/pull/1718

PR #1707, opened 2021-07-13
fix(lottery-v2): Minor fixes to ticket-retrieving node calls

ChefHutch:


src/state/lottery/fetchUnclaimedUserRewards.ts
```diff
@@ -31,16 +31,22 @@ const fetchCakeRewardsForTickets = async (
       params: [roundId, id, rewardBracket],
     }
   })
-  const cakeRewards = await multicallv2(lotteryV2Abi, calls)
```
ChefHutch, line 34: Adding call & processing of response data to a try/catch

src/state/lottery/getUserLotteryData.ts
```diff
@@ -19,7 +19,7 @@ const applyNodeDataToUserGraphResponse = (
         status: nodeRound.status,
         lotteryId: nodeRound.lotteryId.toString(),
         claimed: hasRoundBeenClaimed(ticketDataForRound.userTickets),
-        totalTickets: `${ticketDataForRound.userTickets.length}`,
+        totalTickets: `${ticketDataForRound.userTickets.length.toString()}`,
```
ChefHutch, line 22: This is expecting a string value, so calling `.toString()` on .length (it's rare that this data will ever be returned, hence why this hasn't come up)

src/state/lottery/helpers.ts
```diff
@@ -11,7 +11,7 @@ import { ethersToSerializedBigNumber } from 'utils/bigNumber'
 
 const lotteryContract = getLotteryV2Contract()
 // Variable used to determine how many past rounds should be populated by node data rather than subgraph
-export const NUM_ROUNDS_TO_FETCH_FROM_NODES = 4
+export const NUM_ROUNDS_TO_FETCH_FROM_NODES = 2
```
ChefHutch, line 14: Only use node calls for the past 2 rounds of data - beyond that use subgraph

src/state/lottery/index.ts
```diff
@@ -66,7 +66,7 @@ export const fetchUserTicketsAndLotteries = createAsyncThunk<
   const userTickets = await fetchTickets(account, currentLotteryId, totalTickets)
 
   // user has not bought tickets for the current lottery
-  if (userTickets.length === 0) {
+  if (!userTickets || userTickets.length === 0) {
```
ChefHutch, line 69: Add condition for when `fetchTickets` fails and returns null - so that it still returns `userLotteriesRes` and a users' history shows

src/state/lottery/helpers.ts
```diff
@@ -143,7 +143,7 @@ export const processRawTicketsResponse = (ticketsResponse: UserTicketsResponse):
 
 export const getViewUserTicketInfoCalls = (totalTicketsToRequest: number, account: string, lotteryId: string) => {
   let cursor = 0
-  const perRequestLimit = 1000
+  const perRequestLimit = 100
```
ChefHutch, line 146: Increased margin for success and reduce frequency of 'out of gas' error when requesting tickets with multicall.

Assume a ticket request will error when requesting 4600 tickets, and a user has 4500 tickets:

If requesting 1000 tickets per call, then the request for 4000 - 5000 will fail. Whereas when only requesting 100 per call, 4400 - 4500 will succeed.

hachiojidev (approved):

PR #1717, opened 2021-07-14
refactor: Split state hooks

RabbitDoge:
Predictions hooks not split because Hachi is working on it


.eslintrc
```diff
@@ -17,6 +17,7 @@
     "no-plusplus": 0,
     "prefer-destructuring": ["warn", { "object": true, "array": false }],
     "no-underscore-dangle": 0,
+    "import/prefer-default-export": 0,
```
hachiojidev, line 20: Finally, someone has the balls to remove this rule.

hachiojidev (approved):

PR #1743, opened 2021-07-16
feat(swap-warning-modals): Add warning modals for specific tokens when swapping

ChefHutch:
Preview: https://exchange-warning-modals.netlify.app/swap
To test, select as either swap SAFEMOON (`0x8076c74c5e3f5852037f31ff0093eeb8c8add8d3`), or BONDLY (`0x96058f8C3e16576D9BD68766f3836d9A33158f89 `)

Steps to add future warnings:
1. Add new entry to `swapWarningTokens` with `symbol` and an `address` object. The key should be similar to the token symbol, lowercased (although honestly this can be anything - it's just used as a key to access the object in step 3). 
```js
  safewat: {
    symbol: 'WAT',
    address: {
      56: '0x...',
      97: '',
    },
  },
```

2. Create a warning component which will be passed into the body of the modal. These are super basic and could maybe even be spun up by non-FE engineers, i.e.
```js
const SafeWatWarning: React.FC<DefaultWarningProps> = () => {
  return (
    <>
      <Text mb="24px">A cluster of stuff about safewat</Text>
      <Text>Good project</Text>
    </>
  )
}
```

3. Add a new object to `TOKEN_WARNINGS` for the new warning
```js
    [getAddress(SwapWarningTokensConfig.safewat.address)]: {
      symbol: SwapWarningTokensConfig.safewat.symbol,
      component: <SafeWatWarning />,
    },

    // 0x...: {
    //   symbol: 'WAT',
    //   component: <SafeWatWarning/>,
    // },
```

Done. The `safewat` warning modal will now fire when a user tries to swap `0x...`

### DIFF
--- a/src/components/SearchModal/CurrencySearchModal.tsx
+++ b/src/components/SearchModal/CurrencySearchModal.tsx
@@ -54,8 +54,8 @@ export default function CurrencySearchModal({
 
   const handleCurrencySelect = useCallback(
     (currency: Currency) => {
-      onCurrencySelect(currency)
       onDismiss()
+      onCurrencySelect(currency)
     },
     [onDismiss, onCurrencySelect],
   )
--- a/src/config/constants/swapWarningTokens.ts
+++ b/src/config/constants/swapWarningTokens.ts
@@ -1,7 +1,7 @@
 import tokens from 'config/constants/tokens'
 import { Address } from './types'
 
-const { bondly } = tokens
+const { bondly, safemoon } = tokens
 
 interface WarningToken {
   symbol: string
@@ -13,13 +13,7 @@ interface WarningTokenList {
 }
 
 const SwapWarningTokens = <WarningTokenList>{
-  safemoon: {
-    symbol: 'SAFEMOON',
-    address: {
-      56: '0x8076C74C5e3F5852037F31Ff0093Eeb8c8ADd8D3',
-      97: '',
-    },
-  },
+  safemoon,
   bondly,
 }
 
--- a/src/config/constants/swapWarningTokens.ts
+++ b/src/config/constants/swapWarningTokens.ts
@@ -1,4 +1,18 @@
-const SwapWarningTokens = {
+import tokens from 'config/constants/tokens'
+import { Address } from './types'
+
+const { bondly } = tokens
+
+interface WarningToken {
+  symbol: string
+  address: Address
+}
+
+interface WarningTokenList {
+  [key: string]: WarningToken
+}
+
+const SwapWarningTokens = <WarningTokenList>{
   safemoon: {
     symbol: 'SAFEMOON',
     address: {
@@ -6,13 +20,7 @@ const SwapWarningTokens = {
       97: '',
     },
   },
-  bondly: {
-    symbol: 'BONDLY',
-    address: {
-      56: '0x96058f8C3e16576D9BD68766f3836d9A33158f89',
-      97: '',
-    },
-  },
+  bondly,
 }
 
 export default SwapWarningTokens
--- a/src/config/constants/swapWarningTokens.ts
+++ b/src/config/constants/swapWarningTokens.ts
@@ -1,0 +1,18 @@
+const SwapWarningTokens = {
+  safemoon: {
+    symbol: 'SAFEMOON',
+    address: {
+      56: '0x8076C74C5e3F5852037F31Ff0093Eeb8c8ADd8D3',
+      97: '',
+    },
+  },
+  bondly: {
+    symbol: 'BONDLY',
+    address: {
+      56: '0x96058f8C3e16576D9BD68766f3836d9A33158f89',
+      97: '',
+    },
+  },
+}
+
+export default SwapWarningTokens
--- a/src/views/Swap/components/ImportTokenWarningModal.tsx
+++ b/src/views/Swap/components/ImportTokenWarningModal.tsx
@@ -8,7 +8,7 @@ interface Props extends InjectedModalProps {
   onCancel: () => void
 }
 
-const TokenWarningModal: React.FC<Props> = ({ tokens, onDismiss, onCancel }) => {
+const ImportTokenWarningModal: React.FC<Props> = ({ tokens, onDismiss, onCancel }) => {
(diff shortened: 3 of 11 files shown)


src/components/SearchModal/CurrencySearchModal.tsx
```diff
@@ -54,8 +54,8 @@ export default function CurrencySearchModal({
 
   const handleCurrencySelect = useCallback(
     (currency: Currency) => {
-      onCurrencySelect(currency)
       onDismiss()
+      onCurrencySelect(currency)
```
ChefHutch, line 58: Clear currency search modal before firing `onCurrencySelect`

src/views/Swap/components/ImportTokenWarningModal.tsx
```diff
@@ -8,7 +8,7 @@ interface Props extends InjectedModalProps {
   onCancel: () => void
 }
 
-const TokenWarningModal: React.FC<Props> = ({ tokens, onDismiss, onCancel }) => {
+const ImportTokenWarningModal: React.FC<Props> = ({ tokens, onDismiss, onCancel }) => {
```
ChefHutch, line 11: Rename `TokenWarningModal` to distinguish it more clearly from the new `SwapWarningModal`

src/views/Swap/components/SwapWarningModal/Acknowledgement.tsx
```diff
@@ -0,0 +1,37 @@
+import React, { useState } from 'react'
+import { useTranslation } from 'contexts/Localization'
+import { Text, Flex, Checkbox, Button } from '@pancakeswap/uikit'
+
+interface AcknowledgementProps {
+  handleContinueClick: () => void
+}
+
+const Acknowledgement: React.FC<AcknowledgementProps> = ({ handleContinueClick }) => {
```
ChefHutch, line 9: Generic Acknowledgement component that can easily be plugged into any token warning

src/views/Swap/components/SwapWarningModal/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+  max-width: 420px;
+  border: 1px ${({ theme }) => theme.colors.warning} solid;
+`
+
+interface SwapWarningModalProps {
+  swapCurrency: WrappedTokenInfo
+  onDismiss?: () => void
+}
+
+// Modal is fired by a useEffect and doesn't respond to closeOnOverlayClick prop being set to false
```
ChefHutch, line 20: Not particularly happy about this - but I can't seem to get `useModal` to respond to its second argument `closeOnOverlayClick` - regardless of the value of this argument, the modal closes on overlay click. Open to suggestions here.

src/views/Swap/components/SwapWarningModal/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+      symbol: SwapWarningTokens.safemoon.symbol,
+      component: <SafemoonWarning onDismiss={onDismiss} />,
+    },
+    [SwapWarningTokens.bondly.address[chainId]]: {
+      symbol: SwapWarningTokens.bondly.symbol,
+      component: <BondlyWarning onDismiss={onDismiss} />,
+    },
+  }
+
+  const SWAP_WARNING = WARNING_DATA[swapCurrency.address]
```
ChefHutch, line 57: Dynamically render component/title based on the `swapCurrency`'s address

src/views/Swap/index.tsx
```diff
@@ -214,10 +216,35 @@ export default function Swap({ history }: RouteComponentProps) {
     setSwapState({ tradeToConfirm: trade, swapErrorMessage, txHash, attemptingTxn })
   }, [attemptingTxn, swapErrorMessage, trade, txHash])
 
+  // swap warning state
+  const [swapWarningCurrency, setSwapWarningCurrency] = useState(null)
+  const [onPresentSwapWarningModal] = useModal(<SwapWarningModal swapCurrency={swapWarningCurrency} />)
+
+  const shouldShowSwapWarning = (swapCurrency) => {
```
ChefHutch, line 223: Logic to see if selected currency's address matches any address within the `warningTokenConfig`

src/views/Swap/components/SwapWarningModal/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+      symbol: SwapWarningTokensConfig.safemoon.symbol,
+      component: <SafemoonWarning onDismiss={onDismiss} />,
+    },
+    [getAddress(SwapWarningTokensConfig.bondly.address)]: {
+      symbol: SwapWarningTokensConfig.bondly.symbol,
+      component: <BondlyWarning onDismiss={onDismiss} />,
+    },
+  }
+
+  const SWAP_WARNING = TOKEN_WARNINGS[swapCurrency.address]
```
ChefHutch, line 57: Dynamically render component/title based on the swapCurrency's address

src/config/constants/swapWarningTokens.ts
```diff
@@ -0,0 +1,18 @@
...
+  bondly: {
+    symbol: 'BONDLY',
+    address: {
+      56: '0x96058f8C3e16576D9BD68766f3836d9A33158f89',
+      97: '',
+    },
+  },
+}
+
+export default SwapWarningTokens
```
hachiojidev, line 18: Is possible to avoid keeping a list like this? For example maybe keeping a list of just the symbol and get the address from a static list somewhere (to avoid maintaining another list)?

ChefHutch: Good point man, I understand the reservations about _another_ list. Part of the spec for this was to use addresses rather than symbols when matching selected token with warning token. So while it would be lighter on the config - I figured an array of uncontextualised addresses would also be different to manage.

I could certainly import `bondly` from our `tokens.ts`, but for some - like safemoon in this example, they aren't guaranteed to exist elsewhere in the codebase.

ChefHutch: Levelled up the config, see what you think.

hachiojidev (approved):

PR #1752, opened 2021-07-18
feat: Add loading indicator for staked only when user data is loaded

memoyil:
To review:

https://deploy-preview-1752--pancakeswap-dev.netlify.app/

To reproduce:

1. Connect your account
2. Go to farms or pools
3. Select staked only fast
4. See there is no loading indicator

### DIFF
--- a/src/components/Loading.tsx
+++ b/src/components/Loading.tsx
@@ -1,0 +1,31 @@
+import styled from 'styled-components'
+
+const Loading = styled.div`
+  border: 8px solid #f3f3f3;
+  border-radius: 50%;
+  border-top: 8px solid #ddd;
+  border-bottom: 8px solid #ddd;
+  width: 20px;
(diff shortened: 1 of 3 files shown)


src/components/Loading.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react'
+import styled from 'styled-components'
+
+const Loading: React.FC = () => <StyledLoading />
```
hachiojidev, line 4: I don't understand why this is here.

memoyil: you mean the decleration?

hachiojidev: Yes, why not use the styled component directly?

memoyil: Done

hachiojidev (approved):

PR #1756, opened 2021-07-19
fix(predictions): UI and predictions view-changing bugs

ChefHutch:
Shooting some bugs in Trello

> _[Prediction] When history pane is loading on android dapp browser and you go back to the betting screen, it automatically sends you back to history pane_

Remove `state.isHistoryPaneOpen = true` side effect from `fetchHistory` action

> _[Prediction] Mobile - after closing history pane the main view doesn't go to current round_

Added `useOnViewChange` hook - to move the slider to the live round index, on view change (which only happens on mobile)

> _[Prediction] Update winnings collection bunny (better resolution)_

Replaced bunny with 2x version

> _[Prediction] remove 1px gaps either side of up/down arrows_

Added -1 margin respectively to each arrow


src/views/Predictions/Positions.tsx
```diff
@@ -22,19 +23,23 @@ const StyledSwiper = styled.div`
     width: 320px;
   }
 `
-const Positions: React.FC = () => {
+const Positions: React.FC<{ view?: string }> = ({ view }) => {
   const { setSwiper } = useSwiper()
   const rounds = useGetSortedRounds()
-  const initialIndex = Math.floor(rounds.length / 2)
+  const currentEpoch = useGetCurrentEpoch()
+  const liveEpoch = currentEpoch > 0 ? currentEpoch - 1 : currentEpoch
```
ChefHutch, line 30: When `useGetCurrentEpoch` returns 0, the live epoch isn't -1, so added some logic to catch and return 0. This may be unnecessary given this scenario may only be true in early testing?

hachiojidev, line 30: Can we avoid calling this "live" - it is already confusing enough that current is actually "next" on the UI.

src/views/Predictions/Positions.tsx
```diff
@@ -22,19 +23,23 @@ const StyledSwiper = styled.div`
     width: 320px;
   }
 `
-const Positions: React.FC = () => {
+const Positions: React.FC<{ view?: string }> = ({ view }) => {
```
hachiojidev, line 26: View needs to be `PageView`

src/views/Predictions/hooks/useOnViewChange.ts
```diff
@@ -0,0 +1,19 @@
+import { useEffect } from 'react'
+import usePrevious from 'hooks/usePreviousValue'
+import useSwiper from './useSwiper'
+
+/**
+ * Hooks for actions to be performed when the view changes (mobile)
+ */
+const useOnViewChange = (liveSwiperIndex: number, view?: string) => {
```
hachiojidev, line 8: `PageView`

hachiojidev (changes requested):

src/views/Predictions/Positions.tsx
ChefHutch: Aha yea I wasn't sure how best to name this. I'll go with `previousRound` for now given it's `currentRound` minus 1.

Would that match the naming convention you've used elsewhere?

src/views/Predictions/Positions.tsx
ChefHutch: Yea also thought this - created a circular dependency importing that enum from `Mobile`. Where would you house non-state-related types? I can just create a types.ts at the view level?

hachiojidev: Lets create a `types.ts` file in `views/Predictions` folder. Not ideal I know...

ChefHutch: Sure thing. Pushed all these tweaks

hachiojidev (approved):

PR #1766, opened 2021-07-21
Chore/PageSection divider unification

ChefHutch:
Preview: https://new-page-sections.netlify.app

- Delete old divider SVG and replace with new path approach
- Implement new path approach across app where PageSection is used
- Refactor and rename props


Chef-Cheems (approved):

src/views/Lottery/index.tsx
```diff
@@ -42,15 +42,15 @@ const Lottery = () => {
...
+      <PageSection background={TITLE_BG} index={1} hasCurvedDivider={false}>
         <Hero />
       </PageSection>
       <PageSection
         containerProps={{ style: { marginTop: '-30px' } }}
         background={GET_TICKETS_BG}
         concaveDivider
-        concaveBackgroundLight="#7645D9"
-        curvePosition="top"
+        clipFill={{ light: '#7645D9' }}
```
hachiojidev, line 52: If I am reading this correctly `clipFill` is supposed to be a `string`.

hachiojidev: Looking at `PageSection/types.ts`

ChefHutch: I think it's all good hachi? It's an object with two keys - light & dark, both of which are strings? 

```
{
  light: string
  dark?: string
}
```

You might have spotted something I've missed though!

hachiojidev: Ah yeah you're right. My bad.

hachiojidev (approved):

PR #1775, opened 2021-07-23
feat: Add C98 farm and pool

RabbitDoge:


Chef-Cheems (approved):

ChefKai (approved):

src/config/constants/pools.ts
```diff
@@ -22,7 +22,7 @@ const pools: PoolConfig[] = [
     earningToken: tokens.c98,
     contractAddress: {
       97: '',
-      56: '0x7D44a25Fb2f4452B2f5c54F3CB577c4a53D75Fab',
+      56: '0x5e49531BA07bE577323e55666D46C6217164119E',
```
fleon, line 25: @RabbitDoge Do you know what was the issue was with the older pool? I'm comparing the contracts, but they look exactly the same to me.

Chef-Cheems: No issue, just had to delay release a bit to later block (and it was already too late to update on previous contract since startBlock has passed)

PR #1777, opened 2021-07-25
fix: Staked only setting persistent storage missing in farms

memoyil:
To review: 

https://deploy-preview-1777--pancakeswap-dev.netlify.app/

To reproduce: 

1. Go to farms
2. Toggle staked only
3. Refresh the page
4. See the new setting is not persistent

This fix also keeps the feature of showing staked only farms when finished farms selected (until user toggles a new setting)

### DIFF
--- a/src/views/Farms/Farms.tsx
+++ b/src/views/Farms/Farms.tsx
@@ -16,6 +16,7 @@ import { getFarmApr } from 'utils/apr'
 import { orderBy } from 'lodash'
 import isArchivedPid from 'utils/farmHelpers'
 import { latinise } from 'utils/latinise'
+import { useUserFarmStakedOnly } from 'state/user/hooks'
 import PageHeader from 'components/PageHeader'
 import SearchInput from 'components/SearchInput'
 import Select, { OptionProps } from 'components/Select/Select'
@@ -133,10 +134,7 @@ const Farms: React.FC = () => {
   // Connected users should see loading indicator until first userData has loaded
   const userDataReady = !account || (!!account && userDataLoaded)
 
-  const [stakedOnly, setStakedOnly] = useState(!isActive)
-  useEffect(() => {
-    setStakedOnly(!isActive)
-  }, [isActive])
+  const [stakedOnly, setStakedOnly] = useUserFarmStakedOnly(isActive)
 
   const activeFarms = farmsLP.filter((farm) => farm.pid !== 0 && farm.multiplier !== '0X' && !isArchivedPid(farm.pid))
   const inactiveFarms = farmsLP.filter((farm) => farm.pid !== 0 && farm.multiplier === '0X' && !isArchivedPid(farm.pid))
--- a/src/state/user/actions.ts
+++ b/src/state/user/actions.ts
@@ -13,6 +13,12 @@ export interface SerializedPair {
   token1: SerializedToken
 }
 
+export enum FarmStakedOnly {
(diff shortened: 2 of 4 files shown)


src/views/Farms/Farms.tsx
```diff
@@ -133,10 +133,15 @@ const Farms: React.FC = () => {
   // Connected users should see loading indicator until first userData has loaded
   const userDataReady = !account || (!!account && userDataLoaded)
 
-  const [stakedOnly, setStakedOnly] = useState(!isActive)
+  const [stakedOnlyPersisted, setStakedOnlyPersisted] = usePersistState(null, {
```
hachiojidev, line 136: Why is the default value `null`?

memoyil: Null indicates that there is no user interaction yet therefore the defaul behaviour will be showing staked only farms when user clicks finished. When user toggles it, that behaviour will be ignored and selection from persist store will be applied

hachiojidev: I don't see why this is better than just adding the default value. You said yourself

> default behaviour which showing staked only farms when user clicks finished will occur.

memoyil: I updated the implementation to make it clean, and now there have a default value in persistent store

hachiojidev: @memoyil Sorry dude I am just not understanding why we have a default value if the default value should be `true` or `false`. I think we want to keep it simple and make staked only as `false`. That is the default state the page should render in if a user first lands on the page without visiting before.

Additionally, the enum you created is confusing. If I understand Typescript's enums `DEFAULT` is 0, `TRUE` is 1, and `FALSE` is 2.

memoyil: Because there is a feature there to show staked only when finished farms selected. In order to keep this i added default option. If user wants to disable the feature they have to toggle it.

RabbitDoge (commented):
`usePersistState` shouldn't be used for settings anymore, we have a `user` reducer that persists between sessions.

src/views/Farms/Farms.tsx
hachiojidev: Okay, I will take your word for it. Please update the code to use the user reducer as @RabbitDoge noted below. Also, can you refactor the enum to be a bit more clear.

memoyil: I added strings to enums to make it more descriptive and changed persist state to user reducer.

memoyil:
> `usePersistState` shouldn't be used for settings anymore, we have a `user` reducer that persists between sessions.

Done

src/views/Farms/Farms.tsx
```diff
@@ -16,6 +16,8 @@ import { getFarmApr } from 'utils/apr'
 import { orderBy } from 'lodash'
 import isArchivedPid from 'utils/farmHelpers'
 import { latinise } from 'utils/latinise'
+import { useUserFarmStakedOnly } from 'state/user/hooks'
+import { FarmStakedOnly } from 'state/user/actions'
```
hachiojidev, line 20: Can you remove this unused import.

hachiojidev (changes requested):

src/views/Farms/Farms.tsx
memoyil: Done

hachiojidev (approved):

PR #1782, opened 2021-07-26
fix(farms): Round farm staking balances below 0.0000001 to 10dp

ChefHutch:
Screenshot 2021-07-26 at 11 01 46

Solved by rounding the staked balance to 10 decimal places when it's below `0.0000001`
Screenshot 2021-07-26 at 11 00 56


hachiojidev (approved):
We should think about making this a global utility.

ChefHutch:
> We should think about making this a global utility.

@hachiojidev I guess the solution might be Bugs' idea for global truncation of long decimals?

PR #1798, opened 2021-07-29
Feature/user menu

hachiojidev:


src/components/Menu/UserMenu/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+import WalletUserMenuItem from './WalletUserMenuItem'
+
+const UserMenu = () => {
+  const { t } = useTranslation()
+  const { account } = useWeb3React()
+  const { logout } = useAuth()
+  const { balance, isFetching } = useGetBnbBalance()
+  const { isInitialized, isLoading, profile } = useProfile()
+  const [onPresentWalletModal] = useModal(<WalletModal initialView={WalletView.WALLET_INFO} />)
+  const [onPresentTransactionModal] = useModal(<WalletModal initialView={WalletView.TRANSACTIONS} />)
```
hachiojidev, line 27: I tried using the 3rd parameter that refreshes props but it just didn't work.

src/hooks/useTokenBalance.ts
```diff
@@ -85,22 +85,25 @@ export const useBurnedBalance = (tokenAddress: string) => {
...
 
 export const useGetBnbBalance = () => {
+  const [isFetching, setIsFetching] = useState(false)
   const [balance, setBalance] = useState(BIG_ZERO)
   const { account } = useWeb3React()
   const { lastUpdated, setLastUpdated } = useLastUpdated()
 
   useEffect(() => {
     const fetchBalance = async () => {
+      setIsFetching(true)
```
hachiojidev, line 95: I added a fetching state because we don't know if the balance is zero because it is zero or just because it is the fallback value.

Chef-Cheems: It could use same `FetchStatus` as `useTokenBalance` above.

hachiojidev: Oh, missed that! Thanks.

src/components/Menu/UserMenu/index.tsx
Chef-Cheems: Hm.... I was trying to test it out locally but it looks its like one of the situations Hutch was having recently - he wanted to pass props for modal during `onPresentModal` invocation (like `onClick={() => onPresentModal({ props: { initialView: WalletView.TX }}) }`, which is not possible right now. Third prop doesn't really work here cause modal props would need to depend on container props or state, which here is a bit tricky cause we'd need to keep some state for what is about to open and then update it on click and immediately open modal...
So I think the way you did it now is OK for now. Me and Doge were talking about rewriting whole `useModal` hook cause its really annoying to rerender, third prop is more of a temporary fix for exchange modals (and as they just receive their props from other effects its working fine now without extra junk around it). When we got to rewriting useModal we could also make it possible to reset some props during `onPresentModal` so you could declaratively change what is about to open at the point of opening.

src/components/Menu/UserMenu/WalletModal.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { useTranslation } from 'contexts/Localization'
+import styled from 'styled-components'
+import { FetchStatus, useGetBnbBalance } from 'hooks/useTokenBalance'
+import WalletInfo from './WalletInfo'
+import WalletTransactions from './WalletTransactions'
+
+export enum WalletView {
+  WALLET_INFO = 0,
+  TRANSACTIONS = 1,
+}
```
Chef-Cheems, line 24: nitpick - in case enum values are `0,1,2,3,etc` they can be ommited
```typescript
export enum WalletView {
  WALLET_INFO,
  TRANSACTIONS,
}
```

Chef-Cheems (approved):

PR #1803, opened 2021-07-29
fix: Numberformat compact not shown properly in mobile safari (below 14.7)

memoyil:
To review:

https://deploy-preview-1803--pancakeswap-dev.netlify.app/

To reproduce:

1. Go to homepage with mobile Safari
2. Check the numbers in Used by millions. Trusted with billions.
3. See that compact option could not be applied there

It is not supported in mobile safari until 14.7
https://caniuse.com/?search=numberformat%20notation

Before:

CF8DEAE6-D94B-4CC4-9D60-A0D5DCB9F4A1

After:

image

### DIFF
--- a/src/views/Home/components/MetricsSection/StatCardContent.tsx
+++ b/src/views/Home/components/MetricsSection/StatCardContent.tsx
@@ -1,11 +1,13 @@
 import React from 'react'
-import { Heading, Flex, Text } from '@pancakeswap/uikit'
+import { Heading, Flex, Text, useMatchBreakpoints } from '@pancakeswap/uikit'
 
 const StatCardContent: React.FC<{ headingText: string; bodyText: string; highlightColor: string }> = ({
   headingText,
   bodyText,
   highlightColor,
 }) => {
+  const { isXl } = useMatchBreakpoints()
+  const isMobile = !isXl
   const split = headingText.split(' ')
   const lastWord = split.pop()
   const remainingWords = split.slice(0, split.length).join(' ')
@@ -19,7 +21,11 @@ const StatCardContent: React.FC<{ headingText: string; bodyText: string; highlig
       justifyContent="flex-end"
       mt={[null, null, null, '64px']}
     >
-      <Heading scale="xl">{remainingWords}</Heading>
+      {isMobile && remainingWords.length > 13 ? (
+        <Heading scale="lg">{remainingWords}</Heading>
+      ) : (
+        <Heading scale="xl">{remainingWords}</Heading>
+      )}
       <Heading color={highlightColor} scale="xl" mb="24px">
         {lastWord}
       </Heading>


memoyil:
@hachiojidev I got an error intermittently while loading the page, thats why i marked as WiP.  I am on it.

hachiojidev:
@memoyil Please use a draft for WIP PR's in the future.

memoyil:
@hachiojidev Ready for your review

src/utils/polyfill.ts
```diff
@@ -0,0 +1,97 @@
+import { shouldPolyfill } from '@formatjs/intl-numberformat/should-polyfill'
+
+export async function polyfill(locale: string) {
```
hachiojidev, line 3: Please use the same code style that we use throughout the site. (arrow function)

hachiojidev (changes requested):

ChefHutch:
Thanks for taking a look at this Memoyil - we were happy to accept that users with pre-14.7 Safari would see the full numbers, but it would be nice to get it fixed. 

Performance is very poor when language switching after this change though - reverting to a previously selected language as it switches to a new one. Please can you investigate?

https://user-images.githubusercontent.com/79279477/127975247-061b8727-06fe-4c7b-8e69-a2e64aa8f4c5.mov

memoyil:
> Thanks for taking a look at this Memoyil - we were happy to accept that users with pre-14.7 Safari would see the full numbers, but it would be nice to get it fixed.
> 
> Performance is very poor when language switching after this change though - reverting to a previously selected language as it switches to a new one. Please can you investigate solutions?
> perf.mov

I am on it now

src/utils/polyfill.ts
memoyil: Done

memoyil:
@ChefHutch The root cause of the issue is using dynamic import to import packages for languages for polyfill, it is an async operation and while it is being loaded it switches to default language. Therefore I decided to remove the polyfill implementation and create our own which uses our translations.

src/utils/formatBalance.ts
```diff
@@ -62,13 +63,53 @@ export const formatFixedNumber = (number: ethers.FixedNumber, displayDecimals =
...
+  }
+  if (num >= 1000) {
+    return `${(num / 1000).toFixed(1).replace(/\.0$/, '')} ${t('Thousand').toLocaleLowerCase()}`
+  }
+  return num
+}
+
+const supportsES2020 = () => {
+  try {
+    const s = new Intl.NumberFormat('en', {
```
hachiojidev, line 99: Please use a more descriptive variable name.

hachiojidev (changes requested):

src/utils/formatBalance.ts
```diff
@@ -62,13 +63,53 @@ export const formatFixedNumber = (number: ethers.FixedNumber, displayDecimals =
...
+    return `${(num / 1000000000).toFixed(1).replace(/\.0$/, '')} ${t('Billion').toLocaleLowerCase()}`
+  }
+  if (num >= 1000000) {
+    return `${(num / 1000000).toFixed(1).replace(/\.0$/, '')} ${t('Million').toLocaleLowerCase()}`
+  }
+  if (num >= 1000) {
+    return `${(num / 1000).toFixed(1).replace(/\.0$/, '')} ${t('Thousand').toLocaleLowerCase()}`
+  }
+  return num
+}
```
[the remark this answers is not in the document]
memoyil: True, thats why i keep the original numberformat for new versions of safari. I thought it is better to show some kind of formatted text that fits many languages instead of number. Or we can have the polyfill approach with the issue that @ChefHutch showed in video.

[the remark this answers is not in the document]
ChefHutch: I'm not sure this is any better Memoyil?

We're going from a correctly localised string - albeit the full currency number, to an incorrectly localised text number.

Personally I think the fallback behaviour (i.e. showing the full number) is preferable to an incorrectly localised bit of text, that brings with it this logic weight.

If it isn't possible to make the polyfilling performant - let's leave the code as-is.
If there are display issues with showing the full number - let's fix them.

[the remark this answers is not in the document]
memoyil: Done 👍 Can you review it again?

src/utils/formatBalance.ts
memoyil: Removed, not needed in new approach

PR #1822, opened 2021-08-02
feat: Automatically get LP addresses for auction farms

Chef-Cheems:


src/config/constants/types.ts
```diff
@@ -1,4 +1,5 @@
 import BigNumber from 'bignumber.js'
+import { Token as SDKToken } from '@pancakeswap/sdk'
```
Chef-Cheems, line 2: There is another `Token` type defined below and it is not the same as `Token` in SDK (in SDK symbol is optional)

src/config/constants/farmAuctions.ts
```diff
@@ -1,130 +1,182 @@
...
+  const token0asToken = new Token(ChainId.MAINNET, token0, 18)
+  return Pair.getAddress(token0asToken, token1)
+}
+
 export const whitelistedBidders: FarmAuctionBidderConfig[] = [
   {
     account: '0x9Ed5a62535A5Dd2dB2d9bB21bAc42035Af47F630',
     farmName: 'NAV-BNB',
+    tokenAddress: '0xbfef6ccfc830d3baca4f6766a0d4aaa242ca9f3d',
+    quoteToken: WBNB,
```
ChefKai, line 15: Think it would make more sense to re-use the same `tokens` as the farms config use (since the project is `farm-auctions`); so `tokens.wbnb` and `tokens.busd`

src/config/constants/farmAuctions.ts
```diff
@@ -1,130 +1,182 @@
...
+    tokenAddress: '0xe792f64C582698b8572AAF765bDC426AC3aEfb6B',
+    quoteToken: WBNB,
     tokenName: 'Swirge Network',
     projectSite: 'https://swirge.com/',
   },
   {
     account: '0x58092273a044D6e1d23B5095AE873F6E24E906ed',
     farmName: 'rUSD-BUSD',
+    tokenAddress: '0x07663837218a003e66310a01596af4bf4e44623d',
+    quoteToken: BUSD[ChainId.MAINNET],
```
ChefKai, line 39: Maybe look at creating a `getAddress` helper for `Token` entity (same as for `Address` entity); unless doing the above which wouldn't be necessary

src/config/constants/farmAuctions.ts
hachiojidev: I think this is a good idea.

hachiojidev (changes requested):

ChefKai (approved):

PR #1829, opened 2021-08-03
feat(home): Get farm and pools data on HP 

ChefHutch:
- Preview https://farm-pool-row.netlify.app/
- Add `<FarmsPoolsRow>` to Home page 'Earn' section
Screenshot 2021-08-04 at 14 28 10

- Fix some lil Home ui bugs


src/views/Home/hooks/useGetTopFarmsByApr.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const dispatch = useAppDispatch()
+  const { data: farms } = useFarms()
+  const [fetchStatus, setFetchStatus] = useState(FetchStatus.NOT_FETCHED)
+  const [topFarms, setTopFarms] = useState<FarmWithStakedValue[]>([null, null, null, null, null])
+
+  const cakePriceBusdAsString = usePriceCakeBusd().toString()
+
+  const cakePriceBusd = useMemo(() => {
+    return new BigNumber(cakePriceBusdAsString)
+  }, [cakePriceBusdAsString])
```
hachiojidev, line 29: This already returns a `BigNumber`

src/views/Home/hooks/useGetTopPoolsByApr.tsx
```diff
@@ -0,0 +1,80 @@
...
+    const cakeAutoVault = { ...cakePool, isAutoVault: true }
+    const cakeAutoVaultWithApr = { ...cakeAutoVault, apr: getAprData(cakeAutoVault, performanceFeeAsDecimal).apr }
+    return [cakeAutoVaultWithApr, ...poolsWithoutAutoVault]
+  }, [poolsWithoutAutoVault, performanceFeeAsDecimal])
+
+  const cakePriceBusdAsString = usePriceCakeBusd().toString()
+
+  const cakePriceBusd = useMemo(() => {
+    return new BigNumber(cakePriceBusdAsString)
+  }, [cakePriceBusdAsString])
```
hachiojidev, line 45: Same here

src/views/Home/hooks/useGetTopPoolsByApr.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const { pools: poolsWithoutAutoVault } = useSelector((state: State) => ({
+    pools: state.pools.data,
+    userDataLoaded: state.pools.userDataLoaded,
+  }))
+  const {
+    fees: { performanceFee },
+  } = useCakeVault()
+  const performanceFeeAsDecimal = performanceFee && performanceFee / 100
+  const [fetchStatus, setFetchStatus] = useState(FetchStatus.NOT_FETCHED)
+  const [topPools, setTopPools] = useState<Pool[]>([null, null, null, null, null])
```
hachiojidev, line 31: Why the `null` placeholders?

hachiojidev (changes requested):

src/views/Home/hooks/useGetTopFarmsByApr.tsx
ChefHutch: The memoisation is required to pass `cakePriceBusd` into a dependency array. The BN returned by `usePriceCakeBusd`'s reference changes on each render even when the value is the same, so it triggers a rerender loop if used in a dep array.

I'm not sure if there's a better way to achieve this though? I've had to implement this kind of workaround in a few places to prevent rerenders. 

- https://github.com/pancakeswap/pancake-frontend/blob/7c6c60215e1fb62844a401251d339d21a577c7f0/src/state/pools/hooks.ts#L85-L111
- https://github.com/pancakeswap/pancake-frontend/blob/6d22cd2b36fa9bcad8f5151fd391ff24c70b551b/src/state/lottery/helpers.ts#L147-L157
- https://github.com/pancakeswap/pancake-frontend/blob/d932319d2d1dbeba5d0580ee6fce0340ef498cc1/src/views/Home/components/WinSection/LotteryCardContent.tsx#L35-L37

src/views/Home/hooks/useGetTopPoolsByApr.tsx
ChefHutch: So that I can hold all of the 'loading' state management within the `TopFarmPool` component (eugh that name.. shudder), while still mapping over that array to render the components.

The alternative as I saw it would be to declare a null state and render a separate 'loading' component with 5 different loading skeletons in it until that null state was populated. 

I think the way I have it requires less code, but is slightly less intuitive?

src/views/Home/hooks/useGetTopFarmsByApr.tsx
hachiojidev: If that is the case lets just do it in the hook itself.

ChefHutch: Sure thing. Done 👍

src/views/Home/components/MetricsSection/index.tsx
```diff
@@ -8,9 +8,9 @@ import IconCard, { IconCardData } from '../IconCard'
 import StatCardContent from './StatCardContent'
 import GradientLogo from '../GradientLogoSvg'
 
-// Values fetched from bitQuery effective 28/7/21
-const txCount = 30673865
-const addressCount = 1966700
+// Values fetched from bitQuery effective 5/8/21
+const txCount = 37553080
+const addressCount = 2279782
```
hachiojidev, line 13: Sorry if I missed this, but why are these hard-coded in here?

Chef-Chungus: @hachiojidev - we're waiting on Debank to add these to their API, I approved this being hardcoded until they add this

ChefHutch: We're waiting on Debank providing this info as part of their api response (that we currently use for TVL). So until then it's hard coded data coming from a bitquery response.

hachiojidev (approved):

PR #1841, opened 2021-08-04
chore(lottery): Apply lotteryBuyInstant ID to correct button

ChefHutch:


src/views/Lottery/components/BuyTicketsModal/BuyTicketsModal.tsx
```diff
@@ -416,26 +416,26 @@ const BuyTicketsModal: React.FC<BuyTicketsModalProps> = ({ onDismiss }) => {
...
               <Button
-                id="lotteryBuyInstant"
                 variant="secondary"
                 mt="8px"
-                disabled={disableBuying || isConfirming}
-                onClick={() => {
-                  setBuyingStage(BuyingStage.EDIT)
-                }}
-              >
-                <Flex alignItems="center">
```
ChefHutch, line 430: - Remove unnecessary Flex within Button child

src/views/Profile/components/ApproveConfirmButtons.tsx
```diff
@@ -67,6 +68,7 @@ const ApproveConfirmButtons: React.FC<ApproveConfirmButtonsProps> = ({
   onConfirm,
   buttonArrangement = ButtonArrangement.ROW,
   confirmLabel,
+  confirmId,
```
ChefHutch, line 71: `ApproveConfirmButtons` receive `confirmId` that is passed as an id to the `confirm` buttons

src/views/Profile/components/ApproveConfirmButtons.tsx
```diff
@@ -17,6 +17,7 @@ interface ApproveConfirmButtonsProps {
   onConfirm: () => void
   buttonArrangement?: ButtonArrangement
   confirmLabel?: string
+  confirmId?: string
```
hachiojidev, line 20: Why not use the native `id` prop already there?

ChefHutch: Ah - because the `ApproveConfirmButtons` have multiple buttons within them - an approve, and a confirm. This id is only passed to the confirm buttons

hachiojidev (approved):

PR #1884, opened 2021-08-11
refactor: Merge multicall

hachiojidev:


ChefHutch (approved):
Nice one!

Probably something to be said for looking through the different functions within `/state/multicall` and seeing what we can consolidate with our multicall util, but that's a bigger job for less reward.

RabbitDoge (approved):

PR #1887, opened 2021-08-11
fix: (homepage) Harvest card text shows collect farm info even when only manual cake pool staked

memoyil:
To review:

https://deploy-preview-1887--pancakeswap-dev.netlify.app/

To reproduce:

Manual Cake pool also comes with farmsWithStaked data we should seperate it to show accurate info.

1. Go to homepage
2. Connect wallet with only has cake pool and no farms staked (or both)
3. Check text in harvest card at top
4. Text says you have farm or farms to collect but should contain information about CAKE pool instead.


src/views/Home/components/UserBanner/HarvestCard.tsx
```diff
@@ -24,13 +24,31 @@ const HarvestCard = () => {
...
+  } else if (numFarmsToCollect > 0) {
+    if (numFarmsToCollect > 1) {
+      earningsText = t('%earningsBusd% to collect from %count% farms', data)
+    } else {
+      earningsText = t('%earningsBusd% to collect from %count% farm', data)
+    }
+  } else if (hasCakePoolToCollect) {
+    earningsText = t('%earningsBusd% to collect from CAKE pool', data)
+  }
   const [preText, toCollectText] = earningsText.split(earningsBusd.toString())
```
Chef-Cheems, line 52: I just noticed that `preText` is always gonna be empty string (cause split does not return the separator in the resulting array). 
So this line can be replaced with
```javascript
const [, toCollectText] = earningsText.split(earningsBusd.toString())
```
And this block below can be removed (cause a) it never renders b) there is no point in it anyway, the Balance is showing the number)
```jsx
{preText && (
  <Text mb="4px" color="textSubtle">
    {preText}
  </Text>
)}
```

Can you do it in this PR please? 🙂

src/views/Home/components/UserBanner/HarvestCard.tsx
```diff
@@ -24,13 +24,31 @@ const HarvestCard = () => {
   const masterChefContract = useMasterchef()
   const cakePriceBusd = usePriceCakeBusd()
   const earningsBusd = new BigNumber(farmEarningsSum).multipliedBy(cakePriceBusd)
-  const numFarmsToCollect = farmsWithStakedBalance.length
+  const numTotalToCollect = farmsWithStakedBalance.length
+  const numFarmsToCollect = farmsWithStakedBalance.filter((value) => value.pid !== 0).length
+  const hasCakePoolToCollect = numTotalToCollect - numFarmsToCollect > 0
 
-  const earningsText = t('%earningsBusd% to collect from %count% %farms%', {
+  let earningsText = ''
```
Chef-Cheems, line 31: This will show empty text, so just show `$0` with no context (while loading or when user has no stakes).
It could be changed to this:
```javascript
let earningsText = t('0 to collect') // needs to be added to translations.json
```
And then in the code below some change is needed too because `earningsBusd.toString()` might be `NaN` while loading. If user has real `0` `earningsBusd.toString()` will be `0` (just confirmed), so no need to check for that.
```javascript
const [preText, toCollectText] = earningsText.split(earningsBusd.isNaN() ? '0' : earningsBusd.toString())
```
and `preText` is not really shown so it doesn't matter, this `0` will never be displayed (see other comments about removing `preText`)

Chef-Cheems (changes requested):
Thanks!
Couple of minor details, but after that its good to go :)

To be honest the whole `%earningsBusd% to collect...` seems a bit redundant since we never actually render `%earningsBusd%` part. I'm afraid it might cause troubles in non-English language, but let's keep it this way for now.

src/views/Home/components/UserBanner/HarvestCard.tsx
memoyil: But for example in japan translation %earningsBusd% is in the middle of the translated text, preText should not be empty string there.

`"%count%%farms%から%earningsBusd%を回収"`

Chef-Cheems: In case of those languages pretext wouldn't work anyway, cause the logic works only for English, heh.

memoyil: Why? It is working right now I think

src/views/Home/components/UserBanner/HarvestCard.tsx
memoyil: Done

src/views/Home/components/UserBanner/HarvestCard.tsx
memoyil: Then I am waiting for your approval :)

src/views/Home/components/UserBanner/HarvestCard.tsx
```diff
@@ -24,14 +24,32 @@ const HarvestCard = () => {
...
+    }
+  } else if (numFarmsToCollect > 0) {
+    if (numFarmsToCollect > 1) {
+      earningsText = t('%earningsBusd% to collect from %count% farms', data)
+    } else {
+      earningsText = t('%earningsBusd% to collect from %count% farm', data)
+    }
+  } else if (hasCakePoolToCollect) {
+    earningsText = t('%earningsBusd% to collect from CAKE pool', data)
+  }
```
hachiojidev, line 51: It is unfortunate our translation system doesn't have an intuitive way to handle plurals. Can you move this to a helper file or somewhere out of this component. This is quite a lot of logic to just handle text.

hachiojidev (changes requested):

src/views/Home/components/UserBanner/HarvestCard.tsx
memoyil: Done

hachiojidev (approved):

PR #1893, opened 2021-08-11
Feature/lottery nft

ChefHutch:
Preview: https://lottery-nft.netlify.app/collectibles

### Prior to launch:

- [ ] Needs mainnet contract deploying and then adding to `contracts` - Doge/Nyan to deploy, Hachi or Cheems to add to PR
- [x] Community testing [_in progress_] - Nina to ✅ 

_note - achievements haven't been tested_


src/views/Collectibles/hooks/useBunnySpecialLottery.tsx
```diff
@@ -0,0 +1,217 @@
...
+  const genericClaimChecks = useCallback(
+    async (variationId: number | string): Promise<boolean> => {
+      let hasClaimed = false
+      if (!userRounds) {
+        return false
+      }
+      try {
+        hasClaimed = await lotteryNftContract.hasClaimed(account, variationId)
+      } catch (error) {
+        console.error(`Failed to check hasClaimed for id ${variationId}`, error)
```
hachiojidev, line 111: What happens if this error is thrown? Maybe return true or false?

ChefHutch: Nice catch 👍

src/views/Collectibles/index.tsx
```diff
@@ -13,6 +15,12 @@ const StyledHero = styled.div`
 
 const Collectibles = () => {
   const { t } = useTranslation()
+  const dispatch = useAppDispatch()
+
+  useEffect(() => {
+    // get current lottery ID - used for lottery NFT checks
+    dispatch(fetchCurrentLotteryId())
+  }, [dispatch])
```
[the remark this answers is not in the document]
ChefHutch: Ah - actually, great flag. I will need to move this to the `useBunnySpecialLottery` hook as the global check could be fired from any page and will need the currentLotteryId.

src/views/Collectibles/components/ClaimNftModal.tsx
```diff
@@ -30,14 +30,16 @@ const ClaimNftModal: React.FC<ClaimNftModalProps> = ({ nft, onSuccess, onClaim,
...
 
   const handleConfirm = async () => {
-    const tx = await onClaim()
     setIsConfirming(true)
-    const receipt = await tx.wait()
-    if (receipt.status) {
-      toastSuccess(t('Successfully claimed!'))
-      onDismiss()
-      onSuccess()
-    } else {
```
ChefHutch, line 40: In this previous pattern the `else` condition isn't hit when the tx fails, so the fail state functions never trigger.

src/views/Collectibles/components/GlobalCheckClaimStatus.tsx
```diff
@@ -18,35 +19,56 @@ interface GlobalCheckClaimStatusProps {
...
   const { account } = useWeb3React()
   const { pathname } = useLocation()
+  const { canClaimBaller, canClaimLottie, canClaimLucky } = useBunnySpecialLottery()
 
   // Check claim status
   useEffect(() => {
     const fetchClaimStatus = async () => {
-      const canClaim = await easterNftContract.canClaim(account)
-      setIsClaimable(canClaim)
+      const claimable: Nft[] = []
```
ChefHutch, line 31: Modifications to `GlobalCheckClaimStatus` for the lottery NFTs

src/views/Collectibles/components/NftGiveawayModal.tsx
```diff
@@ -32,25 +29,12 @@ const showConfetti = () => {
...
-  const identifier = Object.keys(teamNftMap).find(
-    (mapNftIdentifier) => teamNftMap[mapNftIdentifier] === profile.team.id,
-  )
-  return nftList.find((nft) => nft.identifier === identifier)
+interface NftGiveawayModalProps extends InjectedModalProps {
+  nfts: Nft[]
 }
 
-const NftGiveawayModal: React.FC<InjectedModalProps> = ({ onDismiss }) => {
+const NftGiveawayModal: React.FC<NftGiveawayModalProps> = ({ onDismiss, nfts }) => {
```
ChefHutch, line 36: Modifying the GiveAway modal to receive an array of nfts and render their images. Any `can user claim nft` logic is to be handled within the `GlobalCheckClaimStatus` hook

src/views/Profile/components/ClaimGiftModal.tsx
```diff
@@ -22,8 +22,12 @@ export const useCanClaim = () => {
   useEffect(() => {
     const fetchClaimStatus = async () => {
       const claimRefundContract = getClaimRefundContract()
-      const walletCanClaim = await claimRefundContract.canClaim(account)
```
ChefHutch, line 25: This function throws an error on first load on **testnet** (the error is thrown on `develop` testnet too)- 

```
Unhandled Rejection (Error): call revert exception (method="canClaim(address)", errorArgs=null, errorName=null, errorSignature=null, reason=null, code=CALL_EXCEPTION, version=abi/5.4.0)
```

The error is handled safely by wrapping it in a try/catch.

src/config/constants/campaigns.ts
```diff
@@ -180,6 +180,27 @@ const campaigns: Campaign[] = [
     description: 'Played a round of Prediction before round 12,120',
     badge: 'clairvoyant.svg',
   },
+  {
```
ChefHutch, line 183: Note that the achievements are also being added in this PR

src/views/Collectibles/helpers.ts
```diff
@@ -0,0 +1,137 @@
+import { getBunnySpecialLotteryContract } from 'utils/contractHelpers'
```
hachiojidev, line 1: I split up the functions to get the 3 different claimable statuses here.

src/views/Collectibles/hooks/useBunnySpecialLottery.tsx
```diff
@@ -0,0 +1,103 @@
...
+      const userParticipation = await getGraphLotteryUser(account, undefined, undefined, {
+        lottery_in: lotteryIdsArray,
+      })
+      setUserRounds(userParticipation.rounds)
+    }
+
+    if (account && currentLotteryIdNum && !userRounds) {
+      getUserData()
+    }
+  }, [currentLotteryIdNum, userRounds, account])
```
hachiojidev, line 66: Before this was two separate effects - however one was dependent on the other so I combined them for readability.

RabbitDoge (approved):

PR #1923, opened 2021-08-16
fix: Farm auction banner text not visible overlaps with background image

memoyil:
To review:

https://deploy-preview-1923--pancakeswap-dev.netlify.app/

To reproduce: 

1. Go to homepage in mobile
2. See farm auction text last part overlaps with background image and it is hard to read

Before:

image

After:

image

It also fixes the formatting issue when browser width is around 576px (sm breakpoint)

image

### DIFF
--- a/src/views/Home/components/Banners/FarmAuctionsBanner.tsx
+++ b/src/views/Home/components/Banners/FarmAuctionsBanner.tsx
@@ -5,10 +5,17 @@ import { useTranslation } from 'contexts/Localization'
 
 const StyledSubheading = styled(Heading)`
   background: -webkit-linear-gradient(#ffd800, #eb8c00);
-  font-size: 24px;
+  font-size: 20px;
   background-clip: text;
   -webkit-background-clip: text;
   -webkit-text-fill-color: transparent;
+  -webkit-text-stroke: 1px rgba(0, 0, 0, 0.5);
+  ${({ theme }) => theme.mediaQueries.xs} {
+    font-size: 24px;
+  }
+  ${({ theme }) => theme.mediaQueries.sm} {
+    -webkit-text-stroke: unset;
+  }
   margin-bottom: 8px;
 `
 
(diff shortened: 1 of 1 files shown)


src/views/Home/components/Banners/FarmAuctionsBanner.tsx
```diff
@@ -9,6 +9,12 @@ const StyledSubheading = styled(Heading)`
   background-clip: text;
   -webkit-background-clip: text;
   -webkit-text-fill-color: transparent;
+  ${({ theme }) => theme.mediaQueries.xs} {
+    -webkit-text-stroke: 1px rgba(0, 0, 0, 0.5);
```
Chef-Cheems, line 13: Maybe text-shadow would be better? Also seems to have support for more browsers.

src/views/Home/components/Banners/FarmAuctionsBanner.tsx
```diff
@@ -44,7 +50,7 @@ const LeftWrapper = styled(Flex)`
   flex-direction: column;
   justify-content: center;
 
-  ${({ theme }) => theme.mediaQueries.sm} {
+  ${({ theme }) => theme.mediaQueries.md} {
```
Chef-Cheems, line 53: If not too difficult - can you please make some fix for iPhone 5-kind screens (320px)? 🙂 
Screenshot 2021-08-16 at 13 55 25

memoyil: Done

src/views/Home/components/Banners/FarmAuctionsBanner.tsx
memoyil: If text-shadow is used, we need to use color instead of background therefore we will lose gradient. And the compliance code is a bit long that you can check here https://codepen.io/Dzorogh/pen/YPeMKd

Text-stroke is also supported by major browsers. https://caniuse.com/?search=text%20stroke

hachiojidev (approved):

Chef-Cheems (approved):

PR #1937, opened 2021-08-18
fix: (harvestcard) NaN shown in harvest card when earnings being loaded

memoyil:
When earnings amount is being loaded but user staked farms data available, depending on the time that query returns, user will see NaN on the screen. (usually it is couple of milliseconds but it can be noticed)

To review: 

https://deploy-preview-1937--pancakeswap-dev.netlify.app/

To reproduce:

1. Go to homepage
2. Restart the page
3. Check harvest card NaN is visible couple of seconds or half a second


src/views/Home/components/UserBanner/HarvestCard.tsx
```diff
@@ -60,7 +60,7 @@ const HarvestCard = () => {
                 {preText}
               </Text>
             )}
-            {earningsBusd && !earningsBusd.isNaN() ? (
```
memoyil, line 63: Don't need to be checked since it is a bignumber and cannot be null

hachiojidev (approved):

PR #1946, opened 2021-08-19
fix: (pools) Auto cake pool and manual one are in the wrong order in total staked

memoyil:
To review:

To reproduce:

1. Go to pools
2. Order by total staked
3. Auto cake vault should be the first instead of second


src/views/Pools/index.tsx
```diff
@@ -195,7 +196,18 @@ const Pools: React.FC = () => {
       case 'totalStaked':
         return orderBy(
           poolsToSort,
-          (pool: Pool) => (pool.isAutoVault ? totalCakeInVault.toNumber() : pool.totalStaked.toNumber()),
+          (pool: Pool) => {
+            let totalStaked
+            if (pool.isAutoVault) {
+              totalStaked = getBalanceNumber(totalCakeInVault, pool.stakingToken.decimals)
+            } else if (pool.sousId === 0) {
+              const manualCakeTotalMinusAutoVault = new BigNumber(pool.totalStaked).minus(totalCakeInVault)
```
hachiojidev, line 204: We are trying to deprecate BigNumber.js. Can you rewrite this to use ether's BigNumber.

hachiojidev (changes requested):

src/views/Pools/index.tsx
memoyil: Done

src/utils/formatBalance.ts
```diff
@@ -15,6 +15,10 @@ export const getBalanceAmount = (amount: BigNumber, decimals = 18) => {
   return new BigNumber(amount).dividedBy(BIG_TEN.pow(decimals))
 }
 
+export const getEthersBalanceAmount = (amount: ethers.BigNumber, decimals = 18) => {
+  return ethers.BigNumber.from(amount).div(ethers.BigNumber.from(10).pow(decimals))
+}
```
hachiojidev, line 20: A couple things:

1. The function accepts a BigNumber then converts it into another BigNumber.

2. Please check https://docs.ethers.io/v5/api/utils/display-logic/

hachiojidev (changes requested):

src/utils/formatBalance.ts
memoyil: Done, I removed the function and use ethers formatUtils now

PR #1967, opened 2021-08-23
refactor: (pool) Dont fetch user farm data in pools

memoyil:
To review:

https://deploy-preview-1967--pancakeswap-dev.netlify.app/

1. Go to pools
2. Connect your account
3. Check there is no difference

Amount of requests is reduced

### DIFF
--- a/src/state/farms/hooks.ts
+++ b/src/state/farms/hooks.ts
@@ -10,7 +10,19 @@ import useRefresh from 'hooks/useRefresh'
 import { fetchFarmsPublicDataAsync, fetchFarmUserDataAsync, nonArchivedFarms } from '.'
 import { State, Farm, FarmsState } from '../types'
 
-export const usePollFarmsData = (includeArchive = false) => {
+export const usePollFarmsPublicData = (includeArchive = false) => {
+  const dispatch = useAppDispatch()
+  const { slowRefresh } = useRefresh()
+
+  useEffect(() => {
+    const farmsToFetch = includeArchive ? farmsConfig : nonArchivedFarms
+    const pids = farmsToFetch.map((farmToFetch) => farmToFetch.pid)
+
+    dispatch(fetchFarmsPublicDataAsync(pids))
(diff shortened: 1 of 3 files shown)


src/state/farms/hooks.ts
```diff
@@ -13,6 +13,18 @@ import { State, Farm, FarmsState } from '../types'
...
+    const farmsToFetch = includeArchive ? farmsConfig : nonArchivedFarms
+    const pids = farmsToFetch.map((farmToFetch) => farmToFetch.pid)
+
+    dispatch(fetchFarmsPublicDataAsync(pids))
+  }, [includeArchive, dispatch, slowRefresh])
+}
+
+export const usePollFarmsWithUserData = (includeArchive = false) => {
+  const dispatch = useAppDispatch()
+  const { slowRefresh } = useRefresh()
```
hachiojidev, line 27: I don't understand how this reduces the number of calls. It does not fetch user data unless there is an account, and if there is an account then we want to fetch user data.

memoyil: In Pools page we dont need user farm data instead we only need public farm data for prices. Farms with user data hook will be used in Farms and the other one will be used used in Pools page.

src/state/farms/hooks.ts
```diff
@@ -13,6 +13,18 @@ import { State, Farm, FarmsState } from '../types'
 export const usePollFarmsData = (includeArchive = false) => {
```
ChefHutch, line 13: Let's rename this one `usePollFarmsPublicData`?

src/state/farms/hooks.ts
ChefHutch: I think you're right Memoyil - we aren't interested in the users' approval state/amount staked/user balance/earnings for farms while on pools. This could be a fairly quick performance win.

src/state/farms/hooks.ts
memoyil: Done

hachiojidev (approved):

PR #1968, opened 2021-08-23
fix: (poll block) Windows visible event triggers rerender app every time

memoyil:
To review:

https://deploy-preview-1968--pancakeswap-dev.netlify.app/

To reproduce:

1. Go to website (especially on desktop)
2. Switch between tabs
3. See when it switches, it renders the all app and sometimes makes it unresponsive

### DIFF
--- a/src/contexts/RefreshContext.tsx
+++ b/src/contexts/RefreshContext.tsx
@@ -1,52 +1,34 @@
-import React, { useState, useEffect, useRef } from 'react'
+import React, { useState, useEffect } from 'react'
+import useIsWindowVisibleRef from 'hooks/useIsWindowVisibleRef'
 
 const FAST_INTERVAL = 10000
 const SLOW_INTERVAL = 60000
 
 const RefreshContext = React.createContext({ slow: 0, fast: 0 })
 
-// Check if the tab is active in the user browser
-const useIsBrowserTabActive = () => {
-  const isBrowserTabActiveRef = useRef(true)
-
-  useEffect(() => {
-    const onVisibilityChange = () => {
-      isBrowserTabActiveRef.current = !document.hidden
-    }
-
-    window.addEventListener('visibilitychange', onVisibilityChange)
-
-    return () => {
-      window.removeEventListener('visibilitychange', onVisibilityChange)
-    }
-  }, [])
-
-  return isBrowserTabActiveRef
-}
-
 // This context maintain 2 counters that can be used as a dependencies on other hooks to force a periodic refresh
 const RefreshContextProvider = ({ children }) => {
   const [slow, setSlow] = useState(0)
   const [fast, setFast] = useState(0)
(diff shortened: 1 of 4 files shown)


src/contexts/RefreshContext.tsx
```diff
@@ -1,52 +1,34 @@
...
 const RefreshContextProvider = ({ children }) => {
   const [slow, setSlow] = useState(0)
   const [fast, setFast] = useState(0)
-  const isBrowserTabActiveRef = useIsBrowserTabActive()
+  const isWindowVisibleRef = useIsWindowVisibleRef()
 
   useEffect(() => {
     const interval = setInterval(async () => {
-      if (isBrowserTabActiveRef.current) {
+      if (isWindowVisibleRef.current) {
```
hachiojidev, line 17: I don't think this works. When you wrap it in a `useRef` it will no longer trigger effects when the window becomes visible again.

src/hooks/useIsWindowVisibleRef.ts
```diff
@@ -0,0 +1,24 @@
+import { useEffect, useRef } from 'react'
+
+const VISIBILITY_STATE_SUPPORTED = 'visibilityState' in document
+
+// Check if the tab is active in the user browser
+export default function useIsWindowVisibleRef() {
+  const isBrowserTabActiveRef = useRef(true)
```
hachiojidev, line 7: This is not needed. A boolean is a primitive value and there is no point in wrapping it in a `useRef`

hachiojidev (changes requested):
Can you go into more detail on what the performance problem is. I have profiled the home page for example and do not see any degradation (it is worth noting that our performance is an ongoing issue).

src/contexts/RefreshContext.tsx
memoyil: Yes, in refresh context I just extract that to utils since we have more than one usage. It is used like that to not trigger useRefresh every time we have an event. Otherwise every tab switch will send requests.

memoyil:
> Can you go into more detail on what the performance problem is. I have profiled the home page for example and do not see any degradation (it is worth noting that our performance is an ongoing issue).

It has been seen in Pools and Farms page mainly

src/contexts/RefreshContext.tsx
hachiojidev: Okay, so in this case instead of creating another hook that copies the functionality just return a ref as well.

In `useIsWindowVisible` the return can look like this

```
return { isWindowVisible, isWindowVisibleRef }
```

memoyil: That should stay separated since hook that contains useState triggers rendering even if we only use ref object

hachiojidev: In this case I think it is okay as it is not an event that fires frequently.

memoyil: Then I am closing the PR for now

PR #1972, opened 2021-08-24
Feature/prediction history

hachiojidev:
Fetches history from node on initial load of the history pane.

### DIFF
--- a/src/state/predictions/index.ts
+++ b/src/state/predictions/index.ts
@@ -1,9 +1,19 @@
 import { createAsyncThunk, createSlice, PayloadAction } from '@reduxjs/toolkit'
+import { ethers } from 'ethers'
+import { formatUnits } from 'ethers/lib/utils'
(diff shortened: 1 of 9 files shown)


src/state/predictions/index.ts
```diff
@@ -156,6 +165,82 @@ export const fetchHistory = createAsyncThunk<{ account: string; bets: Bet[] }, {
...
+            totalBets: 0,
+            totalAmount: parseFloat(formatUnits(round.totalAmount)),
+            bullBets: 0,
+            bullAmount: parseFloat(formatUnits(round.bullAmount)),
+            bearBets: 0,
+            bearAmount: parseFloat(formatUnits(round.bearAmount)),
+            position: getRoundPosition(),
+          },
+        },
+      ]
```
hachiojidev, line 237: I made the decision to transform the node data into the shape of a `Bet` so that when data does come through from the subgraph we can simply merge it to update.

PR #1976, opened 2021-08-24
feat(lottery): Add 'See More' button & request additional lotteries when user has hit subgraph lottery limit

ChefHutch:
- Conditionally render a 'See More' button within Finished Rounds > Your History when the max number of lotteries returned by the subgraph is equal to the number of lotteries being displayed.

Screenshot 2021-08-24 at 15 50 30

- Currently will only appear for users who have entered 100+ rounds
- On click, an additional 100 are fetched. 'See More' will continue to display while the total number of lottery rounds shown matches the max number of lotteries requested.
- As the `numUserRoundsRequested` needs to be persisted outwith the component - I made a small hook to handle the logic and keep the `lottery/index.tsx` file clean.

### DIFF
--- a/src/state/lottery/index.ts
+++ b/src/state/lottery/index.ts
@@ -4,7 +4,7 @@ import { LotteryTicket, LotteryStatus } from 'config/constants/types'
 import { LotteryState, LotteryRoundGraphEntity, LotteryUserGraphEntity, LotteryResponse } from 'state/types'
 import { fetchLottery, fetchCurrentLotteryIdAndMaxBuy } from './helpers'
 import getLotteriesData from './getLotteriesData'
-import getUserLotteryData from './getUserLotteryData'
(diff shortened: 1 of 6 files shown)


src/state/lottery/index.ts
```diff
@@ -85,6 +85,14 @@ export const fetchUserLotteries = createAsyncThunk<
   return userLotteries
 })
 
+export const fetchAdditionalUserLotteries = createAsyncThunk<
```
hachiojidev, line 88: Minor thing, but we can just call this `fetchUserLotteries`

hachiojidev (approved):

src/state/lottery/index.ts
ChefHutch: There is already a `fetchUserLotteries` - the data fetch and the reducer behave a bit differently

PR #1977, opened 2021-08-24
Feature/Toasts in fe + optional txHash prop

ChefHutch:
- Move Toast code out of uikit
- Add optional txHash prop that links to bscscan transaction.


src/components/Toast/DescriptionWithTx.tsx
```diff
@@ -0,0 +1,29 @@
...
+const DescriptionWithTx: React.FC<DescriptionWithTxProps> = ({ description, txHash }) => {
+  const { chainId } = useActiveWeb3React()
+  const { t } = useTranslation()
+
+  return (
+    <>
+      {description && <Text as="p">{description}</Text>}
+      {txHash && (
+        <Link external href={getBscScanLink(txHash, 'transaction', chainId)}>
+          {t('View tx:')} {truncateWalletAddress(txHash, 8, 0)}
```
memoyil, line 22: Wouldn't be better to use 'View on BscScan' here?

src/views/Predictions/components/RoundCard/SetPositionCard.tsx
```diff
@@ -36,7 +36,7 @@ interface SetPositionCardProps {
   togglePosition: () => void
   epoch: number
   onBack: () => void
-  onSuccess: (decimalValue: string, hash: string) => Promise<void>
```
ChefHutch, line 39: This `decimalValue` wasn't used in the corresponding handleSuccess function (https://github.com/pancakeswap/pancake-frontend/pull/1977/files#diff-ef08349844ddbde8ad03f79f8d45dc7f46a0beafb9f9661033ac3594469298fdL100), so I've removed it.

src/config/localization/translations.json
```diff
@@ -1360,5 +1360,6 @@
   "You have no unclaimed v0.1 prizes.": "You have no unclaimed v0.1 prizes.",
   "Nothing to collect": "Nothing to collect",
   "Download .CSV": "Download .CSV",
-  "Download your v0.1 Prediction history below.": "Download your v0.1 Prediction history below."
+  "Download your v0.1 Prediction history below.": "Download your v0.1 Prediction history below.",
+  "View on BscScan:": "View on BscScan:"
```
memoyil, line 1364: There is already a text without colon, I think we can use that instead

PR #1979, opened 2021-08-25
feat(collectibles): Cache URI data

hachiojidev:


src/utils/collectibles.ts
```diff
@@ -31,17 +31,39 @@ export const getAddressByType = (type: NftType) => {
...
+
+    if (localUriData) {
+      const data = JSON.parse(localUriData)
+      return data
+    }
+
+    const uriDataResponse = await fetch(tokenUrl)
+
+    if (!uriDataResponse.ok) {
+      return null
```
ChefHutch, line 46: Worth console.erroring the two potential null return scenarios? 🤷

ChefHutch (approved):

PR #2002, opened 2021-08-27
ci(actions): Dependencies caching

ChefKai:


.github/workflows/test.yml
```diff
@@ -22,12 +20,24 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Setup Node
-        uses: actions/setup-node@v2.1.2
+        uses: actions/setup-node@v2
```
hachiojidev, line 23: Interesting. Is this due to a bug or will `v2` use the latest 2?

hachiojidev (approved):
Weird that GH does not do this automatically.

.github/workflows/test.yml
ChefKai: `v2` tag is re-directing to the latest v2 tag released

PR #2014, opened 2021-08-28
fix(predictions): History not loaded on collect winnings

hachiojidev:
- Temporarily disabled fetching history from the subgraph
- Fixed issue when opening claim modal before opening history

### DIFF
--- a/src/state/predictions/index.ts
+++ b/src/state/predictions/index.ts
@@ -165,81 +165,82 @@ export const fetchHistory = createAsyncThunk<{ account: string; bets: Bet[] }, {
   },
 )
 
-export const fetchNodeHistory = createAsyncThunk<{ account: string; bets: Bet[] }, string>(
-  'predictions/fetchNodeHistory',
-  async (account) => {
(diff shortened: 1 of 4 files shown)


src/state/predictions/index.ts
```diff
@@ -165,81 +165,82 @@ export const fetchHistory = createAsyncThunk<{ account: string; bets: Bet[] }, {
...
+>('predictions/fetchNodeHistory', async (account) => {
+  const userRounds = await fetchLatestUserRounds(account)
+
+  if (!userRounds) {
+    return { account, bets: [], claimableStatuses: {} }
+  }
+
+  const epochs = Object.keys(userRounds).map((epochStr) => Number(epochStr))
+  const roundData = await getRoundsData(epochs)
+  const claimableStatuses = await getClaimStatuses(account, epochs)
```
hachiojidev, line 180: This is the new line. Fetch claim status at the same time.

ChefKai (approved):

PR #2022, opened 2021-08-30
Split config tests

hachiojidev:
Splits up test config so the long-running token tests only run when file is changed.


.github/workflows/test.yml
```diff
@@ -12,8 +12,10 @@ jobs:
 
     strategy:
       fail-fast: true
+      matrix:
+        node: [14]
```
hachiojidev, line 16: @ChefKai I put this back because the GH linting was complaining.

.github/workflows/testConfig.yml
```diff
@@ -0,0 +1,47 @@
+name: Test Config
+
+on:
+  pull_request:
+    paths:
+      - 'src/__tests__/config/**'
```
hachiojidev, line 6: Here is the line that runs this workflow only when a file changes in the config folder.

RabbitDoge (approved):

.github/workflows/test.yml
```diff
@@ -12,8 +12,10 @@ jobs:
...
 
-    name: Did you fuck up again?
```
ChefHutch, line 16: 😢

PR #2032, opened 2021-08-30
fix(priceDecimal): Handle prices lower than 3 decimal

### LINKED ISSUE #2029: Show `<$0.001` when lock price and close price difference is less than 0.001

0xCorgi:
Fix #2029

[Live version](https://endurable-yam.surge.sh/prediction)

Prices that are below `0.001` are now displayed with `<0.001`.


ChefHutch (commented):
Noice! And look at those tests 😍 

Just a few nitpicky things but good stuff

src/views/Predictions/helpers.ts
```diff
@@ -1,15 +1,36 @@
...
+  let prefix = unitPrefix
+  let priceToFormat = value
+  const sign = value.isNegative() ? BigNumber.from(-1) : BigNumber.from(1)
+
+  if (value.abs().lt(minPriceDisplayed)) {
+    prefix = `<${unitPrefix}`
+    priceToFormat = minPriceDisplayed.mul(sign)
+  }
+
+  return `${prefix}${formatBigNumberToFixed(priceToFormat, 3, displayDecimals)}`
```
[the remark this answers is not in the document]
0xCorgi: You're right, I'll fix thix.

src/views/Predictions/helpers.ts
```diff
@@ -1,15 +1,36 @@
-import { ethers } from 'ethers'
+import { BigNumber, ethers } from 'ethers'
 import { NodeRound } from 'state/types'
 import { formatBigNumberToFixed } from 'utils/formatBalance'
 import getTimePeriods from 'utils/getTimePeriods'
 
+const MIN_PRICE_USD_DISPLAYED = BigNumber.from(100000)
+const MIN_PRICE_BNB_DISPLAYED = BigNumber.from('1000000000000000')
```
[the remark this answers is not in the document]
0xCorgi: Yes, the second call to `BigNumber.from` was breaking with a plain number.

src/views/Predictions/helpers.ts
```diff
@@ -1,15 +1,44 @@
...
+  decimals: number
+}
+
+const formatPriceDifference = ({
+  price = BigNumber.from(0),
+  minPriceDisplayed,
+  unitPrefix,
+  decimals,
+}: formatPriceDifferenceProps) => {
+  const value = price
```
hachiojidev, line 23: Any reason to instantiate this variable and not use `price` directly?

src/views/Predictions/helpers.ts
```diff
@@ -1,15 +1,44 @@
...
+  let prefix = unitPrefix
+  let priceToFormat = value
+  const sign = value.isNegative() ? BigNumber.from(-1) : BigNumber.from(1)
+
+  if (value.abs().lt(minPriceDisplayed)) {
+    prefix = `<${unitPrefix}`
+    priceToFormat = minPriceDisplayed.mul(sign)
+  }
```
hachiojidev, line 31: Totally valid code but try and avoid using `let` and return ``<${unitPrefix}`` early.

src/views/Predictions/helpers.test.ts
```diff
@@ -0,0 +1,61 @@
...
+    ${'-100000000000000'}       | ${'<-0.001'}
+    ${'-10000000000000'}        | ${'<-0.001'}
+    ${'-1000000000000'}         | ${'<-0.001'}
+    ${'-100000000000'}          | ${'<-0.001'}
+  `(
+    'should format $priceDifference to $expectedPriceDifferenceFormatted',
+    ({ priceDifference, expectedPriceDifferenceFormatted }) =>
+      expect(formatBnbv2(BigNumber.from(priceDifference))).toEqual(expectedPriceDifferenceFormatted),
+  )
+})
```
hachiojidev, line 61: I wholeheartedly agree to put tests next to the file. We just have to convince @RabbitDoge who seems to be the straggler. So for now can you put it in the `src/__test__/views/Predictions/helpers.test.ts`

hachiojidev (changes requested):
Nice work dude.

src/views/Predictions/helpers.ts
0xCorgi: Good point, it was used for doing a default assignment before. Since it's done in the params now I will clean this.

src/views/Predictions/helpers.ts
0xCorgi: I also like to avoid `let` as much as possible, just wanted to mutualize the condition here.
I will use an early return instead.

hachiojidev (approved):

PR #2038, opened 2021-08-31
refactor: Replace BigNumber.js

hachiojidev:


src/views/Pools/hooks/useUnstakePool.ts
```diff
@@ -1,30 +1,31 @@
...
+import { parseUnits } from 'ethers/lib/utils'
 import { useAppDispatch } from 'state'
 import { updateUserStakedBalance, updateUserBalance, updateUserPendingReward } from 'state/actions'
 import { unstakeFarm } from 'utils/calls'
 import { useMasterchef, useSousChef } from 'hooks/useContract'
-import { BIG_TEN } from 'utils/bigNumber'
 import getGasPrice from 'utils/getGasPrice'
 
-const sousUnstake = async (sousChefContract, amount, decimals) => {
+const sousUnstake = async (sousChefContract: any, amount: string, decimals: number) => {
```
0xCorgi, line 10: We can't use the `Contract` type from `etherjs` here ?

0xCorgi (commented):
Looks good !
Just one question.

src/views/Pools/hooks/useUnstakePool.ts
hachiojidev: Most likely we can but I didn't want to go down the rabbit hole. I also didn't want it not to be types :)

0xCorgi (approved):

ChefHutch (approved):

RabbitDoge (approved):

PR #2040, opened 2021-08-31
refactor: Deprecate usePriceBnbBusd

hachiojidev:
Deprecate `usePriceBnbBusd` in favor of `useBUSDPrice`

#### Differences

`useBUSDPrice` returns `Price`. `Price` has add, subtract, multiply helpers but they require you to pass in another `Price` object. However, it is easy enough to use `toSignificant()` and `toFixed()` to convert in order to multiply by it by another number.


src/views/Home/components/WinSection/PredictionCardContent.tsx
```diff
@@ -42,12 +44,6 @@ const PredictionCardContent = () => {
     }
   }, [slowRefresh, loadData])
 
-  useEffect(() => {
-    if (bnbBusdPrice.gt(0) && bnbWon > 0) {
-      setBnbWonInUsd(bnbBusdPrice.times(bnbWon).toNumber())
-    }
-  }, [bnbBusdPrice, bnbWon])
-
```
hachiojidev, line 50: This was unnecessary and can be calculated on render.

0xCorgi (approved):
Looks good !

RabbitDoge (approved):

PR #2053, opened 2021-09-01
Refactor/voting remove bignumber

hachiojidev:


src/views/Voting/helpers.ts
```diff
@@ -106,7 +104,7 @@ export const calculateVoteResults = (votes: Vote[]): { [key: string]: Vote[] } =
 
 export const getTotalFromVotes = (votes: Vote[]) => {
   return votes.reduce((accum, vote) => {
-    const power = new BigNumber(vote.metadata?.votingPower)
-    return accum.plus(power)
-  }, BIG_ZERO)
+    const power = parseFloat(vote.metadata?.votingPower)
+    return accum + power
+  }, 0)
```
ChefHutch, line 109: Is this broadly going to be the approach? Moving from `BigNumber.js` classes to floats? Or is it case-by-case? 
I thought we were going to try and make use of ethers.BigNumber wherever possible?

Would this generally mean the flow is:
> Receive ethers.BigNumber from contract response > Serialize for storage in state > Convert to float on retrieval?

And if working directly with the contract response values
> Receive ethers.BigNumber from contract response > Convert to float for use in functions etc.?

hachiojidev: This is a one-off. The value comes back as a string and it does not need to interact with any other numbers.

RabbitDoge (approved):

PR #2054, opened 2021-09-01
refactor(modal-input): Deprecate big number

hachiojidev:


src/components/Modal/ModalInput.tsx
```diff
@@ -76,11 +77,9 @@ const ModalInput: React.FC<ModalInputProps> = ({
     if (isBalanceZero) {
       return '0'
     }
-    const balanceBigNumber = new BigNumber(balance)
-    if (balanceBigNumber.gt(0) && balanceBigNumber.lt(0.0001)) {
```
hachiojidev, line 80: I don't think this is needed. `formatBigNumber` will display the decimals available.

memoyil: Not it will show the full balance instead of 3 decimals for the lp's greater than 0.0001

image

hachiojidev: I think this is preferable. Why hide the balance if we have the space?

ChefHutch: @memoyil did you make the change because Chungus/Bugs asked for it? If so we should probably leave the rounding in, but if not - yeet!

memoyil: As far as I remember, at that time there was a bug that if LP price is too high (I think in ALICE farm) therefore user has small portion of it that means it always shows 0 instead of full balance. I added that condition, but if we always show the full balance we can remove that check

Chef-Cheems (approved):

Chef-Cheems (approved):

src/components/Modal/ModalInput.tsx
ChefHutch: Yea it looks like we were originally ([here](https://github.com/pancakeswap/pancake-frontend/pull/886/files#diff-56f0426bac583711c28892f9cbd9f476a4faafcc1d3298d2ea01eac5afd6e3b1L71)) rounding all balances to 4 decimals, which would show 0 for users with balances smaller than that. 

Agree we can just show the full value ✅

ChefHutch (approved):

RabbitDoge (approved):

PR #2064, opened 2021-09-02
feat(predictions): Card state for house wins

hachiojidev:
image

### DIFF
--- a/src/views/Predictions/components/RoundResult/styles.tsx
+++ b/src/views/Predictions/components/RoundResult/styles.tsx
@@ -1,10 +1,12 @@
 import React from 'react'
 import styled, { DefaultTheme } from 'styled-components'
-import { Box, Flex, FlexProps, Text } from '@pancakeswap/uikit'
+import { ethers } from 'ethers'
+import { Box, Flex, FlexProps, Skeleton, Text } from '@pancakeswap/uikit'
 import { useTranslation } from 'contexts/Localization'
 import { BetPosition, NodeRound, Round } from 'state/types'
-import { formatUsdv2, formatBnbv2 } from '../../helpers'
+import { formatUsdv2, formatBnbv2, getRoundPosition, getPriceDifference } from '../../helpers'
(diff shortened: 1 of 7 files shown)


src/views/Predictions/helpers.ts
```diff
@@ -84,3 +84,15 @@ export const getPriceDifference = (price: ethers.BigNumber, lockPrice: ethers.Bi
...
+export const getRoundPosition = (lockPrice: ethers.BigNumber, closePrice: ethers.BigNumber) => {
+  if (!closePrice) {
+    return null
+  }
+
+  if (closePrice.eq(lockPrice)) {
+    return BetPosition.HOUSE
+  }
+
+  return closePrice.gt(lockPrice) ? BetPosition.HOUSE : BetPosition.HOUSE
```
ChefKai, line 97: ?

hachiojidev: Nice catch. Was using that for testing!

RabbitDoge (approved):

src/views/Predictions/components/RoundResult/styles.tsx
```diff
@@ -140,6 +143,41 @@ export const RoundResultBox: React.FC<RoundResultBoxProps> = ({
...
+export const RoundPrice: React.FC<RoundPriceProps> = ({ lockPrice, closePrice }) => {
+  const betPosition = getRoundPosition(lockPrice, closePrice)
+  const priceDifference = getPriceDifference(closePrice, lockPrice)
+
+  const getTextColor = () => {
+    switch (betPosition) {
+      case BetPosition.BULL:
+        return 'success'
+      case BetPosition.BEAR:
+        return 'success'
```
Chef-Cheems, line 160: I think this one should be `failure`

Chef-Cheems (approved):

src/views/Predictions/components/RoundResult/styles.tsx
hachiojidev: Another great catch.

PR #2081, opened 2021-09-03
feat: Add namespace to redux local storage

### LINKED ISSUE #2072: [Feature Request] Handle persisted redux storage better if new keys are added

### Is your feature request related to a problem you are facing?

Whenever we add new key to redux store that is persisted via local storage (e.g. `user`) it gets initialized imporperly. I.e. if you add key `lol` to `user` state and initialize the value to `kek` - when a user with user state saved in localstorage comes to the page the `lol` value in redux will be `undefined` because the object loaded from local storage didn't contain it.

That forces us to put extra checks for undefined values when adding new keys.

### Describe the solution you'd like

Maybe namespace feature of [redux-localstorage-simple](https://www.npmjs.com/package/redux-localstorage-simple) can be used to account for that. Or some other way to update localstorage when there is new key in the state.
It must somehow reference either package version or git commit hash.

### How does this feature benefit PancakeSwap users and CAKE holders?

No crashes if new key is introduced

### Anything else?

_No response_

hachiojidev:
Closes #2072 

Adds a namespace to all keys in local storage. This will enable us to "clear" the cache if we push a breaking change.

### DIFF
--- a/.env
+++ b/.env
@@ -1,1 +1,2 @@
+REACT_APP_VERSION=$npm_package_version
 REACT_APP_API_PROFILE = https://profile.pancakeswap.com
--- a/src/state/index.ts
+++ b/src/state/index.ts
@@ -1,6 +1,7 @@
 import { configureStore, getDefaultMiddleware } from '@reduxjs/toolkit'
 import { save, load } from 'redux-localstorage-simple'
 import { useDispatch } from 'react-redux'
+import { LS_PREFIX } from 'config'
 import farmsReducer from './farms'
 import poolsReducer from './pools'
 import predictionsReducer from './predictions'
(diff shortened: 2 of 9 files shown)


.env
```diff
@@ -1 +1,2 @@
+REACT_APP_VERSION=$npm_package_version
```
hachiojidev, line 1: https://create-react-app.dev/docs/adding-custom-environment-variables/#expanding-environment-variables-in-env

src/state/index.ts
```diff
@@ -47,8 +49,11 @@ const store = configureStore({
     multicall,
     lists,
   },
-  middleware: [...getDefaultMiddleware({ thunk: true }), save({ states: PERSISTED_KEYS })],
-  preloadedState: load({ states: PERSISTED_KEYS }),
+  middleware: [
+    ...getDefaultMiddleware({ thunk: true }),
+    save({ states: PERSISTED_KEYS, namespace: `${LS_PREFIX}-${appVersion}` }),
+  ],
+  preloadedState: load({ states: PERSISTED_KEYS, namespace: `${LS_PREFIX}-${appVersion}` }),
```
Chef-Cheems, line 56: So this will erase user's previous settings? 🤔 
I was thinking of copying available settings to new namespace or something. If it nukes everything then its gonna be annoying for users to find out all their exchange settings, watchlist, etc are gone

src/utils/migrateLocalStorage.ts
```diff
@@ -0,0 +1,39 @@
...
+            localStorage.setItem(newKey, value)
+          }
+        }
+      }
+      localStorage.setItem(hasMigratedKey, JSON.stringify(true))
+    }
+  } catch {
+    console.error('Error migrating local storage')
+  }
+}
```
hachiojidev, line 37: This function takes a users' existing local storage and adds the proper namespace.

```
redux_localstorage_simple_lists
```
Becomes
```
pancakeswap-0.1.0-lists
```

It does nothing with the existing value in storage just in case something goes wrong.

PR #2095, opened 2021-09-06
fix: Lottery bunny resize at footer in safari mobile

memoyil:
Somehow safari mobile doesn't take the height proportions correctly that we have to manually define it

To review:

To reproduce:

1. Go to lottery with iOS Safari 
2. Check bottom
3. See bunny logo height is not correct and image looks stretched

### DIFF
--- a/src/views/Lottery/components/HowToPlay.tsx
+++ b/src/views/Lottery/components/HowToPlay.tsx
@@ -1,6 +1,6 @@
 import React from 'react'
 import styled from 'styled-components'
-import { Box, Flex, Text, Heading, useMatchBreakpoints, Link } from '@pancakeswap/uikit'
+import { Box, Flex, Text, Heading, useMatchBreakpoints, Link, Image } from '@pancakeswap/uikit'
 import { useTranslation } from 'contexts/Localization'
 import useTheme from 'hooks/useTheme'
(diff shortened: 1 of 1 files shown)


src/views/Lottery/components/HowToPlay.tsx
```diff
@@ -338,8 +338,8 @@ const HowToPlay: React.FC = () => {
       </GappedFlex>
       <Divider />
       <Flex justifyContent="center" alignItems="center" flexDirection={['column', 'column', 'row']}>
-        <Flex maxWidth="240px" mr="8px" mb="16px">
-          <img src="/images/lottery/tombola.png" alt="tombola bunny" />
+        <Flex mr="8px" mb="16px">
+          <img style={{ width: '240px', height: '172px' }} src="/images/lottery/tombola.png" alt="tombola bunny" />
```
hachiojidev, line 342: Can you use the `Image` component instead?

hachiojidev (changes requested):

src/views/Lottery/components/HowToPlay.tsx
memoyil: Done

hachiojidev (approved):

PR #2106, opened 2021-09-06
feat: nft Collections & Profile header

ChefHutch:
### Example implementation, showing composability: `/nft/market/Profile/ProfileHeader.tsx`

Screenshot 2021-09-06 at 18 13 27


ChefHutch:
Especially for @hachiojidev 's eyes as he would make use of this on the collections page.

src/views/nft/market/Profile/EditProfileAvatar.tsx
```diff
@@ -0,0 +1,43 @@
...
+  &:hover {
+    cursor: pointer;
+
+    ${EditOverlay} {
+      opacity: 1;
+    }
+  }
+`
+
+const EditProfileAvatar: React.FC<{ src: string; alt?: string }> = ({ src, alt }) => {
```
ChefHutch, line 32: Practically a duplicate of the existing component used on the Profile page. Will kill the legacy one when we kill the legacy page.

hachiojidev (approved):

PR #2118, opened 2021-09-07
chore: Yarn prettier changes and change format script directory level

memoyil:
To review:

https://deploy-preview-2118--pancakeswap-dev.netlify.app/


ChefKai:
Surprised at first that they were not caught up by Prettier, but we need to update the `format:check` and `format:write` scripts in order to check at root level if there is any issue

memoyil:
> Surprised at first that they were not caught up by Prettier, but we need to update the `format:check` and `format:write` scripts in order to check at root level if there is any issue

Done

ChefKai (approved):

hachiojidev (approved):

PR #2140, opened 2021-09-08
chore: Checksum addresses

ChefKai:
Checksum everythiiiing. Tests will be clearer to analyze for future failures too :)

Not sure if we want to fix the last warning about tests with are related to Redux reducer.. That will be another PR anyway.


hachiojidev (approved):
Nice. We should consider adding the checksum test we have in token list now.

PR #2141, opened 2021-09-08
fix: Add liquidity shows undefined in address bar when selecting token

memoyil:
To review:

To reproduce:

1. Go to add liquidity
2. Select token from first input panel
3. See undefined in browser address bar


ChefKai (commented):
Have no idea what I am looking at but I was able to confirm that this issue exists and this is a good catch as always!

hachiojidev (approved):

PR #2142, opened 2021-09-08
fix(info): Hardcoded exchange links

ChefKai:


Chef-Cheems (approved):

src/views/Info/Pools/PoolPage.tsx
```diff
@@ -149,16 +149,12 @@ const PoolPage: React.FC<RouteComponentProps<{ address: string }>> = ({
                 </Link>
               </Flex>
               <Flex>
-                <a
-                  href={`https://exchange.pancakeswap.finance/#/add/${poolData.token0.address}/${poolData.token1.address}`}
-                >
+                <a href={`/add/${poolData.token0.address}/${poolData.token1.address}`}>
```
hachiojidev, line 152: So now that we have the exchange in the same app we should use the `Link` component from `react-router-dom` so the page does not reload.

hachiojidev (changes requested):

src/views/Info/Pools/PoolPage.tsx
Chef-Cheems: @hachiojidev replaced it with Link - 129d5847304cf8ecead47ebfca5eaaaa1e08d219

hachiojidev (approved):

PR #2149, opened 2021-09-09
fix: (prediction) Buttons moving when larger digits in timer on menu in prediction mobile

memoyil:
To review:

https://deploy-preview-2149--pancakeswap-dev.netlify.app/

To reproduce:

1. Go to prediction in mobile (iphone 8 size)
2. See button are moving in top menu above cards when timer is counting down

### DIFF
--- a/src/views/Predictions/components/Menu.tsx
+++ b/src/views/Predictions/components/Menu.tsx
@@ -20,7 +20,7 @@ const HelpButtonWrapper = styled.div`
   order: 1;
   margin: 0 2px 0 8px;
 
-  ${({ theme }) => theme.mediaQueries.lg} {
+  ${({ theme }) => theme.mediaQueries.sm} {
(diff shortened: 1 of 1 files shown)


src/views/Predictions/components/Menu.tsx
```diff
@@ -20,22 +20,19 @@ const HelpButtonWrapper = styled.div`
...
     margin: 0 0 0 8px;
   }
 `
 
 const TimerLabelWrapper = styled.div`
   order: 3;
-  max-width: 100px;
+  width: 100px;
 
   ${({ theme }) => theme.mediaQueries.sm} {
```
memoyil, line 33: SM width is enough to put timer label to left most place, therefore we can set width to auto for different languages especially for closing text translation

hachiojidev (approved):

PR #2160, opened 2021-09-10
fix: Balance loading indicator missing when user selects token in swap input

memoyil:
To review:

https://deploy-preview-2160--pancakeswap-dev.netlify.app/

To reproduce: 

1. Go to swap with your account
2. Select token
3. See balance inside the input panel shows "-" a few time without loading indicator

### DIFF
--- a/src/components/CurrencyInputPanel/index.tsx
+++ b/src/components/CurrencyInputPanel/index.tsx
@@ -96,8 +96,8 @@ export default function CurrencyInputPanel({
               <Text fontSize="14px">{translatedLabel}</Text>
               {account && (
                 <Text onClick={onMax} fontSize="14px" style={{ display: 'inline', cursor: 'pointer' }}>
-                  {!hideBalance && !!currency && selectedCurrencyBalance
-                    ? t('Balance: %amount%', { amount: selectedCurrencyBalance?.toSignificant(6) ?? '' })
+                  {!hideBalance && !!currency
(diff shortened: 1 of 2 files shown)


src/config/localization/translations.json
```diff
@@ -1124,7 +1124,6 @@
   "Your transaction will revert if it is pending for more than this long.": "Your transaction will revert if it is pending for more than this long.",
   "minutes": "minutes",
   "Token Amount": "Token Amount",
-  "Balance: %amount%": "Balance: %amount%",
```
memoyil, line 1127: It is redundant

src/components/CurrencyInputPanel/index.tsx
```diff
@@ -96,8 +96,8 @@ export default function CurrencyInputPanel({
               <Text fontSize="14px">{translatedLabel}</Text>
               {account && (
                 <Text onClick={onMax} fontSize="14px" style={{ display: 'inline', cursor: 'pointer' }}>
-                  {!hideBalance && !!currency && selectedCurrencyBalance
-                    ? t('Balance: %amount%', { amount: selectedCurrencyBalance?.toSignificant(6) ?? '' })
+                  {!hideBalance && !!currency
+                    ? t('Balance: %balance%', { balance: selectedCurrencyBalance?.toSignificant(6) ?? t('Loading') })
```
memoyil, line 100: I tried to put skeleton here but it is hardly noticeable since it has kind of same colors with input panel

hachiojidev (approved):

PR #2161, opened 2021-09-10
refactor: Remove bignumber in constants

memoyil:
To review:

https://deploy-preview-2161--pancakeswap-dev.netlify.app/

Check apr's on farms and pools work properly.

### DIFF
--- a/src/utils/apr.ts
+++ b/src/utils/apr.ts
@@ -27,7 +27,8 @@ export const getPoolApr = (
  * @param poolWeight allocationPoint / totalAllocationPoint
  * @param cakePriceUsd Cake price in USD
  * @param poolLiquidityUsd Total pool liquidity in USD
- * @returns
+ * @param farmAddress Farm Address
+ * @returns Farm Apr
  */
 export const getFarmApr = (
   poolWeight: BigNumber,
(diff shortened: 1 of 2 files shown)


src/utils/apr.ts
```diff
@@ -27,15 +27,16 @@ export const getPoolApr = (
...
+ * @returns Farm Apr
  */
 export const getFarmApr = (
   poolWeight: BigNumber,
   cakePriceUsd: BigNumber,
   poolLiquidityUsd: BigNumber,
   farmAddress: string,
 ): { cakeRewardsApr: number; lpRewardsApr: number } => {
-  const yearlyCakeRewardAllocation = CAKE_PER_YEAR.times(poolWeight)
+  const yearlyCakeRewardAllocation = poolWeight ? poolWeight.times(CAKE_PER_YEAR) : new BigNumber(NaN)
```
hachiojidev, line 39: Why `NaN`? Wouldn't it be better to have it be 0?

memoyil: If poolWeight is null and others not, (although it is not possible because in https://github.com/pancakeswap/pancake-frontend/blob/363b7400c168097585df14ed35bed97de96e60b2/src/state/farms/fetchPublicFarmData.ts#L90 we put zero if we don't have totalAllocPoint) , 

if we assign 0 to yearlyCakeRewardAllocation, cakeRewardsApr will be 0 instead of null that is why I choose NaN there.

hachiojidev (approved):

PR #2167, opened 2021-09-13
refactor(nft): Move routes inside index

hachiojidev:
- Adds a loading state so we can fetch the collections before rendering. This will ensure that it is available on all sub-routes
- Moved routes into the `views/Nft/market`

### DIFF
--- a/src/state/nftMarket/reducer.ts
+++ b/src/state/nftMarket/reducer.ts
@@ -1,9 +1,9 @@
 import { createSlice, createAsyncThunk } from '@reduxjs/toolkit'
 import { getNftsMetadata, getNftsMarketData, getCollectionsApi, getCollectionsSg } from './helpers'
-import { State, Collection, NFT, NFTMarketLoadingState } from './types'
+import { State, Collection, NFT, NFTMarketInitializationState } from './types'
 
 const initialState: State = {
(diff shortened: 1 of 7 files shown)


src/state/nftMarket/reducer.ts
```diff
@@ -1,9 +1,9 @@
 import { createSlice, createAsyncThunk } from '@reduxjs/toolkit'
-import { getCollectionsApi, getCollectionsSg, getNftsMetadata, getNftsMarketData } from './helpers'
-import { State, Collection, NFT } from './types'
+import { getNftsMetadata, getNftsMarketData, getCollectionsApi, getCollectionsSg } from './helpers'
+import { State, Collection, NFT, NFTMarketLoadingState } from './types'
 
 const initialState: State = {
-  isInitializing: true,
+  loadingState: NFTMarketLoadingState.IDLE,
```
hachiojidev, line 6: This pattern is more flexible and is used in other states.

RabbitDoge: We probably want to make a difference between the first loading and other loadings

RabbitDoge (approved):
I'm not sure about the changes on the loading stuff, but otherwise it's ok

PR #2179, opened 2021-09-13
chore: Bring back custom node

Chef-Cheems:
`+` fix cypress test if prediction is paused


Chef-Chungus (approved):

ChefHutch (approved):

ChefHutch:
Integration tests for prediction page ded because markets are paused. We probably need to remove the dependency on that @Chef-Cheems ?

PR #2181, opened 2021-09-13
Feature/nft profile activity

ChefHutch:
- SG fetch for user `buyTradeHistory`, `sellTradeHistory`, `askOrderHistory`
- Store in `user` nftMarket state
- Merge and sort data as part of `useUserActivity` hook
- Render to `Activity` component

### Data for this:

Screenshot 2021-09-15 at 09 14 03

### DIFF
--- a/src/state/nftMarket/queries.ts
+++ b/src/state/nftMarket/queries.ts
@@ -1,4 +1,4 @@
-export const getBaseNftsFields = () => `
+export const getBaseNftFields = () => `
   tokenId
   metadataUrl
   currentAskPrice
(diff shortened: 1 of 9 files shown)


src/state/nftMarket/queries.ts
```diff
@@ -7,4 +7,22 @@ export const getBaseNftsFields = () => `
   tradeVolumeBNB
   totalTrades
   isTradable
+  collection {
+    id
```
hachiojidev, line 11: I don't think we should include sub queries in the base fields.

src/state/nftMarket/queries.ts
```diff
@@ -7,4 +7,22 @@ export const getBaseNftsFields = () => `
...
+  timestamp
+  askPrice
+  netPrice
+  withBNB
+  buyer {
+    id
+  }
+  seller {
+    id
+  }
```
hachiojidev, line 27: Same here.

hachiojidev (approved):

src/state/nftMarket/queries.ts
ChefHutch: Ah I hear ya - this one in particular is pretty key for every NFT query though. We can refactor later?

PR #2210, opened 2021-09-16
refactor: Address search

hachiojidev:
- Moved address input search to global to be used in NFT market

### DIFF
--- a/src/state/predictions/index.ts
+++ b/src/state/predictions/index.ts
@@ -66,6 +66,7 @@ const initialState: PredictionsState = {
   ledgers: {},
   claimableStatuses: {},
   leaderboard: {
+    selectedAddress: null,
     loadingState: LeaderboardLoadingState.INITIAL,
     filters: {
       address: null,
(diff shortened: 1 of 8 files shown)


src/state/predictions/index.ts
```diff
@@ -66,6 +66,7 @@ const initialState: PredictionsState = {
   ledgers: {},
   claimableStatuses: {},
   leaderboard: {
+    selectedAddress: null,
```
hachiojidev, line 69: This was necessary because it was working on the Leader board only by accident. It spawns from our issue with the modals and props.

src/views/Predictions/Leaderboard/components/WalletStatsModal.tsx
```diff
@@ -43,8 +47,10 @@ const ExternalLink = styled(LinkExternal)`
 const WalletStatsModal: React.FC<WalletStatsModalProps> = ({ account, onDismiss, onBeforeDismiss }) => {
   const { t } = useTranslation()
   const { theme } = useTheme()
-  const result = useGetOrFetchLeaderboardAddressResult(account)
-  const profileAvatar = useGetProfileAvatar(account)
+  const selectedAddress = useGetSelectedAddress()
+  const address = account || selectedAddress
```
hachiojidev, line 51: This is where we accept the passed in account and default to the selected on in state.

ChefHutch (approved):
This is neat nice one.

Might be next week before I sort the page rendering different accounts btw 👍

PR #2213, opened 2021-09-16
chore: Update collection header to use values from API

hachiojidev:


src/state/nftMarket/types.ts
```diff
@@ -115,6 +115,13 @@ export interface Collection {
   owner: string
   totalSupply: BigNumberish
   verified: boolean
+  avatar: string
+  banner: {
+    large: string
+    small: string
+  }
+  createdAt: string
```
RabbitDoge, line 123: line 123 and 124 need to be removed to avoid confusion with subgraph data

hachiojidev: Updated.

ChefHutch (approved):

RabbitDoge (approved):

PR #2223, opened 2021-09-16
fix: Remove duplicated constants

ChefKai:


src/config/constants/info.ts
```diff
@@ -7,9 +7,6 @@ export const LP_HOLDERS_FEE = 0.0017
 export const TREASURY_FEE = 0.0003
 export const BUYBACK_FEE = 0.0005
 
-export const BASE_URL = 'https://pancakeswap.finance'
-export const BASE_BSC_SCAN_URL = 'https://bscscan.com'
```
ChefKai, line 11: We have `getBscScanLink` helper for that

src/config/constants/info.ts
```diff
@@ -7,9 +7,6 @@ export const LP_HOLDERS_FEE = 0.0017
 export const TREASURY_FEE = 0.0003
 export const BUYBACK_FEE = 0.0005
 
-export const BASE_URL = 'https://pancakeswap.finance'
```
ChefKai, line 10: Already present in `src/config/index`

hachiojidev (approved):

PR #2227, opened 2021-09-17
fix: (pools) Help button can be clicked beyond its size

memoyil:
To review:

https://deploy-preview-2227--pancakeswap-dev.netlify.app/

To reproduce:

1. Go to pools
2. See Help button in the header can be clicked beyond its size to the left


src/views/Pools/components/HelpButton.tsx
```diff
@@ -3,37 +3,35 @@ import styled from 'styled-components'
 import { Text, Button, HelpIcon, Link } from '@pancakeswap/uikit'
 import { useTranslation } from 'contexts/Localization'
 
-const ButtonText = styled(Text)`
```
memoyil, line 6: This is not needed since the smallest size is xs

hachiojidev (approved):

PR #2228, opened 2021-09-17
fix: Fix hidden profile header

chenxsan:
Currently part of the profile header is hidden under the navigation bar:

And here's where that 57 number comes:

Here's how It looks like after:

### DIFF
--- a/src/views/Profile/components/HeaderWrapper.tsx
+++ b/src/views/Profile/components/HeaderWrapper.tsx
@@ -4,6 +4,7 @@ const HeaderWrapper = styled.div`
   border-bottom: 2px solid ${({ theme }) => theme.colors.textSubtle};
   margin-bottom: 24px;
   padding-bottom: 24px;
+  margin-top: 57px;
 `
 
 export default HeaderWrapper


src/views/Profile/components/HeaderWrapper.tsx
```diff
@@ -4,6 +4,7 @@ const HeaderWrapper = styled.div`
   border-bottom: 2px solid ${({ theme }) => theme.colors.textSubtle};
   margin-bottom: 24px;
   padding-bottom: 24px;
+  margin-top: 57px;
```
ahmadSS85, line 7: To
0x3EA21b58B8F73bB70A8df9Fb4528246668BAE46E

ahmadSS85 (approved):
0x3EA21b58B8F73bB70A8df9Fb4528246668BAE46E

chenxsan:
@hachiojidev Hi there, just curious, was the issue already fixed somewhere? Or something's wrong with this pull request?

Just in case you want to know how to reproduce the issue, here're the steps:

1. open the home page either on you mobile phone or any desktop browsers
2. connect your wallet
3. access the profile page - now you should see the truncated header hidden from the navbar at the top.

hachiojidev:
@chenxsan Apologies. I saw the wallet address that was commented and I assumed it was the person who opened the PR so I closed it in anger :)

While this solution will fix it I don't think it is sustainable one having an arbitrary margin applied. I will re-open it and let @0xCorgi  decide if it is an acceptable solution.

chenxsan:
@hachiojidev Thanks for reopening, I'll file an issue first next time.

0xCorgi:
@chenxsan  @hachiojidev 

This problem occurs on all pages without a SubNav, it's because currently, it's the SubNav that is responsible for pushing the content of the page based on the menu height. It's done here in the UIKit: https://github.com/pancakeswap/pancake-toolkit/blob/master/packages/pancake-uikit/src/widgets/Menu/Menu.tsx#L132

Your fix would work obviously @chenxsan , but a better solution would be to apply this margin elsewhere than on the SubNav which is not always displayed.

chenxsan:
Glad it helps.

PR #2241, opened 2021-09-18
Fix: Display the correct USD amount for the best round in PnL tab of the Prediction's History

manelrovira:
In the PnL Tab of the history of predictions the USD value of the best round is displayed incorrectly. Instead of showing the USD value of the gains is showing the USD value of the multiplier. This PR fixes the problem.

Captura de Pantalla 2021-09-18 a les 11 01 45


hachiojidev:
@manelrovira Thank you for submitting this fix. Just update your commit message based on [our guidelines](https://docs.pancakeswap.finance/code/contributing#committing) and we're good to go.

manelrovira:
> @manelrovira Thank you for submitting this fix. Just update your commit message based on [our guidelines](https://docs.pancakeswap.finance/code/contributing#committing) and we're good to go.

Done.

PR #2245, opened 2021-09-19
fix: Recents transactions can be seen from another account on the same device

memoyil:
To review:

https://deploy-preview-2245--pancakeswap-dev.netlify.app/

To reproduce:

1. Do some transactions
2. Logout
3. Login with another account
4. Go to recent transactions
5. See transactions from previous account still visible

### DIFF
--- a/src/hooks/useAuth.ts
+++ b/src/hooks/useAuth.ts
@@ -1,5 +1,5 @@
 import { useCallback } from 'react'
-import { useWeb3React, UnsupportedChainIdError } from '@web3-react/core'
+import { UnsupportedChainIdError, useWeb3React } from '@web3-react/core'
 import { NoBscProviderError } from '@binance-chain/bsc-connector'
 import {
   NoEthereumProviderError,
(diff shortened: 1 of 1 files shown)


src/hooks/useAuth.ts
```diff
@@ -16,11 +16,13 @@ import useToast from 'hooks/useToast'
...
 import { useAppDispatch } from 'state'
 import { useTranslation } from 'contexts/Localization'
+import { clearAllTransactions } from '../state/transactions/actions'
+import useActiveWeb3React from './useActiveWeb3React'
 
 const useAuth = () => {
   const { t } = useTranslation()
   const dispatch = useAppDispatch()
-  const { activate, deactivate } = useWeb3React()
+  const { chainId, activate, deactivate } = useActiveWeb3React()
```
memoyil, line 25: @hachiojidev Is it ok to use useActiveWeb3React here?

src/hooks/useAuth.ts
```diff
@@ -16,11 +16,12 @@ import useToast from 'hooks/useToast'
 import { profileClear } from 'state/profile'
 import { useAppDispatch } from 'state'
 import { useTranslation } from 'contexts/Localization'
+import { clearAllTransactions } from 'state/transactions/actions'
 
 const useAuth = () => {
   const { t } = useTranslation()
   const dispatch = useAppDispatch()
-  const { activate, deactivate } = useWeb3React()
+  const { chainId, activate, deactivate } = useWeb3React()
```
memoyil, line 24: @hachiojidev Should we use useActiveWeb3React here or keep it as it is?

hachiojidev (approved):

PR #2253, opened 2021-09-20
feat(nft-collections): Trait page setup

hachiojidev:


src/state/nftMarket/hooks.ts
```diff
@@ -14,17 +13,17 @@ export const useFetchCollections = () => {
   const dispatch = useAppDispatch()
   useEffect(() => {
     dispatch(fetchCollections())
-    dispatch(fetchNftsFromCollections('0x60935f36e4631f73f0f407e68642144e07ac7f5e'))
+    dispatch(fetchNftsFromCollections(pancakeBunniesAddress))
```
ChefHutch, line 16: Should this maybe be populated via the collections object in Redux (that comes from the API) rather than a constant?

Chef-Cheems (approved):

src/views/Nft/market/Profile/components/SubMenu.tsx
```diff
@@ -19,15 +10,15 @@ const SubMenuComponent: React.FC = () => {
   const ItemsConfig = [
     {
       label: t('Items'),
-      href: `${nftsBaseUrl}/profile`,
+      href: '/nft/market/profile',
```
RabbitDoge, line 13: Any reason why you don't use `nftsBaseUrl` ?  Most likely the url will change

src/state/nftMarket/hooks.ts
ChefHutch: Hmmn I realise it's maybe overkill when we only have one collection and we're trying to move quickly - but I think modifying the `fetchNftsFromCollections` to accept multiple addresses would allow us to do all of this dynamically?

src/views/Nft/market/index.tsx
```diff
@@ -30,13 +28,10 @@ const Market = () => {
         <BuySellDemo />
       </Route>
       <Route exact path={`${nftsBaseUrl}/collections`}>
-        <CollectibleOverview />
+        <Redirect to={`${nftsBaseUrl}/collections/${pancakeBunniesAddress}`} />
```
RabbitDoge, line 31: Maybe you can remove this , I did this page #2279

src/views/Nft/market/Collection/IndividualNFTPage/MoreFromThisCollection.tsx
```diff
@@ -38,7 +39,7 @@ const StyledSwiper = styled.div`
 const MoreFromThisCollection: React.FC = () => {
   const { t } = useTranslation()
   const { isMobile } = useMatchBreakpoints()
-  const nftList = useNftsFromCollection('0x60935f36e4631f73f0f407e68642144e07ac7f5e')
+  const nftList = useNftsFromCollection(pancakeBunniesAddress)
```
ChefHutch, line 42: See other comment RE not hard-coding this value

RabbitDoge (approved):
Some comments, but it's ok

src/views/Nft/market/Profile/components/SubMenu.tsx
ChefHutch, line 13: Revert these - this is the old URL structure

src/state/nftMarket/hooks.ts
hachiojidev: The only reason for the constant was because we had it hard-coded everywhere. Just to avoid lowercase and checksummed differences I put it into a temporary variable.

src/views/Nft/market/Profile/components/SubMenu.tsx
hachiojidev: Just a revert from the rebase.

PR #2259, opened 2021-09-20
fix: (pools & farms) To top shows the first row item under the fixed header

memoyil:
To reproduce:

1. Go to pools or farms table
2. Go to bottom of the table
3. Click "To top"
4. See the first row of the table is under the fixed header


ChefHutch (approved):
Works better, nice one

hachiojidev (approved):

PR #2265, opened 2021-09-20
fix: Make a profile doesn't show collectible after minting in set profile picture

memoyil:
To review:

https://deploy-preview-2265--pancakeswap-dev.netlify.app/

To reproduce:

1. Go to make a profile with an account that doesn't have profile https://pancakeswap.finance/profile
2. Complete first step
3. See We couldn’t find any Pancake Collectibles in your wallet. in the next step


ChefHutch (approved):
Not 100% this fixes the issue all the time - but it's a logical thing to do.

All of this logic is getting migrated and changed with the new nft market regardless

hachiojidev (approved):

PR #2269, opened 2021-09-20
chore(nft-market): Create const for '/nfts' route

ChefHutch:
- Create `nftsBaseUrl` const. 
- Use throughout nfts links and routing.


src/views/Nft/market/Profile/index.tsx
```diff
@@ -46,14 +47,14 @@ const NftProfile = () => {
         </TabMenuWrapper>
       </MarketPageHeader>
       <Page style={{ minHeight: 'auto' }}>
-        <Route path="/nfts/profile/achievements">
+        <Route path={`${nftsBaseUrl}/profile/achievements`}>
```
hachiojidev, line 50: So in this case we should use `useRouteMatch`'s path I think.

hachiojidev (approved):

PR #2270, opened 2021-09-20
feat(buy-sell-modals): Cheap way to trigger user nfts hook

ChefHutch:
Cheap solution - needs something outwith a hook than can be triggered within an `onSuccess` callback


Chef-Cheems (approved):

ChefHutch:
New actions being added for this https://github.com/pancakeswap/pancake-frontend/pull/2289

PR #2286, opened 2021-09-21
feat(nft-market): Move lowestPriceToken logic into redux

ChefHutch:
Proof of concept for how this could be stored in redux and accessed in components (for PBs).

It works, but it has two dependencies that won't scale well. 

1. It relies on us retrieving and storing an array of all token IDs for each NFT. As an example - Sleepy's tokens array will have 98k ids on mainnet.
2. It relies on sorting the market data for _all_ tokens of an NFT type. When thousands of tokens are on sale - we would need to gather this data over many requests.

And another thing:
 - The `useGetLowestPricedPB` hook will only work for PBs because it uses the PB attributes to access the nft in the `collections` redux.

Success:
Screenshot 2021-09-21 at 13 25 25

But not scalable:
Image from iOS (1)

### DIFF
--- a/src/state/nftMarket/hooks.ts
+++ b/src/state/nftMarket/hooks.ts
@@ -59,6 +59,15 @@ export const useUserActivity = (): (Transaction | AskOrder)[] => {
   return []
 }
 
+export const useGetLowestPricedPB = (nft: NFT): NftTokenSg => {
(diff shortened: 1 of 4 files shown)


R-K-H:
Screen Shot 2021-09-21 at 9 00 31 AM

Let's get you a more up to date screenshot.

:-p

hachiojidev (approved):

PR #2290, opened 2021-09-21
chore: Resolve type mismatches, add missing types, refactor

Chef-Cheems:


src/views/Nft/market/Collection/Items/CollectionNfts.tsx
```diff
@@ -45,7 +45,13 @@ const CollectionNfts: React.FC<CollectionNftsProps> = ({ collection, sortBy = 'u
       alignItems="start"
     >
       {orderedNfts.map((nft) => {
-        return <CollectibleLinkCard key={`${nft.tokenId}-${nft.collectionName}`} nft={nft} />
+        return (
+          <CollectibleLinkCard
+            key={`${nft.name}-${nft.collectionName}`}
```
Chef-Cheems, line 50: In this case these nfts are `PancakeBunnyNftWithTokens`, i.e. different pancake bunnies with tokens. They don't have tokenId (on main level), but name is unique.
We might consider storing bunnyId if we would fetch all attributes, but I'm not sure its needed at this stage. Can be improved later of course, this PR is not "final fix for everything"

src/views/Nft/market/components/CollectibleCard/CardBody.tsx
```diff
@@ -1,21 +1,17 @@
...
 import { CollectibleCardProps } from './types'
 
-const CollectibleCardBody: React.FC<CollectibleCardProps> = ({ nft, nftLocation, currentAskPrice }) => {
+const CollectibleCardBody: React.FC<CollectibleCardProps> = ({ nft, nftLocation, currentAskPrice, lowestPrice }) => {
   const { t } = useTranslation()
-  const { name, image, tokens } = nft
+  const { name, image } = nft
   const bnbBusdPrice = useBNBBusdPrice()
 
-  const lowestPriceToken = tokens && minBy(Object.values(tokens), 'currentAskPrice')
```
Chef-Cheems, line 16: This actually haven't worked before. 
- In Newest component `tokens` were passed here as "hacked" tokens with only 1 element
- in CollectionNfts.tsx lowestPrice was already calculated and populated into `meta` property
- MoreFromThisCollection is skeleton UI component with bunch of BS inside

Lowest price comes as separate prop since it is easier to calculate it up in hierarchy.

src/views/Nft/market/Home/Newest.tsx
```diff
@@ -1,41 +1,57 @@
...
-  return state
+  return newestNfts
 }
 
 const Newest: React.FC = () => {
   const nfts = useNewestNfts()
 
+  // Get lowest price among same PancakeBunnies
+  // Note - most certainly temporary, not scalable on mainnet
+  const lowestPrices = nfts
```
Chef-Cheems, line 41: Find lowest price for each bunny and store it in map.
Obviously temp solution, still unknown how to handle mainnet amounts of tokens...

src/views/Nft/market/components/ProfileNftModal.tsx
```diff
@@ -52,7 +52,7 @@ const ProfileNftModal: React.FC<ProfileNftModalProps> = ({ nft, onDismiss }) =>
           </Text>
         </TextWrapper>
         <Flex flexDirection="column" pt="4px" pb="16px" px="16px" alignItems="center">
-          <Button as={Link} variant="text" mb="8px" href={`${BASE_URL}/nft/market/item/${nft.tokenId}`}>
+          <Button as={Link} variant="text" mb="8px" href={`${BASE_URL}/nft/market/item/${nft.name}`}>
```
Chef-Cheems, line 55: Right now there's no easy way to access bunny id.
Will figure out this part in next PR, for now its just fix so it doesn't throw errors

src/state/nftMarket/helpers.ts
```diff
@@ -254,101 +351,29 @@ export const fetchWalletTokenIdsForCollections = async (
   return walletNfts.flat()
 }
 
-/**
- * Get the most recently listed NFTs
- * @param first Number of nfts to retrieve
- * @returns NftTokenSg[]
- */
-export const getLatestListedNfts = async (first: number): Promise<NftTokenSg[]> => {
```
Chef-Cheems, line 262: Lots of lines in diff here are just moving functions up or down (to group API, Subgraph and other helpers).
Most real changes are just types and some other minor things

src/state/nftMarket/helpers.ts
```diff
@@ -254,101 +351,29 @@ export const fetchWalletTokenIdsForCollections = async (
...
-
 /**
  * Helper to combine data from the collections' API and subgraph
- * TODO: Type this properly
  */
-export const combineCollectionData = (collectionApiData: any, collectionSgData: any): Record<string, Collection> => {
-  const collectionsMarketObj = collectionSgData.reduce(
+export const combineCollectionData = (
+  collectionApiData: ApiCollection[],
+  collectionSgData: CollectionMarketDataBaseFields[],
```
Chef-Cheems, line 359: This is typed now

src/state/nftMarket/helpers.ts
```diff
@@ -254,101 +351,29 @@ export const fetchWalletTokenIdsForCollections = async (
...
     (prev, current) => ({ ...prev, [current.id]: { ...current } }),
     {},
   )
 
   return collectionApiData.reduce((accum, current) => {
     const collectionMarket = collectionsMarketObj[current.address.toLowerCase()]
+    const collection: Collection = {
+      ...current,
+      ...collectionMarket,
+      slug: slugify(current.name),
```
Chef-Cheems, line 371: Not sure if `slug` field is actually used anywhere now, I think there's just one place. Plugged it here since its the only field missing from internal `Collection` type. Previously slug field was undefined. 🤔

src/state/nftMarket/types.ts
```diff
@@ -93,23 +87,43 @@ export interface NftTokenSg {
...
-  nftLocation?: NftLocation
 }
 
-export interface NFT {
-  tokenId: string
+// This type is somewhat specific to PancakeBunnies NFTs
+// It stores basic PancakeBunny NFT data together with `tokens` object containing specific token ID market data
+// It is used when retrieving all bunny tokens from redux.
+// In case you need to work with specific stand-alone token use NftToken
+export interface PancakeBunnyNftWithTokens {
```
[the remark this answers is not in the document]
Chef-Cheems: Also some fields like `collectionName` and `collectionAddress` were not populated at all. Again TS didn't complain cause of missing types..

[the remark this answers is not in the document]
hachiojidev: I am a bit confused on the name. This interface represents any NFT from any collection correct? I think to avoid confusion the name should be more generic.

src/state/nftMarket/helpers.ts
hachiojidev: You can remove it.

Chef-Cheems: Oh yeah, you're right, its not used at all, the slug used in that one place was from different `Collection` type.
Removed this one

src/state/nftMarket/types.ts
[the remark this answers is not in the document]
Chef-Cheems: @hachiojidev Ideally in the future when we integrated Squads and other collections this should be reused, yes. But in current form its not clear if we'll able to just reuse it as is, some stuff is too PancakeBunny-specific, especially `tokens` field, right now its one of the main ones that we use in a lot of places but for Squads and other collection it won't exist at all.

as Hutch said in DM

> it makes it _super_ obvious what that type is (easy to find and replace all names later if we decide to)

hachiojidev (approved):

PR #2295, opened 2021-09-22
feat: PB traits

hachiojidev:


src/state/nftMarket/helpers.ts
```diff
@@ -58,16 +58,14 @@ export const getCollectionApi = async (collectionAddress: string): Promise<ApiCo
...
   if (res.ok) {
-    const json = await res.json()
-    return json.data
+    const data = await res.json()
+    return data
   }
   console.error(`API: Failed to fetch NFT tokens for ${collectionAddress} collection`, res.statusText)
-  return {}
+  return null
 }
```
hachiojidev, line 69: This function needs to return the entire response.

ChefHutch (commented):
The value is going to be used by all PB collectible cards too - so we'll need to access it across the app. IMO I think we need to build this fetch into the `fetchNftsFromCollections` action rather than fetched each time a component needs it? 

Currently it's being stored as a token, to `lowestPricedToken` on `PancakeBunnyNftWithTokens`

https://github.com/pancakeswap/pancake-frontend/blob/a11819f414b64bdc3f1fa908cf9f1f9815fc98dd/src/state/nftMarket/reducer.ts#L98-L101

This could maybe be changed to just a price?

Either way we want to replace that logic in `fetchNftsFromCollection` because it's reliant on the tokens array containing every token.

We have a hook - `useGetLowestPricedPB` which accesses that `lowestPricedToken` value for a given nft.

Chef-Cheems (approved):

PR #2306, opened 2021-09-22
chore: Fix typos

memoyil:
To review:

https://deploy-preview-2306--pancakeswap-dev.netlify.app/


src/state/info/queries/tokens/tokenData.ts
```diff
@@ -81,11 +81,11 @@ const fetchTokenData = async (
     return { data, error: false }
   } catch (error) {
     console.error('Failed to fetch token data', error)
-    return { erro: true }
+    return { error: true }
```
memoyil, line 84: This could have been an issue since error attribute in the object would be undefined if we return this object

ChefKai (approved):

hachiojidev (approved):

PR #2314, opened 2021-09-23
fix: Links on homepage refreshes the website and firefox scroll to top

memoyil:
To review:

https://deploy-preview-2314--pancakeswap-dev.netlify.app/

To reproduce: 

1. Click trade now or play buttons on homepage and see the website refreshes completely
2. In firefox we have to give small timeout in order to scroll to work

### DIFF
--- a/src/hooks/useScrollOnRouteChange.ts
+++ b/src/hooks/useScrollOnRouteChange.ts
@@ -4,11 +4,13 @@ import history from 'routerHistory'
 const useScrollOnRouteChange = () => {
   useEffect(() => {
     const unlisten = history.listen(() => {
-      window.scrollTo({
-        top: 0,
-        left: 0,
-        behavior: 'smooth',
-      })
+      setTimeout(() => {
+        window.scroll({
+          top: 0,
+          left: 0,
+          behavior: 'smooth',
+        })
+      }, 50)
     })
 
     return () => unlisten()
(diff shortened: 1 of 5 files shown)


src/views/Home/components/Hero.tsx
```diff
@@ -113,7 +114,7 @@ const Hero = () => {
           </Heading>
           <Flex>
             {!account && <ConnectWalletButton mr="8px" />}
-            <Link mr="16px" href="/swap">
```
memoyil, line 116: margin right is not needed here since no component is there after it

src/hooks/useScrollOnRouteChange.ts
```diff
@@ -3,12 +3,24 @@ import history from 'routerHistory'
...
+            behavior: 'smooth',
+          })
+        }, 50)
+      } else {
+        window.scroll({
+          top: 0,
+          left: 0,
+          behavior: 'smooth',
+        })
+      }
```
hachiojidev, line 23: Lets just add the setTimeout on everything to simplify the code.

hachiojidev (changes requested):

src/hooks/useScrollOnRouteChange.ts
memoyil: Done

hachiojidev (approved):

PR #2315, opened 2021-09-23
refactor: Token array removal and other cleanup

Chef-Cheems:


src/state/nftMarket/reducer.ts
```diff
@@ -73,47 +69,58 @@ export const fetchCollection = createAsyncThunk<Record<string, Collection>, stri
...
-          : null
-      return {
-        name: nft.name,
-        description: nft.description,
-        collectionName: nft.collection.name,
-        collectionAddress,
-        image: nft.image,
-        tokens,
-        attributes,
+      if (collectionAddress === pancakeBunniesAddress) {
```
Chef-Cheems, line 85: here's the most intense change - mapping happens a bit different cause we now return `NftToken[]` from here and not object with pancakebunnies

src/state/nftMarket/types.ts
```diff
@@ -21,7 +21,7 @@ export interface State {
   initializationState: NFTMarketInitializationState
   data: {
     collections: Record<string, Collection> // string is the address
-    nfts: Record<string, Record<string, PancakeBunnyNftWithTokens>> // first string is the collection address address, second string is bunny id
+    nfts: Record<string, NftToken[]> // string is the collection address
```
ChefHutch, line 24: I guess now bunny ID is implicitly index position in array? Are we comfortable with this? I'd maybe prod @RabbitDoge?

All looks good otherwise 👍

Chef-Cheems: No. From this point on we forget about bunny ids.
`nfts` state (for **any** collection) is just array of `NftToken`
For squads its ideal situation since they are all "unique", just pile em up into `nfts` and that's it.
For bunnies it is the same, its all just 100 random bunnies in `nfts`, you can get then and display on any page. If you need bunny id - you can address each token's bunny id via `nft.attributes[0].value` (since bunnyId is the only attribute). Or there's other way with `nft.marketData.otherId` - same thing but from subgraph.

There are couple of places in code that does that.
In the individual NFT page I'll make a hook to get all tokens with specific bunny id `useNftsWithBunnyId(bunnyId)` to get all similar ones if needed (I think that hook would only be needed on individual page though).

ChefHutch: Got it man - makes sense, let's get it in 👍

ChefHutch (approved):

PR #2103, opened 2021-09-06
fix: Claim button is visible even after the claiming deadline passed

memoyil:
Fixes #2098 

Not sure if this is the correct way, @ChefNina and @Chef-Chungus can you confirm?

To review: 

https://deploy-preview-2103--pancakeswap-dev.netlify.app/

Issue #2098, opened 2021-09-06
[BUG] Collectibles - Claim button is visible even after the claiming deadline passed.


ChefNina:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Product

Other

### Current Behavior

The claiming deadline is passed already, but collectibles page still shows “Claim” button.
![image](https://user-images.githubusercontent.com/82930772/132593194-ee5f7040-5495-49b5-b4dc-be7d65cf1ba0.png)

### Expected Behavior

Once the claiming deadline is passed, "Claim" button is not shown on collectibles page.

### Steps To Reproduce

Go to collectibles page with the claiming right for NFT which claiming deadline is passed.

### Environment

```markdown
- Device: any
- OS: any
- Browser: any
- Wallet: any
```

memoyil:
the image is missing

ChefNina:
re-updated the image!

memoyil:
> re-updated the image!

Fix is ready you can test it

PR #2336, opened 2021-09-25
Revert "fix: Claim button is visible even after the claiming deadline passed (#2103)" for lottery nfts

memoyil:
This will revert the fix for the issue https://github.com/pancakeswap/pancake-frontend/issues/2098 since the claim period is extended. 

@Chef-Cheems @ChefNina @0xCorgi Can you confirm this is the issue?


Chef-Cheems:
@memoyil I assume this PR is intended to show the claim button on collectibles page? We recently reverted the change for claiming lottery NFTs in #2335, so no need for that now. Gonna fix it on Monday (or tomorrow if anyone is available).
Although maybe your PR does indeed fixes it. I'll ping @ChefHutch and @0xCorgi , they know more about this stuff.

Chef-Cheems:
Seems like this is not needed anymore, sorry :(

PR #2342, opened 2021-09-26
fix: Trade volume fix

Chef-Cheems:
Subgraph response is just a string, no need for ethers conversions

### DIFF
--- a/src/views/Nft/market/Collection/Header.tsx
+++ b/src/views/Nft/market/Collection/Header.tsx
@@ -1,10 +1,8 @@
 import React from 'react'
 import { useLocation, useParams } from 'react-router'
-import { ethers } from 'ethers'
 import { Text } from '@pancakeswap/uikit'
 import { Collection } from 'state/nftMarket/types'
 import { useTranslation } from 'contexts/Localization'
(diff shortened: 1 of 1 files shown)


ChefKai (approved):

src/views/Nft/market/Collection/Header.tsx
```diff
@@ -25,8 +23,11 @@ const Header: React.FC<HeaderProps> = ({ collection }) => {
   const { totalSupply, numberTokensListed, totalVolumeBNB, banner, avatar } = collection
   const { t } = useTranslation()
   const { pathname } = useLocation()
-  const owners = ethers.BigNumber.from(numberTokensListed)
-  const volume = ethers.BigNumber.from(totalVolumeBNB)
+
+  const volume = parseFloat(totalVolumeBNB).toLocaleString(undefined, {
```
ChefKai, line 27: We should have an helper for this tbh (if not already here)

Chef-Cheems: yeah its the same in bunch of places, agree, good item to refactor 👍

PR #2354, opened 2021-09-27
fix: (info): Wrong tab bar in info

memoyil:
This fix also contains https://github.com/pancakeswap/pancake-frontend/pull/2355 fixes by @ChefHutch to nftmarket branch.

Fixes #2214 as well

To review:

https://deploy-preview-2354--pancakeswap-dev.netlify.app/

1. Go to info https://pancakeswap.finance/info
2. Click pools on its own tab bar
3. See farms pools(which is syrup pools) tab bar shown above that

Issue #2214, opened 2021-09-16
[BUG] Tab system on INFO needs to be updated to match menu style


Chef-Chungus:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Product

Info

### Current Behavior

on https://pancakeswap.finance/info/pools there are two different tab systems. 
![image](https://user-images.githubusercontent.com/71833681/133567940-e81f907c-69d2-4f8d-b008-3893ef01205e.png)
(note the sub-menu doesn't cover all of info, only /info/pools/)

### Expected Behavior

We should apply the sub-menu to the entire info as designed in the figma here
https://www.figma.com/file/zmeOAYsXghgSysqoDTObg7/Info-Site?node-id=0%3A1

### Steps To Reproduce

navigate to https://pancakeswap.finance/info/pools

### Environment

```markdown
N/A
```

PR #2355, opened 2021-09-27
chore(menu-items): Improve getActivesubMenuItem specificity

ChefHutch:
Previous logic uses `.find()` & return the first match for `pathname.includes(menuItem.href)` which means that if you have an array of menuItems like `/nfts`, `/nfts/collection/0x...`, and a pathname  `/nfts/collection/0x...` - it returns `/nfts` as the active item.

This fixes it. Interested in what @0xCorgi thinks though. It doesn't feel particularly elegant.

Screenshot 2021-09-27 at 12 31 44

### DIFF
--- a/src/components/Menu/utils.ts
+++ b/src/components/Menu/utils.ts
@@ -6,9 +6,9 @@ export const getActiveMenuItem = ({ pathname, menuConfig }: { pathname: string; 
 export const getActiveSubMenuItem = ({ pathname, menuItem }: { pathname: string; menuItem?: ConfigMenuItemsType }) => {
   const activeSubMenuItems = menuItem?.items.filter((subMenuItem) => pathname.includes(subMenuItem.href))
 
-  // Pathname doesn't include any submenu item href - return null
+  // Pathname doesn't include any submenu item href - return undefined
   if (activeSubMenuItems.length === 0) {
-    return null
+    return undefined
   }
 
   // Pathname includes one sub menu item href - return it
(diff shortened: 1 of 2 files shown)


memoyil:
@ChefHutch @0xCorgi Check this pr https://github.com/pancakeswap/pancake-frontend/pull/2354 , It uses starts with

ChefHutch:
> @ChefHutch @0xCorgi Check this pr #2354 , It uses starts with

Much more elegant, but doesn't fix the issue. See my comment.

src/components/Menu/utils.ts
```diff
@@ -1,7 +1,28 @@
...
+
+  // Pathname includes one sub menu item href - return it
+  if (activeSubMenuItems.length === 1) {
+    return activeSubMenuItems[0]
+  }
+
+  // Pathname includes multiple sub menu item hrefs - find the most specific match
+  const splitIntoParts = activeSubMenuItems.map((subMenuItem) => {
+    return { label: subMenuItem.label, href: subMenuItem.href.split('/') }
+  })
```
0xCorgi, line 22: We could also check the `href` length for sorting instead of splitting it.

0xCorgi (commented):
Good for this fix !
Might be worth adding a small test case for this (the test file already exist [here](https://github.com/pancakeswap/pancake-frontend/blob/develop/src/__tests__/utils/utils.test.ts#L41))

memoyil:
Should we apply this fix directly to develop? It will fix this https://github.com/pancakeswap/pancake-frontend/issues/2214

0xCorgi (approved):
All good !

PR #2367, opened 2021-09-27
fix number tickets available in the pre-sale

dsimanov:
fix number tickets available in the pre-sale for owners of Gen0 (id 1-4)

### DIFF
--- a/src/views/PancakeSquad/components/EventDescriptionSection/config.tsx
+++ b/src/views/PancakeSquad/components/EventDescriptionSection/config.tsx
@@ -17,7 +17,7 @@ const eventDescriptionConfigBuilder = ({ t }: EventDescriptionType) => ({
       id: 1,
       content: (
         <>
-          {t('490 are available in the pre-sale for owners of ')}
+          {t('360 are available in the pre-sale for owners of ')}
           <Text display="inline-block" color="primary" bold>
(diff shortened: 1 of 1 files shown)


dsimanov:
Hello =) What are your plans for the burnt nft gen0? The sum of the first 4 issued nft = 112 + 101 + 74 +73 = 360. But some of the tokens were burned for cakes and they are now out of circulation. Then the fair number of participants in the presale is gene0 = (112 - 63) + (101 - 16) + (74 - 6) + (73 - 4) = 271

pl4novaya:
> Hello =) I saw that you reduced the number of collections for Gen0 from 5 to 4. Therefore, the total number of tickets should also decrease from 490 to 360. But perhaps this is not fair) What are your plans for the burnt NFT Gen0? The sum of the first 4 minted NFT = 112 + 101 + 74 +73 = 360. But some of the tokens were burned for Cakes and they are now out of circulation. Then the fair number of participants in the pre-sale for owner of Gen0 = (112 - 63) + (101 - 16) + (74 - 6) + (73 - 4) = 271
> Good luck with the launch of the market! We are all very much looking forward to this and are grateful to you for your work. Thanks a lot for what you are doing!

That is not correct. The ids are 0,1,2,3,4. It’s 490 exist. 110 are burned.

PR #2368, opened 2021-09-27
Feature/pancake squad

dsimanov:
Hello =) What are your plans for the burnt NFT Gen0? The sum of the first 5 collections minted NFT is 490. But some of the tokens were burned for Cakes and they are now out of circulation. Then the fair number of participants in the pre-sale for owner of Gen0 = (112 - 63) + (101 - 16) + (74 - 6) + (73 - 4) + ( 130 - 21) = 380.

Good luck with the launch of the market! We are all very much looking forward to this and are grateful to you for your work. Thanks a lot for what you are doing!


hachiojidev:
@dsimanov Thank you for catching this. We have plans on updating this so I will close this for now.

PR #2379, opened 2021-09-28
feat: PancakeSquad landing

0xCorgi:
Add page for Pancake Squad event.


.gitignore
```diff
@@ -35,3 +35,6 @@ yarn-debug.log*
 yarn-error.log*
 debug.log*
 
+# ABI
+nftSale.json
+pancakeSquad.json
```
hachiojidev, line 40: This is temporary right?

src/views/Home/components/ColoredWordHeading.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+interface HeadingProps extends TextProps {
+  text: string
+  firstColor?: keyof Colors
+}
+
+const ColoredWordHeading: React.FC<HeadingProps> = ({ text, firstColor, mb = '24px', ...props }) => {
+  const { theme } = useTheme()
+  const split = text.split(' ')
+  const firstWord = split[0]
```
hachiojidev, line 13:
```
const [firstWord] = text.split(' ')
```

src/views/PancakeSquad/components/ArtistSection/config.tsx
```diff
@@ -0,0 +1,41 @@
...
+    'Cecy´s birth place is definitely unknown',
+    'But legend said she was sailing all alone',
+    'Through the mountains and across the sea',
+    'She found an island in which to live',
+    'Was full of rabbitts!! that caught her attention ',
+    'making stuff beyond her comprehension',
+    'The bunnies were working on a kitchen ',
+    'baking pancakes with one Big mission ',
+    'To send sweet bread very soon',
+    'so they all together can reach the moon.',
```
hachiojidev, line 22: We are not going to translate these?

src/views/PancakeSquad/hooks/useEventInfos.tsx
```diff
@@ -0,0 +1,59 @@
...
+        }))
+
+        const [
+          currentMaxSupply,
+          currentMaxPerAddress,
+          currentPricePerTicket,
+          curentMaxPerTransaction,
+          currentTotalTicketsDistributed,
+          currentSaleStatus,
+          currentStartTimestamp,
```
hachiojidev, line 35:
```
const [
  [currentMaxSupply],
  ...
]
```
?

src/views/PancakeSquad/components/ArtistSection/config.tsx
0xCorgi: No good catch, it should.

.gitignore
0xCorgi: Yes

src/views/Home/components/ColoredWordHeading.tsx
0xCorgi: We need `split` just below, probably need to keep the variable here.

src/hooks/useApproveConfirmTransaction.ts
```diff
@@ -118,20 +122,22 @@ const useApproveConfirmTransaction = ({
...
+    handleConfirm: async (params = {}) => {
       dispatch({ type: 'confirm_sending' })
       try {
-        const tx = await onConfirm()
+        const tx = await onConfirm(params)
         const receipt = await tx.wait()
         if (receipt.status) {
           dispatch({ type: 'confirm_receipt' })
           onSuccess({ state, receipt })
+          if (showSuccessToast) toastSuccess(t('Transaction has succeeded!'))
```
Chef-Cheems, line 140: I think in other places where this hook is used we just handle toasts inside `onApproveSuccess ` and `onSuccess` handler with customised message depending on the contract. Take a look at lottery/farmAuction/nftMarket, they all use this hook.

src/state/nftMarket/hooks.ts
```diff
@@ -97,3 +99,8 @@ export const useGetNFTInitializationState = () => {
 export const useUserNfts = (): UserNftsState => {
   return useSelector((state: State) => state.nftMarket.data.user)
 }
+
+export const useHasGen0Nfts = (): boolean => {
+  const userNfts = useSelector((state: State) => state.nftMarket.data.user)
+  return userNfts.nfts.some((nft) => Number(nft.tokenId) <= MAX_GEN0_ID)
```
Chef-Cheems, line 105: `tokenId` is not Gen0 identifier, it is ordinal "mint" number of this bunny. You need to check `nft.attributes[0].value`.
For PancakeBunnies `attribute` field looks like this:
```
[
  {
      traitType: 'bunnyId',
      value: 3,
      displayType: null,
  }
]
```
so you take 0st attribute and `value` is the bunnyId number

src/views/Home/components/ColoredWordHeading.tsx
Chef-Cheems: I wonder how it would work with Japanese/Chinese where there are no spaces 😅 
Did a quick test - seems to be coloring whole sentence and `remainingWords` becomes `''`.
But I guess there's no need to cover for that, not sure we'll even have translations by the time (although if we'd keep the page for long time maybe worth checking what to do with it)

src/state/nftMarket/hooks.ts
0xCorgi: Oh I wasn't aware of that. Will change thanks.

src/hooks/useApproveConfirmTransaction.ts
0xCorgi: Thought it might be useful to have this behavior in the hook but I can extract it.

src/views/Home/components/ColoredWordHeading.tsx
Chef-Cheems: oh lol, just realized it is upgrade of old PurpleWordHeading, so it was like this for a long time 😄

src/views/PancakeSquad/assets/ArtistBottomWave.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react'
+import { Svg, SvgProps } from '@pancakeswap/uikit'
+
+const Icon: React.FC<SvgProps & { isDark: boolean }> = ({ isDark, ...props }) => {
```
Chef-Cheems, line 4: I suppose this is the curves between page sections?
We actually have `CurveDivider` component that handles it
But I'd suggest leave it like you have now, its quite a pain to do them sometimes, not worth the time 😅 
Just noting for the future 🙂

src/views/PancakeSquad/assets/EventStepsBottomWave.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+const Icon: React.FC<SvgProps & { isDark: boolean }> = ({ isDark, ...props }) => {
+  return (
+    <Svg preserveAspectRatio="none" viewBox="0 0 1280 17" fill="none" xmlns="http://www.w3.org/2000/svg" {...props}>
+      <path
+        d="M0 16.002V0.00195312C159.43 9.20767 387.163 14.9682 640 14.9682C892.837 14.9682 1120.57 9.20767 1280 0.00195312V16.002H0Z"
+        fill="url(#paint0_linear)"
+      />
+      <defs>
+        <linearGradient id="paint0_linear" gradientUnits="userSpaceOnUse">
```
Chef-Cheems, line 12: Just a note - its better to give more descriptive name to svg ids. We had a problem where some ids were had the same name and were interfering with each other. E.g. if some other `<linearGradient>` on the page would use `paint0_linear` id but with different colors - this might use that other `pain0_linear` instead of the one here.

src/views/Home/components/ColoredWordHeading.tsx
0xCorgi: Yes indeed, I just renamed/changed it a bit to make it more generic.

src/views/PancakeSquad/components/Buttons/ActivateProfile.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { UserStatusEnum } from 'views/PancakeSquad/types'
+
+type ActivateProfileButtonProps = {
+  t: ContextApi['t']
+  userStatus: UserStatusEnum
+}
+
+const ActivateProfileButton: React.FC<ActivateProfileButtonProps> = ({ t, userStatus }) => (
+  <>
+    {(userStatus === UserStatusEnum.NO_PROFILE || userStatus === UserStatusEnum.UNCONNECTED) && (
```
Chef-Cheems, line 14: I'd personally prefer simple
```jsx
if (userStatus === UserStatusEnum.NO_PROFILE || userStatus === UserStatusEnum.UNCONNECTED) {
  return (<Button as={Link} to="/create-profile" mr="4px">
        {t('Activate Profile')}
   </Button>)
}
return null
```
but its just a nitpick of course :)

src/views/PancakeSquad/components/EventStepsSection/styles.tsx
```diff
@@ -0,0 +1,16 @@
+import { Flex } from '@pancakeswap/uikit'
+import styled from 'styled-components'
+
+export const StyledEventStepsSectionContainer = styled(Flex)<{ $isDark: boolean }>`
+  position: relative;
+  background: ${({ $isDark }) =>
+    $isDark
+      ? 'linear-gradient(180deg, #0B4576 0%, #091115 100%);'
+      : 'linear-gradient(180deg, #6fb6f1 0%, #eaf2f6 100%);'};
```
Chef-Cheems, line 9: This seems to put `;;` in css. Probably doesn't break anything though.

src/views/PancakeSquad/components/Header/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+        </Link>
+      </Flex>
+      <StyledSquadTitle my="32px" color={lightColors.invertedContrast} bold textAlign="center">
+        {t('Pancake Squad')}
+      </StyledSquadTitle>
+      <Text color={lightColors.warning} textAlign="center" bold>
+        {t('Mint Cost: To Be Announced', {
+          minCost: pricePerTicket ? formatBigNumber(pricePerTicket, 0) : DEFAULT_CAKE_COST,
+        })}
+        <br />
```
Chef-Cheems, line 79: Usually in other components we don't use `br` and just wrap text into multiple `<Text>` components. Sorry for nitpicks 😅

src/views/PancakeSquad/components/Buttons/ActivateProfile.tsx
0xCorgi: Do we have a convention in the team on that ? If not we could decide on one.

src/views/PancakeSquad/components/EventStepsSection/styles.tsx
0xCorgi: Your probably right, will clean that.

src/views/PancakeSquad/components/Header/index.tsx
0xCorgi: Can do, no problem.
Using `<br />` in place of a react component might be better for performance in this case though.

src/views/Nft/market/Collection/IndividualNFTPage/MoreFromThisCollection.tsx
```diff
@@ -97,9 +103,11 @@ const MoreFromThisCollection: React.FC<MoreFromThisCollectionProps> = ({ collect
 
   return (
     <Box pt="56px" mb="52px">
-      <Text bold mb="24px">
-        {t('More from this collection')}
-      </Text>
+      {title && (
+        <Text bold mb="24px">
+          {t(title)}
```
Chef-Cheems, line 108: Should it be title? 🤔 
And even if it is better - probably no need to `t()` it - I don't think collection names are translatable?

Chef-Cheems: Also title will be present all the time since it has default value, no need to `&&` it

0xCorgi: Title is not the collection name here. The component was always displaying `More from this collection` above the slider.
I made it a prop so the component can be more reusable.

src/views/PancakeSquad/components/FaqSection/config.tsx
```diff
@@ -0,0 +1,51 @@
+import React from 'react'
+import { Text } from '@pancakeswap/uikit'
+import { ContextApi } from 'contexts/Localization/types'
+import { StyledLinkFAQs } from './styles'
```
Chef-Cheems, line 4: 'StyledLinkFAQs' is defined but never used

Chef-Cheems (approved):

src/views/PancakeSquad/components/FaqSection/config.tsx
```diff
@@ -0,0 +1,50 @@
...
+  {
+    title: 'I can’t see my NFT’s picture!',
+    description: [
+      'Wait for the reveal! After all 10,000 members of the Pancake Squad have been minted, their images will remain hidden for a few days. Just be patient :)',
+    ],
+  },
+  {
+    title: 'How many can I mint?',
+    description: [
+      'The max limit per wallet is 20 NFTs.',
```
R-K-H, line 27: This number should be 10 not 20 as outlined here.

src/views/PancakeSquad/components/FaqSection/config.tsx
```diff
@@ -0,0 +1,50 @@
...
+      'Users holding Gen 0 Pancake Bunny NFTs at the snapshot may also purchase one additional Pancake Squad NFT in the presale for each Pancake Bunny they hold.',
+      'For example, if you have 5 Gen 0 bunnies, you can buy 5 minting tickets in the presale, then max. 20 in the public sale.',
```
R-K-H, line 29: Again 10 not 20.

PR #2381, opened 2021-09-28
chore: Reuse useGetProfileAvatar hook and add loading flags

Chef-Cheems:


src/state/profile/hooks.ts
```diff
@@ -18,34 +17,14 @@ export const useFetchProfile = () => {
   }, [account, dispatch])
 }
 
-export const useProfileForAddress = (address: string) => {
```
Chef-Cheems, line 21: This is similar hook I made recently, yeet 🗑️

src/state/profile/hooks.ts
```diff
@@ -55,10 +34,10 @@ export const useGetProfileAvatar = (account: string) => {
       dispatch(fetchProfileAvatar(account))
     }
 
-    if (!username && nft && address) {
+    if (!username && address) {
```
Chef-Cheems, line 37: Users might remove their profile picture but still have username, I think this should not check for `nft`.
There is a drawback that for unregistered users it will spam console with 404 errors, but I think showing as many usernames as possible is more important 🤔

src/state/profile/index.tsx
```diff
@@ -56,16 +56,39 @@ export const profileSlice = createSlice({
     },
   },
   extraReducers: (builder) => {
+    builder.addCase(fetchProfileUsername.pending, (state, action) => {
```
Chef-Cheems, line 59: Added loading flags to show loaders while waiting for profile pic / username

src/state/profile/hooks.ts
ChefHutch: Link to yours Cheems? Should yours now live in this file?

Chef-Cheems: What do you mean? This is the hook I used before I learned that Hachi has one already. Was used on individual NFT page only, now replaced with Hachi's hook

src/state/profile/hooks.ts
ChefHutch: Is there another datapoint that could be added to the `profileAvatars` store to prevent this? I'm not sure how it's all structured but we have a `hasRegistered` value on profiles. It requires its own node call to fetch though 😬 

https://github.com/hachiojidev/pancake-frontend/blob/e6edf4816f5d1d5f94fbc1d98437eccc1fea2cca/src/state/profile/helpers.ts#L73

Chef-Cheems: yeah it could be done, but it would require some refactoring in the profile code that _might_ be breaking 🥲 
Wouldn't want to get stuck in it...

ChefHutch: I'm actually torn on this one man. I think deactivated accounts (username but no nft) are going to be much less common than never-activated accounts (no username, no nft). And it's the second, more common occurrence, that is gonna fire network requests that 404.

I'd probably return this check to as it was previously until we have a reliable way of identifying the deactivated account scenario.

Chef-Cheems: actually one sec, it seems that `getProfileAvatar` is not used anywhere else....

src/state/profile/hooks.ts
ChefHutch: Ah sorry - I thought this was Hachi's hook that you were trashing in favour of yours. All good 👍

src/state/profile/helpers.ts
```diff
@@ -57,14 +57,14 @@ export const getProfileAvatar = async (address: string) => {
...
-    if (!isActive) {
-      return null
+    let nft = null
+    if (isActive) {
+      nft = await getNftByTokenId(nftAddress, tokenId)
     }
 
-    const nft = await getNftByTokenId(nftAddress, tokenId)
-    return nft
+    return { nft, hasRegistered }
```
Chef-Cheems, line 65: Returning hasRegistered here now

src/state/profile/hooks.ts
```diff
@@ -18,47 +17,32 @@ export const useFetchProfile = () => {
...
-    if (!nft && address) {
+    if (!nft && avatarFetchStatus !== ProfileAvatarFetchStatus.FETCHED && address) {
       dispatch(fetchProfileAvatar(account))
     }
 
-    if (!username && nft && address) {
-      dispatch(fetchProfileUsername(account))
+    if (
+      !username &&
+      avatarFetchStatus === ProfileAvatarFetchStatus.FETCHED &&
```
Chef-Cheems, line 39: this check is here to prevent fetching username before we figured out if they're registered or not (checked during avatar fetching)

src/state/profile/index.tsx
```diff
@@ -12,21 +12,24 @@ const initialState: ProfileState = {
...
-    const username = await getUsername(account)
-    return { account, username }
-  },
-)
+export const fetchProfileUsername = createAsyncThunk<
+  { account: string; username: string },
+  { account: string; hasRegistered: boolean }
+>('profile/fetchProfileUsername', async ({ account, hasRegistered }) => {
+  if (!hasRegistered) {
+    return { account, username: '' }
```
Chef-Cheems, line 28: Return empty username and don't make requests for username. Needed to set the fetching flag to FETCHED

src/state/profile/index.tsx
```diff
@@ -56,28 +59,120 @@ export const profileSlice = createSlice({
...
+        }
+      } else {
+        state.profileAvatars[account] = {
+          hasRegistered: false,
+          username: null,
+          nft: null,
+          // I think in theory this else should never be reached since we only check for username after we checked for profile/avatar
+          // just in case I set isFetchingAvatar will be ProfileAvatarFetchStatus.FETCHED at this point to avoid refetching
+          usernameFetchStatus: ProfileAvatarFetchStatus.FETCHING,
+          avatarFetchStatus: ProfileAvatarFetchStatus.FETCHED,
```
ChefHutch, line 77: How does this username thunk being `pending` relate to `avatarFetchStatus` being `FETCHED`?

This point goes throughout these thunks - it seems we're consistently overwirting the loading state for the other fetch status within these `else` conditions.

src/state/profile/hooks.ts
```diff
@@ -18,47 +17,32 @@ export const useFetchProfile = () => {
   }, [account, dispatch])
 }
 
-export const useProfileForAddress = (address: string) => {
```
ChefHutch, line 21: IN A MASSIVE PLOT TWIST I WOLUD LIKE TO KEEP THIS HOOK.

I want to fetch Profile information (avatar, username, achievements, points, team) for non-connected wallets.

This is just a nice wrapper around the `getProfile` function

hachiojidev (approved):

ChefHutch (approved):
I still think the reducer fetch status changes look a bit bonkers but :yeet:

PR #2385, opened 2021-09-28
fix(nft market): Update swipper index on swipe

0xCorgi:


src/views/Nft/market/Collection/IndividualNFTPage/MoreFromThisCollection.tsx
```diff
@@ -90,6 +90,10 @@ const MoreFromThisCollection: React.FC<MoreFromThisCollectionProps> = ({ collect
     swiperRef.slideTo(index)
   }
 
+  const updateActiveIndex = () => {
+    setActiveIndex(Math.ceil(swiperRef.activeIndex / slidesPerView))
+  }
```
hachiojidev, line 95: Did you check the parameters passed in? Seems unlikely that a `onActiveIndexChange` handler wouldn't pass in the old and new.

0xCorgi: You were right, this fixed.

PR #2386, opened 2021-09-28
fix: Scroll top behaviour

0xCorgi:
A proposition to use `#` for same page navigation.
It relies on a new component `HashRoute` that just wrap `Route` from react-router and check a `hash` prop to render its component.
This would allow us to keep the existing behavior while landing on a page and ignore it afterward while using internal tabs navigation.


hachiojidev (approved):
I am not sure this is the right solution but we will roll with it.

ChefHutch (approved):

PR #2388, opened 2021-09-28
Fix: Disable view all button on hot collections

0xCorgi:


hachiojidev (changes requested):
I think we just need to remove this until we get the squad collection done.

RabbitDoge:
This code will be updated when the squad goes live so I don't think we need to merge this

PR #2389, opened 2021-09-28
feat: Add rarity sort

0xCorgi:
Add a sort button on rarity column.


hachiojidev (approved):

adrien-chassinat:
Hello, why the api doesn't take in consideration the burn count for the attributesDistribution  ? attributesDistribution return the total supply not the current supply. Example on gen0: 

Api returning:
"0":175,
"1":117,
"2":80,
"3":77,
"4":151

PR #2391, opened 2021-09-28
fix: (info): Interval leakage in top token movers

memoyil:
Interval inside top token movers is never cleared after detaching. I also added a feature to scroll back to first when it reaches to last. 

To review:

https://deploy-preview-2391--pancakeswap-dev.netlify.app/

### DIFF
--- a/src/views/Info/components/TopTokenMovers/index.tsx
+++ b/src/views/Info/components/TopTokenMovers/index.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useRef, useState, useEffect } from 'react'
+import React, { useMemo, useRef, useEffect } from 'react'
 import styled from 'styled-components'
 import { Text, Flex, Box, Card } from '@pancakeswap/uikit'
 import { Link } from 'react-router-dom'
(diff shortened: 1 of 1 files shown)


src/views/Info/components/TopTokenMovers/index.tsx
```diff
@@ -73,20 +73,30 @@ const TopTokenMovers: React.FC = () => {
   }, [allTokens])
 
   const increaseRef = useRef<HTMLDivElement>(null)
-  const [increaseSet, setIncreaseSet] = useState(false)
+  const moveLeftRef = useRef<boolean>(true)
+
   // const [pauseAnimation, setPauseAnimation] = useState(false)
```
RabbitDoge, line 78: remove commented comment plz, it shouldn't be there

Chef-Cheems (approved):

src/views/Info/components/TopTokenMovers/index.tsx
Chef-Cheems: yeah, PR overall good, just maybe remove those. My fault for keeping them there though 😬

hachiojidev (approved):

src/views/Info/components/TopTokenMovers/index.tsx
memoyil: Done, https://github.com/pancakeswap/pancake-frontend/pull/2398

PR #2399, opened 2021-09-29
feat: Add price sort on nft sale

0xCorgi:
image


Chef-Cheems (approved):
Some of it probably gonna change soon but overall nice, thanks!

PR #2418, opened 2021-09-30
fix(pancake squad): Add live status on nft entry

0xCorgi:


src/components/Menu/config/config.ts
```diff
@@ -59,6 +59,10 @@ const config: (t: ContextApi['t']) => ConfigMenuItemsType[] = (t) => [
       {
         label: t('Overview'),
         href: `${nftsBaseUrl}`,
+        status: {
+          text: 'LIVE',
```
memoyil, line 63: It would be better if we have a translation on that

Chef-Cheems (approved):

PR #2429, opened 2021-10-01
chore(profile-creation): Remove all reliance on pb consts 

ChefHutch:
`src/config/constants/nfts/pancakeBunnies.ts` ded 🥳


src/views/ProfileCreation/Mint.tsx
```diff
@@ -37,6 +40,22 @@ const Mint: React.FC = () => {
...
   const { callWithGasPrice } = useCallWithGasPrice()
 
+  useEffect(() => {
+    const getStarterNfts = async () => {
+      const { data: allPbTokens } = await getNftsFromCollectionApi(pancakeBunniesAddress)
+      const nfts = STARTER_NFT_BUNNY_IDS.map((bunnyId) => {
+        if (allPbTokens && allPbTokens[bunnyId]) {
+          return { ...allPbTokens[bunnyId], bunnyId }
+        }
+        return undefined
```
ChefHutch, line 50: @0xCorgi wdyt?

I want to prevent an app crash if the the fetch fails - which would happen if you attempted something like 

```
// data = null
data['5']
```

This solution seems a little inelegant though

0xCorgi: Not sure it's that much better but we could do that:

```suggestion
     STARTER_NFT_BUNNY_IDS
        .filter(bunnyId => !allPbTokens?[bunnyId])
        .map(bunnyId => { ...allPbTokens[bunnyId], bunnyId })
```

ChefHutch: Akh yea cheers. Let's leave it as is.

hachiojidev (approved):

PR #2432, opened 2021-10-01
chore(lowest-price): Add check for PB to lowest price market hooks

ChefHutch:
Add `isPancakeBunny` checks to `useGetLowestPriceFromNft` hook to prevent crashes for non-pb NFTs


src/views/Nft/market/hooks/useGetLowestPrice.ts
```diff
@@ -0,0 +1,73 @@
...
+}
+
+export const useGetLowestPriceFromNft = (nft: NftToken): LowestNftPrice => {
+  const [isFetching, setIsFetching] = useState<boolean>(false)
+  const [lowestPrice, setLowestPrice] = useState<number>(null)
+  const isPancakeBunny = nft.collectionAddress?.toLowerCase() === pancakeBunniesAddress.toLowerCase()
+
+  useEffect(() => {
+    const fetchLowestPrice = async () => {
+      const bunnyIdAttr = getBunnyIdFromNft(nft)
```
Chef-Cheems, line 48: Maybe
- Take `const bunnyIdAttr = getBunnyIdFromNft(nft)` can be taken out of `useEffect` 
- Call `useGetLowestPriceFromBunnyId(bunnyIdAttr)`
- Remove useStates (as they are in the other hook already)

i.e.
```
export const useGetLowestPriceFromNft = (nft: NftToken): LowestNftPrice => {
   const isPancakeBunny = nft.collectionAddress?.toLowerCase() === pancakeBunniesAddress.toLowerCase()
   const bunnyIdAttr = getBunnyIdFromNft(nft) ?? null
   const { isFetching, lowestPrice } = useGetLowestPriceFromBunnyId(bunnyIdAttr)
   if (isPancakeBunny) {
    return { isFetching, lowestPrice }
   }
  return 0 // to be done later for other collections
}
```

src/views/Nft/market/hooks/useGetLowestPrice.ts
```diff
@@ -0,0 +1,73 @@
...
+import { NftToken } from 'state/nftMarket/types'
+import { pancakeBunniesAddress } from '../constants'
+
+export interface LowestNftPrice {
+  isFetching: boolean
+  lowestPrice: number
+}
+
+export const getBunnyIdFromNft = (nft: NftToken): string =>
+  nft.attributes?.find((attr) => attr.traitType === 'bunnyId').value.toString()
```
Chef-Cheems, line 12: Can you add guards here for nft without bunny id?
```js
const bunnyId = nft.attributes?.find((attr) => attr.traitType === 'bunnyId')?.value
return bunnyId ? bunnyId.toString() : null
```

Just realized this would fail for other collections

src/views/Nft/market/hooks/useGetLowestPrice.ts
Chef-Cheems: agh, for other collectibles it should `return { isFetching: false, lowestPrice: nullOrZero }` to be compatible with some components. Its not gonna hit anything yet though, but as placeholder it better replicate what we gonna have in the future.

src/views/Nft/market/hooks/useGetLowestPrice.ts
ChefHutch: Sure - makes sense to add those catches.

This shouldn't be used for collections without a `bunnyID` traitType.

Chef-Cheems: yeah but if you'd refactor the hook for my suggestion about it might be that other collection calls this function. Result will be ignored in this case though. :)

src/views/Nft/market/hooks/useGetLowestPrice.ts
ChefHutch: Yea I see what you mean - I'd rather not couple the hooks at this point though. This logic may well change - particularly as we develop the lowest priced traits UI & subgraph queries.

In the immediate term though, we need a way to load the `CollectibleCard`s without the app crashing.

Chef-Cheems: No problem 👍 
Yeah it probably will change a lot

Chef-Cheems (approved):

src/views/Nft/market/hooks/useGetLowestPrice.ts
ChefHutch: Yea sorry I didn't realise these suggestions came as a package! Makes total sense in the context of the other.

I liked this catch regardless tho so added it 😉

PR #2434, opened 2021-10-02
refactor: Rewrote views to use router hooks instead of props

zoltan31:
This is a small refactor on how views use router functions.
Now views inside App.tsx are using the router hooks instead of router props.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -134,21 +134,48 @@ const App: React.FC = () => {
               <Info />
             </Route>
 
-            {/* Using this format because these components use routes injected props. We need to rework them with hooks */}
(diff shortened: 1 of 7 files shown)


memoyil:
What is the advantage of it?

zoltan31:
> What is the advantage of it?

It's a code refactor. There was a comment which asked for this rewrite. The codebase is now more united because every view uses hooks instead of props and hooks combined.

Chef-Cheems:
Sorry for ignoring this for long time... but yea, we're [moving to NextJS](https://github.com/pancakeswap/pancake-frontend/pull/2978) very soon, so this is not necessary and will make more conflicts if we merge 😅

PR #2438, opened 2021-10-04
Feature/nft filters

hachiojidev:
image


src/views/Nft/market/Collection/config.ts
```diff
@@ -0,0 +1 @@
+export const REQUEST_SIZE = 100
```
ChefHutch, line 1: If we _could_ request 1k 'traits' at once (and store all the data, but only render 100 at a time in the FE) - then we'd be able to add price/recently listed sorting to the `FilteredCollectionNfts`. 

Something to consider for later.

ChefHutch (approved):

RabbitDoge (approved):

PR #2441, opened 2021-10-04
fix: Fetch data for non PB collections

RabbitDoge:
[ ] Explain what your PR does

### DIFF
--- a/src/state/nftMarket/reducer.ts
+++ b/src/state/nftMarket/reducer.ts
@@ -25,6 +25,8 @@ import {
   NftToken,
   NftLocation,
   ApiSingleTokenData,
+  NftAttribute,
+  TokenMarketData,
 } from './types'
 
 const initialState: State = {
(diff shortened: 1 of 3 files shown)


src/state/nftMarket/reducer.ts
```diff
@@ -123,8 +124,21 @@ export const fetchNftsFromCollections = createAsyncThunk(
         })
       }
 
-      // TODO: revisit this for other collecitons
-      return []
+      return Object.keys(nfts.data).map((id) => {
+        const apiMetadata = nfts.data[id]
+        const marketData = nftsMarket[id] || {}
```
Chef-Cheems, line 129: this should probably `nftMarket.find((nft) => nft.tokenId === id)`. Cause right now it relies on the fact that array returned from subgraph is ordered from 0 to 10000, which is not the case if some are not on sale and/or subgraph orders them not by tokenId

RabbitDoge: Smart

hachiojidev (approved):

PR #2446, opened 2021-10-05
feat: Auto update refinement

Chef-Cheems:
- Adds update for existing NFTs in the state every 10 seconds (in case they get sold or price gets modified)
- Adds update indicator to the table
- Removes pre-fetching random pancake bunny NFTs into state as they are not used anywhere.

### DIFF
--- a/src/state/nftMarket/hooks.ts
+++ b/src/state/nftMarket/hooks.ts
@@ -3,8 +3,7 @@ import { useSelector } from 'react-redux'
 import { useAppDispatch } from 'state'
 import { pancakeBunniesAddress } from 'views/Nft/market/constants'
 import { isAddress } from 'utils'
-import useRefresh from 'hooks/useRefresh'
-import { fetchCollection, fetchCollections, fetchNftsByBunnyId, updateNftTokensData } from './reducer'
+import { fetchCollection, fetchCollections, fetchNewPBAndUpdateExisting } from './reducer'
 import { State } from '../types'
 import { NftToken, UserNftsState } from './types'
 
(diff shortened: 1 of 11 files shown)


src/state/nftMarket/hooks.ts
```diff
@@ -22,18 +21,23 @@ export const useFetchCollection = (collectionAddress: string) => {
   }, [dispatch, collectionAddress])
 }
 
-// Returns a function that fetches more NFTs when called and puts them into redux state.
-// Also returns loading flag and time of latest successful fetch
-export const useFetchByBunnyId = (bunnyId: string) => {
+// Returns a function that fetches more NFTs for specified bunny id
+// as well as updating existing PB NFTs in state
+// Note: PancakeBunny specific
+export const useFetchByBunnyIdAndUpdate = (bunnyId: string) => {
```
Chef-Cheems, line 27: Originally I rewrote this to account for other collections too, but the logic became too complex and resulted in more queries (even if user just sits on Dollop page it would update Squads, etc).
Its better to make a more maintainable hook for other collections, PB are quite different in terms of updating since they have multiple instances.

src/state/nftMarket/reducer.ts
```diff
@@ -148,112 +121,62 @@ export const fetchNftsFromCollections = createAsyncThunk(
 )
 
 /**
- * Fetch fresh marketdata for existing tokens in the store
+ * This action keeps data on the individual PancakeBunny page up-to-date. Operation is a twofold
+ * 1. Update existing NFTs in the state in case some were sold or got price modified
+ * 2. Fetch 30 more NFTs with specified bunny id
  */
-export const updateNftTokensData = createAsyncThunk<
+export const fetchNewPBAndUpdateExisting = createAsyncThunk<
```
Chef-Cheems, line 128: Getting new NFTs and updating old ones are combined into 1 hook to not have a racing game between two update mechanisms.

hachiojidev (approved):

PR #2447, opened 2021-10-05
feat: Farm Auction five

Chef-Cheems:


RabbitDoge (approved):

ChefHutch (approved):

src/views/Home/components/Banners/FarmAuctionsBanner.tsx
```diff
@@ -88,7 +88,7 @@ const FarmAuctionsBanner = () => {
     <Wrapper>
       <Inner>
         <LeftWrapper>
-          <StyledSubheading>{t('%num% Contenders...', { num: 39 })}</StyledSubheading>
+          <StyledSubheading>{t('%num% Contenders...', { num: 46 })}</StyledSubheading>
```
ChefKai, line 91: Can we hook the contract data to pull total of bidders?

Chef-Cheems: @ChefKai we can, but we'd better at the same time modify the whole modal thing to automatically pop up when there is live auction. Not now for sure, but at some point yeah 👍

PR #2463, opened 2021-10-06
fix: (pancakesquad): Fix typos and make translations test compatible

memoyil:


src/views/PancakeSquad/components/Header/SaleProgress.tsx
```diff
@@ -12,13 +12,28 @@ type PreEventProps = {
...
+  const data = {
+    remaining,
+    total,
+  }
+
+  switch (saleStatus) {
+    case SaleStatusEnum.Pending:
+      return t('', data)
+    case SaleStatusEnum.Premint:
+      return t('', data)
```
hachiojidev, line 25: No need to wrap this in a translation function.

hachiojidev (changes requested):

src/views/PancakeSquad/components/Header/SaleProgress.tsx
memoyil: Done

PR #2515, opened 2021-10-13
feat: Improve filters

Chef-Cheems:


src/components/ScrollToTopButton/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  const scrollToTop = () => {
+    window.scrollTo({
+      top: 400,
+      behavior: 'auto',
+    })
+  }
+
+  window.addEventListener('scroll', toggleVisible)
```
memoyil, line 30: Listener should be removed when unmount

PR #2517, opened 2021-10-13
fix: (profile): Claimable achievements throws error when fetching

memoyil:
To review:

https://deploy-preview-2517--pancakeswap-dev.netlify.app/

To reproduce:

1. Go to profile page
2. Connect an account that has claimable achievements
3. See the error on console when fetching claimable achievements

Uncaught (in promise) Error: invalid address (argument="address", value=5, code=INVALID_ARGUMENT, version=address/5.4.0) (argument=null, value=5, code=INVALID_ARGUMENT, version=abi/5.4.0)

### DIFF
--- a/src/utils/achievements.ts
+++ b/src/utils/achievements.ts
@@ -71,7 +71,7 @@ export const getClaimableIfoData = async (account: string): Promise<Achievement[
       const [claimStatus] = claimStatusArr
 
       if (claimStatus === true) {
-        return [...accum, { address: getPointCenterIfoAddress(), name: 'ifos', params: [index] }]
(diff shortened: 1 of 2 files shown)


src/utils/achievements.ts
```diff
@@ -80,18 +80,19 @@ export const getClaimableIfoData = async (account: string): Promise<Achievement[
 
   // Transform response to an Achievement
   return claimableIfoData.reduce((accum, claimableIfoDataItem) => {
-    if (!campaignMap.has(claimableIfoDataItem.campaignId)) {
+    const claimableCampaignId = claimableIfoDataItem.campaignId.toString()
```
memoyil, line 83: campaign id is bignumber that needs to be converted to plain string

PR #2532, opened 2021-10-15
fix(router): Fix add new token address

lioncho:
[x] Before opening a pull request, please read the [contributing guidelines](https://github.com/pancakeswap/pancake-frontend/blob/master/CONTRIBUTING.md) first
[x] If your PR is work in progress, open it as `draft`
[x] Before requesting a review, all the checks need to pass
[x] Explain what your PR does

When I clone to my local, "add new token address" is not working. 
It was caused by the router_address is wrong. I replaced to current pancakeswap router then now it works correctly like the release version.

### DIFF
--- a/src/state/swap/hooks.ts
+++ b/src/state/swap/hooks.ts
@@ -88,8 +88,8 @@ export function tryParseAmount(value?: string, currency?: Currency): CurrencyAmo
 
 const BAD_RECIPIENT_ADDRESSES: string[] = [
   '0x5C69bEe701ef814a2B6a3EDD4B1652CB9cc5aA6f', // v2 factory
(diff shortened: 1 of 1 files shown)


Chef-Cheems:
@ryancho hi! What do you mean by `When I clone to my local, "add new token address" is not working.`. The code you've changed seems to affect only the transfer to a another recipient 🤔

Chef-Cheems:
`BAD_RECIPIENT_ADDRESSES` is only used when using expert mode and setting different recipient, its not related in any way to token addresses.

The feature to add recipient will be removed soon (together with the code this PR changes), so I'm closing this PR.

PR #2576, opened 2021-10-22
feat(chart): Add charts on swap page

0xCorgi:


src/views/Swap/components/Chart/utils.ts
```diff
@@ -0,0 +1,34 @@
+import { BNB_ADDRESS } from './constants'
+
+const MIN_VALUE_DISPLAYED = 0.001
+
+export const getTimeWindowChange = (lineChartData) => {
+  if (lineChartData.length > 0) {
+    const firstValue = lineChartData.find(({ value }) => !!value && value > 0)?.value
```
chef-jojo, line 7:
```suggestion
    const firstValue = lineChartData.find(({ value }) => !!value && value > 0)?.value ?? 0
```

chef-jojo (approved):
kudos

PR #2627, opened 2021-11-02
feat(ifo): Show estimated token price with fee when live

chef-bun:
Show estimated token price with participation fee (real cost) when an IFO is live

It uses `sumTaxesOverflow`, `raisingAmountPool`, `offeringAmountPool` from `poolCharacteristic` (SC) to work out an ratio between the "cost with fee" and "bare cost". Then use that ratio on `tokenOfferingPrice` to get the cost in USD

It should better help users to make decisions before joining unlimited sales


src/views/Ifos/components/IfoFoldableCard/IfoPoolCard/IfoCardDetails.tsx
```diff
@@ -51,6 +51,16 @@ const IfoCardDetails: React.FC<IfoCardDetailsProps> = ({ poolId, ifo, publicIfoD
   const totalLPCommittedInUSD = currencyPriceInUSD.times(totalLPCommitted)
   const totalCommitted = `~$${formatNumber(totalLPCommittedInUSD.toNumber(), 0, 0)} (${totalCommittedPercent}%)`
 
+  const PricePerTokenWithFeeToOriginalRaio = poolCharacteristic.sumTaxesOverflow
```
Chef-Cheems, line 54: variable names should start with lower case letter
same for PricePerTokenWithFee

Chef-Cheems:
Change implemented in #2637

PR #2657, opened 2021-11-09
refactor: (lottery): Remove unused and broken reducer

memoyil:
@Chef-Cheems That reducer is not used anywhere and it could not set the state since it doesn't return the new one. Should we remove it or keep it fixed?

To review:

https://deploy-preview-2657--pancakeswap-dev.netlify.app/

### DIFF
--- a/src/state/lottery/index.ts
+++ b/src/state/lottery/index.ts
@@ -103,11 +103,7 @@ export const setLotteryIsTransitioning = createAsyncThunk<{ isTransitioning: boo
 export const LotterySlice = createSlice({
   name: 'Lottery',
   initialState,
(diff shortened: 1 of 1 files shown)


src/state/lottery/index.ts
```diff
@@ -106,6 +106,7 @@ export const LotterySlice = createSlice({
   reducers: {
     setLotteryPublicData: (state, action) => {
       state = action.payload
+      return state
```
memoyil, line 109: We have to return the state if we change it completely

Chef-Cheems (approved):
yep, seems like a dead code, let's remove 👍

PR #2658, opened 2021-11-09
fix: (roi): Expected roi input validation and error message

memoyil:
To reproduce: 

1. Go to pools
2. Click apy calculator button any of the pools
3. Go to expected roi section
4. Change value to something else
5. See error message which shouldn't be there

image

6. Also Input value accepts non-digit values which shouldn't do

To review:

https://deploy-preview-2658--pancakeswap-dev.netlify.app/

### DIFF
--- a/src/components/RoiCalculatorModal/RoiCard.tsx
+++ b/src/components/RoiCalculatorModal/RoiCard.tsx
@@ -1,4 +1,4 @@
-import React, { useRef, useEffect } from 'react'
+import React, { useRef, useEffect, useState } from 'react'
 import styled from 'styled-components'
 import { Box, Flex, Text, Input, CheckmarkIcon, PencilIcon, IconButton } from '@pancakeswap/uikit'
 import { useTranslation } from 'contexts/Localization'
@@ -76,6 +76,7 @@ interface RoiCardProps {
 }
 
 const RoiCard: React.FC<RoiCardProps> = ({ earningTokenSymbol, calculatorState, setTargetRoi, setCalculatorMode }) => {
+  const [expectedRoi, setExpectedRoi] = useState('')
   const inputRef = useRef<HTMLInputElement | null>(null)
   const { roiUSD, roiTokens, roiPercentage } = calculatorState.data
   const { mode } = calculatorState.controls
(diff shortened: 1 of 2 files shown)


src/components/RoiCalculatorModal/useRoiCalculatorReducer.ts
```diff
@@ -273,7 +273,7 @@ const useRoiCalculatorReducer = (
     const targetRoiAsTokens = new BigNumber(amount).div(earningTokenPrice)
     dispatch({
       type: 'setTargetRoi',
-      payload: { roiUSD: amount, roiTokens: targetRoiAsTokens.isNaN() ? 0 : targetRoiAsTokens.toNumber() },
+      payload: { roiUSD: +amount, roiTokens: targetRoiAsTokens.isNaN() ? 0 : targetRoiAsTokens.toNumber() },
```
memoyil, line 276: roiUSD type is number in initial state so we have to convert it to number otherwise it will end up string in runtime.

src/components/RoiCalculatorModal/RoiCard.tsx
```diff
@@ -90,13 +91,18 @@ const RoiCard: React.FC<RoiCardProps> = ({ earningTokenSymbol, calculatorState,
 
   const onEnterEditing = () => {
     setCalculatorMode(CalculatorMode.PRINCIPAL_BASED_ON_ROI)
+    setExpectedRoi('')
```
chef-jojo, line 94: this will reset the input before enter editing, is this an expected behaviour?

memoyil: Yes, Imho when user clicks button to edit It should see empty input field that can put input easier than first removing the previous roi.

Chef-Cheems: @memoyil it clears expectedRoi but it leaves all other data in place (i.e. Token value for USD). Let's leave it as is maybe. Users might want to change "1034.33$" to "1000$" 🤷 
Screenshot 2021-11-18 at 9 56 54

memoyil: Done

Chef-Cheems (approved):

PR #2680, opened 2021-11-16
fix: (ifo): Skeleton in ifo card not disappear when no account is connected

memoyil:
To reproduce:

1. Open incognito browser with no account is connected at first
2. Go to ifo
3. See ifo cards skeletons doesn't disappear even after data loaded

To review: 

https://deploy-preview-2680--pancakeswap-dev.netlify.app/

### DIFF
--- a/src/state/profile/index.tsx
+++ b/src/state/profile/index.tsx
@@ -5,7 +5,7 @@ import { getProfile, getProfileAvatar, getUsername } from './helpers'
 
 const initialState: ProfileState = {
   isInitialized: false,
-  isLoading: true,
+  isLoading: false,
   hasRegistered: false,
   data: null,
   profileAvatars: {},


src/state/profile/index.tsx
```diff
@@ -5,7 +5,7 @@ import { getProfile, getProfileAvatar, getUsername } from './helpers'
 
 const initialState: ProfileState = {
   isInitialized: false,
-  isLoading: true,
+  isLoading: false,
```
memoyil, line 8: It should be stay false until we call fetchProfile (which requires account) otherwise if there is no account connected It will stay loading true forever

Chef-Cheems, line 8: hm... I wonder if it breaks something else somewhere else 🤔

memoyil: It is used in 

1. Profile creation page 
2. User menu
3. IFO
4. Profile page
5. User detail banner in homepage

I see no issues so far

Chef-Cheems: tested locally - seems to not cause any issues.
I wonder why it was true in the first place. 😑

Chef-Cheems (approved):

PR #2700, opened 2021-11-18
refactor: (ifo): Percentage of total when total amount 0

memoyil:
When total amount is zero it falls back to big number 0 which calls toLocaleString with options object that bignumber's function doesn't have that parameters although it doesn't create problem on runtime, just to be on the safe side.

To review: 
https://deploy-preview-2700--pancakeswap-dev.netlify.app/


chef-jojo (approved):
👍

Chef-Cheems (approved):

PR #2703, opened 2021-11-19
fix: typo

0xdie:
- Fix Turkish typo


ChefNina:
I'll take care of this along with the translators team.

ChefNina:
Reviewed by native Turkish admins, the second and third change will be applied. 
The first points - Farm* will be remained as they are, as "Farm" is more clear to the DeFi users.
Thanks for your suggestion @0xdie !!

PR #2718, opened 2021-11-21
[Ru-Translation] Was very confusing

Serhioromano:
It sounded like when you trade in NFT for CAKE you lose access to CAKE forever. At leas it is very unclear from the translation, to what asset you are going to lose access CAKE or NFT.


Chef-Cheems (approved):
This string seems to be unused anymore, but thanks anyway :)

I'll try to make a script that removes unused translations, there's a bunch of them I presume...

PR #2732, opened 2021-11-22
fix: User state still visible when wallet disconnects with Auto-Lock or WalletConnect

memoyil:
To reproduce:

With injected wallet

1. Go to your wallet settings (i.e Metamask) set Auto - Lock timer (Settings -> Advanced)
2. Connect to website
3. Wait idle
4. Check redux state from browser that user profile and nft state still visible

With walletconnect

1. Connect to website via WalletConnect
2. Disconnect from the device that you use to scan qr code
3. Check redux state from browser that user profile and nft state still visible

To review:

https://deploy-preview-2732--pancakeswap-dev.netlify.app/

### DIFF
--- a/src/hooks/useInactiveListener.ts
+++ b/src/hooks/useInactiveListener.ts
@@ -1,0 +1,38 @@
+import { useEffect } from 'react'
+import { useWeb3React } from '@web3-react/core'
+import { connectorLocalStorageKey } from '@pancakeswap/uikit'
+import { profileClear } from '../state/profile'
+import { resetUserNftState } from '../state/nftMarket/reducer'
+import { clearAllTransactions } from '../state/transactions/actions'
+import { useAppDispatch } from '../state'
+import { connectorsByName } from '../utils/web3React'
+
+export const useInactiveListener = () => {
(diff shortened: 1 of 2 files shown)


src/hooks/useInactiveListener.ts
```diff
@@ -0,0 +1,38 @@
...
+          connectorsByName.walletconnect.close()
+          connectorsByName.walletconnect.walletConnectProvider = null
+        }
+        window.localStorage.removeItem(connectorLocalStorageKey)
+        if (chainId) {
+          dispatch(clearAllTransactions({ chainId }))
+        }
+      }
+
+      connector.addListener('Web3ReactDeactivate', handleDeactivate)
```
memoyil, line 30: Web3ReactDeactivate event occurs only when wallet disconnected externally

chef-jojo:
MetaMask works pretty well from my side 🤔 
is this a fix only for WalletConnect?

memoyil:
> MetaMask works pretty well from my side 🤔 is this a fix only for WalletConnect?

After auto-lock time kicks in and disconnects the wallet externally (not via logout), I still see the user related info in redux storage.

PR #2739, opened 2021-11-22
fix: (nft): Collection list stays at old scroll position when page changes

memoyil:
To review: 

1. Go collections https://pancakeswap.finance/nfts/collections/
2. Go to second page by scrolling down first to reach pagination
3. Since it has less item than the first one, the item list is not visible you have to scroll up 

To review:

https://deploy-preview-2739--pancakeswap-dev.netlify.app/

### DIFF
--- a/src/views/Nft/market/Collections/index.tsx
+++ b/src/views/Nft/market/Collections/index.tsx
@@ -56,6 +56,16 @@ const Collectible = () => {
   const [maxPage, setMaxPage] = useState(1)
 
   useEffect(() => {
+    setTimeout(() => {
+      window.scroll({
+        top: 0,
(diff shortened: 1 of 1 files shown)


src/views/Nft/market/Collections/index.tsx
```diff
@@ -55,6 +55,16 @@ const Collectible = () => {
   const [page, setPage] = useState(1)
   const [maxPage, setMaxPage] = useState(1)
 
+  useEffect(() => {
+    setTimeout(() => {
```
memoyil, line 59: This timeout for firefox compatibility

Chef-Cheems (approved):

PR #2746, opened 2021-11-23
chore: Fix typos

memoyil:


src/state/nftMarket/types.ts
```diff
@@ -200,7 +200,7 @@ export interface ApiCollection {
 
 // Get all collections
 // ${API_NFT}/collections/
-export interface ApiCollectionsReponse {
+export interface ApiCollectionsResponse {
```
chef-jojo, line 203: Looks like this type never used
Let's reference it into https://github.com/pancakeswap/pancake-frontend/blob/develop/src/state/nftMarket/helpers.ts#L37

memoyil: Done

chef-jojo (approved):

Chef-Cheems (approved):

PR #2747, opened 2021-11-23
feat: (farm-auction): Add input validation to auction history archive

memoyil:
To review:

1. Go to farm auction
2. Go to archive
3. Put non numeric values to input
4. See it doesn't accept to values

https://deploy-preview-2747--pancakeswap-dev.netlify.app/

### DIFF
--- a/src/views/FarmAuction/components/AuctionHistory.tsx
+++ b/src/views/FarmAuction/components/AuctionHistory.tsx
@@ -75,20 +75,18 @@ const AuctionHistory: React.FC<AuctionHistoryProps> = ({ mostRecentClosedAuction
   const endDate = selectedAuction ? format(selectedAuction.auction.endDate, 'MMM. dd yyyy, hh:mm aa') : null
 
   const handleHistoryAuctionChange = (event: React.ChangeEvent<HTMLInputElement>) => {
-    const {
-      target: { value },
-    } = event
-    if (value) {
-      let newAuctionId = value
-      if (parseInt(value, 10) <= 0) {
-        newAuctionId = ''
-      }
-      if (parseInt(value, 10) >= mostRecentClosedAuctionId) {
-        newAuctionId = mostRecentClosedAuctionId.toString()
-      }
+    if (event.currentTarget.validity.valid) {
+      const {
+        target: { value },
+      } = event
+      const valueAsNumber = +value
+      const newAuctionId =
+        valueAsNumber >= mostRecentClosedAuctionId
+          ? mostRecentClosedAuctionId.toString()
+          : valueAsNumber <= 0
+          ? ''
+          : value
       setHistoryAuctionId(newAuctionId)
-    } else {
-      setHistoryAuctionId('')
     }
   }
 
@@ -111,7 +109,9 @@ const AuctionHistory: React.FC<AuctionHistoryProps> = ({ mostRecentClosedAuction
           <Box width="62px" mr={['4px', '16px']}>
             <Input
               disabled={!mostRecentClosedAuctionId}
-              type="input"
+              type="text"
+              inputMode="numeric"
+              pattern="^[0-9]+$"
               value={historyAuctionId}
               onChange={handleHistoryAuctionChange}
             />


src/views/FarmAuction/components/AuctionHistory.tsx
```diff
@@ -75,20 +75,22 @@ const AuctionHistory: React.FC<AuctionHistoryProps> = ({ mostRecentClosedAuction
...
-    } = event
-    if (value) {
-      let newAuctionId = value
-      if (parseInt(value, 10) <= 0) {
-        newAuctionId = ''
+    if (event.currentTarget.validity.valid) {
+      const {
+        target: { value },
+      } = event
+      if (value) {
```
chef-jojo, line 82: how about `valueAsNumber` here?

memoyil: Done

src/views/FarmAuction/components/AuctionHistory.tsx
```diff
@@ -112,6 +110,8 @@ const AuctionHistory: React.FC<AuctionHistoryProps> = ({ mostRecentClosedAuction
             <Input
               disabled={!mostRecentClosedAuctionId}
               type="input"
+              inputMode="numeric"
+              pattern="^[0-9]+$"
```
chef-jojo, line 114: we have use native input validation for max
```suggestion
              type="number"
              max={mostRecentClosedAuctionId}
              min={1}
```

src/views/FarmAuction/components/AuctionHistory.tsx
```diff
@@ -75,20 +75,18 @@ const AuctionHistory: React.FC<AuctionHistoryProps> = ({ mostRecentClosedAuction
...
-      if (parseInt(value, 10) >= mostRecentClosedAuctionId) {
-        newAuctionId = mostRecentClosedAuctionId.toString()
-      }
+    if (event.currentTarget.validity.valid) {
+      const {
+        target: { value },
+      } = event
+      const valueAsNumber = +value
+      const newAuctionId =
+        valueAsNumber >= mostRecentClosedAuctionId
```
chef-jojo, line 84: if we have `max` in input, no need for this line

also get the number type value by `event.target.valueAsNumber` should do the job

src/views/FarmAuction/components/AuctionHistory.tsx
chef-jojo: need to change `historyAuctionId` to type number in this way

memoyil: If type number is used, it doesn't show numpad in mobile browsers (on safari in particular), thats why i use type text with numeric input mode

src/views/FarmAuction/components/AuctionHistory.tsx
memoyil: I modified the code to use event.target.valueAsNumber. But since we need to use type=text for the reason that i explained above I think we should keep that line.

memoyil: I tested with valueAsNumber It always comes NaN. I reverted it back

src/views/FarmAuction/components/AuctionHistory.tsx
chef-jojo: It's first time I heard about this restriction 🤔 you can do your way

memoyil: When type="number" used

image

when inputMode="numeric" used 

image

memoyil:
@chef-jojo Can you re review it ? :)

src/views/FarmAuction/components/AuctionHistory.tsx
Chef-Cheems: @memoyil what if we use type number and inputMode numeric? ios still shows text stuff?
To be honest as long as it works well on mobile I don't really care if its text type 🤷

memoyil: You can't use both, it should be either type=text and numeric or type=number

chef-jojo (approved):
👍🏽

Chef-Cheems (approved):

PR #2759, opened 2021-11-24
[Ru-translation] few changes

Serhioromano:
Here are few changes. Those rae diferent but I made them in one PR because they are all obviouse.

1. Голосование по is a translation mistake and should be Голосование за
1. Always when we reffer to prize, word pool should be translated as фонд as in other places Prizes pot is translated as фонд. It might confuse if in one place it is translated as фонд and in the other place as пул.
3. Подлежит обявления not self explanatory and  Будет анонсиравано  even word to word translation is more correct.


Serhioromano:
@ChefNina Посмотрите этот ПР пожалуйста.

Chef-Cheems (approved):

PR #2764, opened 2021-11-25
fix(ifo): Commit token image

chef-jojo:
CleanShot 2021-11-25 at 11 44 35@2x

### DIFF
--- a/src/views/Ifos/components/IfoFoldableCard/IfoPoolCard/IfoCardTokens.tsx
+++ b/src/views/Ifos/components/IfoFoldableCard/IfoPoolCard/IfoCardTokens.tsx
@@ -15,6 +15,7 @@ import { useWeb3React } from '@web3-react/core'
 import { Token } from '@pancakeswap/sdk'
 import { Ifo, PoolIds } from 'config/constants/types'
 import tokens from 'config/constants/tokens'
+import { cakeBnbLpToken } from 'config/constants/ifo'
 import { PublicIfoData, WalletIfoData } from 'views/Ifos/types'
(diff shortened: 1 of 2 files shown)


src/views/Ifos/components/IfoFoldableCard/IfoPoolCard/IfoCardTokens.tsx
```diff
@@ -156,14 +160,14 @@ const IfoCardTokens: React.FC<IfoCardTokensProps> = ({
       )
     }
     if (publicIfoData.status === 'finished') {
-      return userPoolCharacteristics.amountTokenCommittedInLP.isEqualTo(0) ? (
+      return userPoolCharacteristics.amountTokenCommittedInLP.isEqualTo(1) ? (
```
memoyil, line 163: It should be 0 no ?

chef-jojo: ah good catch. it's just for debug lol 🙏

dinhnhatbang: Should we use constant for a value ? Just make the number meaningful

chef-jojo: This line is pretty self explain: *amount token committed in LP is equal to 0*
I will leave it for now, thanks for the suggestion

Chef-Cheems (approved):

PR #2791, opened 2021-11-29
fix: (swap-chart): Add timeout when graphql server is unresponsive

memoyil:
When graphql is unresponsive fetching operation keeps sending new request on each render, (possibly it is because heavy operation to keep all the function call data in memory) that makes browser hard to scroll or other things.

To review:
https://deploy-preview-2791--pancakeswap-dev.netlify.app/


Chef-Cheems (approved):
Great stuff, thanks!

PR #2795, opened 2021-11-29
fix: (ifo): Reduce amount of requests on both live and past ifos

memoyil:
To reproduce on past ifos:

1. Go to ifo
2. Go to past ifo
3. See amount of requests done that returns same result

To reproduce on live ifo:

Check the frequency of the requests. (See comment below) It is sufficent to have new request when we have a new block

To review:
https://deploy-preview-2795--pancakeswap-dev.netlify.app/

### DIFF
--- a/src/views/Ifos/hooks/v1/useGetPublicIfoData.ts
+++ b/src/views/Ifos/hooks/v1/useGetPublicIfoData.ts
@@ -58,6 +58,7 @@ const useGetPublicIfoData = (ifo: Ifo): PublicIfoData => {
         : ((currentBlock - releaseBlockNumber) / (startBlockNum - releaseBlockNumber)) * 100
 
     setState((prev) => ({
+      ...prev,
       status,
       blocksRemaining,
       secondsUntilStart: (startBlockNum - currentBlock) * BSC_BLOCK_TIME,
@@ -65,9 +66,6 @@ const useGetPublicIfoData = (ifo: Ifo): PublicIfoData => {
       secondsUntilEnd: blocksRemaining * BSC_BLOCK_TIME,
       startBlockNum,
       endBlockNum,
-      currencyPriceInUSD: null,
-      numberPoints: null,
-      thresholdPoints: undefined,
       [PoolIds.poolUnlimited]: {
         ...prev.poolUnlimited,
         raisingAmountPool: raisingAmount ? new BigNumber(raisingAmount[0].toString()) : BIG_ZERO,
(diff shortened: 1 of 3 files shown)


src/views/Ifos/hooks/v1/useGetPublicIfoData.ts
```diff
@@ -74,9 +73,43 @@ const useGetPublicIfoData = (ifo: Ifo): PublicIfoData => {
     }))
   }, [address, currentBlock, releaseBlockNumber])
 
+  const fetchInActiveIfoData = useCallback(async () => {
```
Chef-Cheems, line 76: @memoyil not entirely sure what's the point of this. It seems to make same amount of calls as `fetchIfoData`, the only difference is some blocks/progress calculations are omitted. I think it might be better to use `fetchIfoData` for readability sake. Unless I miss something 🤔

src/views/Ifos/hooks/v2/useGetPublicIfoData.ts
```diff
@@ -131,11 +130,45 @@ const useGetPublicIfoData = (ifo: Ifo): PublicIfoData => {
       endBlockNum,
       numberPoints: numberPoints ? numberPoints[0].toNumber() : 0,
     }))
-  }, [address, currentBlock, releaseBlockNumber])
+  }, [currentBlock, releaseBlockNumber, ifoCalls])
+
+  const fetchInActiveIfoData = useCallback(async () => {
```
Chef-Cheems, line 135: Same question as in v1, do we actually make less calls here? Seems like just a bit of saved performance with blocks/progress calculations. 🤔

src/views/Ifos/hooks/v1/useGetPublicIfoData.ts
```diff
@@ -74,9 +73,43 @@ const useGetPublicIfoData = (ifo: Ifo): PublicIfoData => {
...
+    }
+  }, [fetchIfoData, ifo])
+
   useEffect(() => {
-    fetchIfoData()
-  }, [fetchIfoData])
+    if (!ifo.isActive) {
+      fetchInActiveIfoData()
+    }
+  }, [fetchInActiveIfoData, ifo])
```
Chef-Cheems, line 112: Perhaps this can be 1 hook?
```js
if (ifo.isActive) {
  fetchIfoData()
} else {
  fetchInActiveIfoData()
}
```

src/views/Ifos/hooks/v1/useGetPublicIfoData.ts
Chef-Cheems: also nitpick -  `Inactive` instead of `InActive` 😄 
if we even need it 🤔

Chef-Cheems: I understand your point about `useBlocks` updating v2 IFOs.
I'm just thinking the fetchIfoData can be used for finished IFOs too. If you wanna improve performance with removing progress calculations maybe we could just put a check for that inside `fetchIfoData`. The `isActive` can be passed as an argument to it to make it easier on `useCallback` dep array

src/views/Ifos/hooks/v1/useGetPublicIfoData.ts
memoyil: But that means everytime fetchIfoData changes we will call fetchInActiveIfoData

src/views/Ifos/hooks/v2/useGetPublicIfoData.ts
memoyil: Yes, it only depends on ifoCalls which depends on address that means we will only call it once.

src/views/Ifos/hooks/v1/useGetPublicIfoData.ts
memoyil: > 

Done

memoyil:
@Chef-Cheems I will close this for now until i come up with better solution

PR #2796, opened 2021-11-30
perf(pool): Improve Pool performance

chef-jojo:
Reducing client side computation and UI main thread blocking when navigate to Pools.
Also cancelling batch fetch farms action if there's already.

https://deploy-preview-2796--pancakeswap-dev.netlify.app/

### DIFF
--- a/src/utils/providers.ts
+++ b/src/utils/providers.ts
@@ -1,3 +1,4 @@
+/* eslint-disable prefer-destructuring */
 import { ethers, logger } from 'ethers'
 import { Logger } from '@ethersproject/logger'
 import { Network } from '@ethersproject/networks'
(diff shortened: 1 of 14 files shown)


src/utils/providers.ts
```diff
@@ -3,6 +3,6 @@ import getRpcUrl from 'utils/getRpcUrl'
 
 const RPC_URL = getRpcUrl()
 
-export const simpleRpcProvider = new ethers.providers.StaticJsonRpcProvider(RPC_URL)
+export const simpleRpcProvider = new ethers.providers.JsonRpcBatchProvider(RPC_URL)
```
memoyil, line 6: If we use that it sends too much requests to get chain id from node which is known already and always the same.

memoyil: https://github.com/pancakeswap/pancake-frontend/pull/2571

chef-jojo: yea, I was trying this batch provider, everything seems fine apart from this.

PR #2826, opened 2021-12-03
fix: (contract): Use simple rpc provider when signer not needed

memoyil:


src/state/multicall/retry.ts
```diff
@@ -47,6 +47,7 @@ export function retry<T>(
         }
         break
       } catch (error) {
+        console.error(error)
```
Chef-Cheems, line 50: I think this might be a bit too much to console.error. Also probably not gonna be useful really :)

memoyil: Since we won't have wallet injection anymore, without this we will not see the error I think

Chef-Cheems (approved):

PR #2831, opened 2021-12-05
fix: (trading view): Remove pancakesymbol prefix from trading view

memoyil:
To reproduce

1. Go to swap
2. Select major pairs like BNB/ETH
3. See chart not shown

### DIFF
--- a/src/views/Swap/components/Chart/TradingViewChart.tsx
+++ b/src/views/Swap/components/Chart/TradingViewChart.tsx
@@ -37,8 +37,6 @@ const LoadingWrapper = styled.div<{ $isDark: boolean }>`
   }
 `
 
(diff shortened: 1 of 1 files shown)


Chef-Cheems:
@memoyil `PANCAKESWAP:` prefix is required to get data from our liquidity pools. If you remove it it would take data from Binance (or something else).
`BNB/ETH` is not displayed cause TradingView only tracks `ETH/BNB` in our pools, so you'll have to switch the chart (or use basic chart).

We can auto-switch the chart if we detect that certain pair doesn't have a chart. @chef-jojo will look at it next week.

memoyil:
@Chef-Cheems 👍

chef-jojo:
Fallback to non specific exchange chart is a good idea! I will try this later

PR #2835, opened 2021-12-05
refactor: Reduce chart rerenders

memoyil:
It stops rerendering the trading view chart on every price change

It also fixes trading chart not completely visible in normal mode

To review:
https://deploy-preview-2835--pancakeswap-dev.netlify.app/

### DIFF
--- a/src/views/Swap/index.tsx
+++ b/src/views/Swap/index.tsx
@@ -34,7 +34,6 @@ import ProgressSteps from './components/ProgressSteps'
 import { AppBody } from '../../components/App'
 import ConnectWalletButton from '../../components/ConnectWalletButton'
 
-import { INITIAL_ALLOWED_SLIPPAGE } from '../../config/constants'
 import useActiveWeb3React from '../../hooks/useActiveWeb3React'
 import { useCurrency, useAllTokens } from '../../hooks/Tokens'
 import { ApprovalState, useApproveCallbackFromTrade } from '../../hooks/useApproveCallback'
@@ -396,8 +395,8 @@ export default function Swap({ history }: RouteComponentProps) {
                   setIsChartDisplayed={setIsChartDisplayed}
                   isChartDisplayed={isChartDisplayed}
                 />
-                <Wrapper id="swap-page">
-                  <AutoColumn gap="md">
+                <Wrapper id="swap-page" style={{ minHeight: '412px' }}>
+                  <AutoColumn gap="sm">
                     <CurrencyInputPanel
                       label={
                         independentField === Field.OUTPUT && !showWrap && trade ? t('From (estimated)') : t('From')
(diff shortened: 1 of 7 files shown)


chef-jojo:
CleanShot 2021-12-06 at 10 24 25@2x
height is different 🤔

memoyil:
> CleanShot 2021-12-06 at 10 24 25@2x
> 
> height is different 🤔

Yes, I increase the height a bit to make trading view dashboard bar on left fully visible

src/views/Swap/index.tsx
```diff
@@ -377,8 +376,8 @@ export default function Swap({ history }: RouteComponentProps) {
                   setIsChartDisplayed={setIsChartDisplayed}
                   isChartDisplayed={isChartDisplayed}
                 />
-                <Wrapper id="swap-page">
-                  <AutoColumn gap="md">
+                <Wrapper id="swap-page" style={{ minHeight: '412px' }}>
```
memoyil, line 379: minHeight is only for `Insufficient liquidity for this trade.Try enabling multi-hop trades.` error. Other that this it behaves like fixed height

chef-jojo (approved):

chef-jojo:
@memoyil Can you explain to me what's re-rendering you fix on 5e05521?

memoyil:
> @memoyil Can you explain to me what's re-rendering you fix on [5e05521](https://github.com/pancakeswap/pancake-frontend/commit/5e0552137839495a894d0e7c946f8c79d3795e84)?

To use Twchart symbol outside of the tradingview component and put it in a same Flex like it is now in develop (both of them are inside the tw component currently and it cause re-render everytime price changes).

chef-jojo:
I tried with this preview link, looks like it doesn't work 🤔, besides I don't think it's not worth to work around for TVchart symbol in this case. it's just a link with label, rerender doesn't cost much of resources 
https://user-images.githubusercontent.com/94336009/146734183-97680d24-b9b3-48ed-88b7-3170efb73601.mp4

memoyil:
> I tried with this preview link, looks like it doesn't work 🤔, besides I don't think it's worth to work around for TVchart symbol in this case. it's just a link with label, rerender doesn't cost much of resources https://user-images.githubusercontent.com/94336009/146734183-97680d24-b9b3-48ed-88b7-3170efb73601.mp4

No, It is not about tv chart symbol actually, It is about price tag (but we need to take both out for visual). Currently when the price changes It rerenders the whole trading view component no?

chef-jojo:
I see you what you mean. But the rerender still happening like 6s when refresh from context. I don't see any benefits of it or am I missed anything?

memoyil:
> I see you what you mean. But the rerender still happening like 6s when refresh from context. I don't see any benefits of it or am I missed anything?

Yes, rerender still happens on the swap page (I think on every new block thats why you see it in every 6 seconds) but TW Chart now is a memoized component. I actually tested by putting console.log in twchart component and it doesn't print it when price changes.

I think think the video shows rerendering the outer component not the twchart right?

chef-jojo (approved):
maybe I cloned the wrong branch 👀

Chef-Cheems (approved):

PR #2842, opened 2021-12-07
refactor: (ifo): Make ifo currency token generic

memoyil:
We can use both lp and other tokens for ifo

To review:
https://deploy-preview-2842--pancakeswap-dev.netlify.app/


memoyil:
Close to adapt ifo v3

PR #2844, opened 2021-12-07
refactor: Use contract

memoyil:


src/utils/index.ts
```diff
@@ -82,20 +83,17 @@ export function getProviderOrSigner(library: Web3Provider, account?: string): We
 }
 
 // account is optional
-export function getContract(address: string, ABI: any, library?: Web3Provider, account?: string): Contract {
-  if (!library) {
```
memoyil, line 86: Currently if address is wrong we return contract

Chef-Cheems (approved):

PR #2847, opened 2021-12-08
fix: First Navigation to Swap and Pools page makes app frozen and page loader not shown

memoyil:
To reproduce:

1. Go to webpage (in mobile preferred)
2. Click pools
3. See app is frozen and there is no loading indicator (if there is a poor connection waiting time might be a lot)

To review:
https://deploy-preview-2847--pancakeswap-dev.netlify.app/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -19,9 +19,6 @@ import PageLoader from './components/Loader/PageLoader'
 import EasterEgg from './components/EasterEgg'
 import GlobalCheckClaimStatus from './components/GlobalCheckClaimStatus'
 import history from './routerHistory'
-// Views included in the main bundle
-import Pools from './views/Pools'
-import Swap from './views/Swap'
 import {
   RedirectDuplicateTokenIds,
   RedirectOldAddLiquidityPathStructure,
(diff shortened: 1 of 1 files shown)


chef-jojo:
@memoyil  I tried this one before. 
This is only works for the first time. Because Suspense handle the loading with lazy load. After the route based js is loaded. It still slow as hell.
I believe the main issue is the redux dispatch here that blocks the UI thread
https://github.com/pancakeswap/pancake-frontend/pull/2796/files#diff-cb0b887b607d19b6acaf612014693c6adc881309e973af2433492aa8426f1f4fR65-R67

memoyil:
> @memoyil I tried this one before. It only works for the first time. Because Suspense handle the loading with lazy load. After the route based js is loaded. It will still frozen when navigate to pool or farm. I believe the main issue is the redux dispatch here somehow blocks the UI thread https://github.com/pancakeswap/pancake-frontend/pull/2796/files#diff-cb0b887b607d19b6acaf612014693c6adc881309e973af2433492aa8426f1f4fR65-R67

True, I agree but we should show loading indicator for the first navigation to users imho. I tested your branch, at first navigation app still freezes a bit but after that it works smoothly. So this pr will fix the first navigation.

Chef-Cheems:
Let's close this as migration to NextJS will probably solve all this

PR #2851, opened 2021-12-09
perf: (farms): Refactor public farm data fetch

memoyil:
It will reduce the first fetch around 9 seconds (on good connections) to 1 second. (This might benefit more if user has poor connection)

It will reduce the network calls by %99 percent.

Instead of multicalls per farm, this will make two multicall for all farms to fetch.

It will also benefit the apr calculations on pools page too.

To review:

https://deploy-preview-2851--pancakeswap-dev.netlify.app/


chef-jojo (commented):
I think overall is good. It reduces a lot network request from ~1400 to ~400 👀
nit:
1. `multicallv2` will always return same result length, only difference is the result may be `null` if not success. so checking the length result is not needed, we can check the result contains `null` or just let it be `null` and handle properly on `fetchFarms`
2. not sure if this related, sometimes apy is not showing at first.  `fetchPoolsPublicDataAsync` depends on farm data
```
// fetchPoolsPublicDataAsync
// get directly from state, the farm data is not ready yet
const prices = getTokenPricesFromFarm(getState().farms.data)
```

memoyil:
> I think overall is good. It reduces a lot network request from ~1400 to ~400 👀 nit:
> 
>     1. `multicallv2` will always return same result length, only difference is the result may be `null` if not success. so checking the length result is not needed, we can check the result contains `null` or just let it be `null` and handle properly on `fetchFarms`

**With requireSuccess (default true) option in `multicallv2`  it fails **
> 
>     2. not sure if this related, sometimes apy is not showing at first.  `fetchPoolsPublicDataAsync` depends on farm data
> 
> 
> ```
> // fetchPoolsPublicDataAsync
> // get directly from state, the farm data is not ready yet
> const prices = getTokenPricesFromFarm(getState().farms.data)
> ```



> I think overall is good. It reduces a lot network request from ~1400 to ~400 👀 nit:
> 
>     1. `multicallv2` will always return same result length, only difference is the result may be `null` if not success. so checking the length result is not needed, we can check the result contains `null` or just let it be `null` and handle properly on `fetchFarms`
> 
>     2. not sure if this related, sometimes apy is not showing at first.  `fetchPoolsPublicDataAsync` depends on farm data
> 
> 
> ```
> // fetchPoolsPublicDataAsync
> // get directly from state, the farm data is not ready yet
> const prices = getTokenPricesFromFarm(getState().farms.data)
> ```

1. Removed the size checks, since fetchFarms is null resistant there will be no problem.
2. I saw the same issue, I think with the new approach with less blocking calls fetchPoolsPublicDataAsync runs before farms data available, therefore user has to wait the next cycle of it (which uses slowRefresh 60sec) to get complete farm data. So i created bugfix on different pr (https://github.com/pancakeswap/pancake-frontend/pull/2863), but I pushed it also here now for you to check.

src/state/farms/fetchFarmsPrices.ts
```diff
@@ -85,7 +85,7 @@ const getFarmQuoteTokenPrice = (
   return BIG_ZERO
 }
 
-const fetchFarmsPrices = async (farms: SerializedFarm[]) => {
+const fetchFarmsPrices = (farms: SerializedFarm[]) => {
```
Chef-Cheems, line 88: Should we rename this to `getFarmPrices`? Since its not really fetching anything, just processing.

memoyil: Done

src/state/farms/fetchFarms.ts
```diff
@@ -1,15 +1,47 @@
 import { SerializedFarmConfig } from 'config/constants/types'
-import fetchFarm from './fetchFarm'
```
Chef-Cheems, line 2: I think `fetchFarm.ts` is not used anymore and can be deleted?

memoyil: Done

src/state/pools/hooks.ts
```diff
@@ -22,13 +22,16 @@ import {
...
+import { fetchFarmsPublicDataAsync, nonArchivedFarms } from '../farms'
 
 export const useFetchPublicPoolsData = () => {
   const dispatch = useAppDispatch()
   const slowRefresh = useSlowFresh()
 
   useEffect(() => {
     batch(() => {
+      const activeFarms = nonArchivedFarms.filter((farm) => farm.pid !== 0)
+      await dispatch(fetchFarmsPublicDataAsync(activeFarms.map((farm) => farm.pid)))
```
Chef-Cheems, line 34: I'm not sure if `await` here is respected given that its inside `batch` 🤔 
It might work in a sense that UI updates would be batched but other things executed as stated, but I'm not sure, docs & google don't say much about this.
Also the function doesn't have `async` keyword in front of it, I wonder why linter is ok with it. 🤷‍♂️ 

In any case, maybe its safer to refactor it to await for first dispatch and then call `batch` with the other 2? 🙏

src/state/farms/fetchPublicFarmData.ts
```diff
@@ -15,6 +16,64 @@ type PublicFarmData = {
...
+      resultArray[chunkIndex] = [] // start a new chunk
+    }
+
+    resultArray[chunkIndex].push(item)
+
+    return resultArray
+  }, [])
+}
+
 const fetchFarm = async (farm: SerializedFarm): Promise<PublicFarmData> => {
```
Chef-Cheems, line 77: this `fetchFarm` was only used in `fetchFarm.ts` which is now removed. Save to remove 🕺

memoyil: Done

src/state/pools/hooks.ts
memoyil: Done

chef-jojo: I'm wondering does the `fetchPoolsPublicDataAsync` and `fetchPoolsStakingLimitsAsync` depends on `fetchFarmsPublicDataAsync`? why we need to await `fetchFarmsPublicDataAsync` at the first place? 🤔

memoyil: For Apr calculation, if we don't use await and farms data comes after pools otherwise user has to wait for the next slow refresh cycle to see them which is 1 minute.

memoyil: The issue that i have fixed separately (https://github.com/pancakeswap/pancake-frontend/pull/2863) but it happens with this refactor more often because of less blocking calls so i decided to merge it in here

Chef-Cheems (approved):

PR #2863, opened 2021-12-10
fix: (pools): Pools apr not shown for one slow refresh cycle although the farm data is available

memoyil:
This is an intermittent issue, if at the time fetchPoolsPublicDataAsync called there is no farms data but comes after that short time, user has to wait one slowRefresh cycle (which is 1 minute) to see them

To reproduce:

1. Go to pools page
2. See apr's (except cake pools) not loading for 1 minute

To review:
https://deploy-preview-2863--pancakeswap-dev.netlify.app/

### DIFF
--- a/src/state/farms/fetchFarmsPrices.ts
+++ b/src/state/farms/fetchFarmsPrices.ts
@@ -85,7 +85,7 @@ const getFarmQuoteTokenPrice = (
   return BIG_ZERO
 }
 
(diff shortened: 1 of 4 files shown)


memoyil:
Issue happens more with this pr probably because less blocking calls makes pool fetch runs before farm data available. I merged this pr into https://github.com/pancakeswap/pancake-frontend/pull/2851 therefore we don’t need it seperately

PR #2860, opened 2021-12-10
perf: (pool): Refactor pool staking limit calls

memoyil:
Uses multicall instead of call per pool.

With farm call refactor, total pools calls will be down from ~900 to ~10.

To review:

https://deploy-preview-2860--pancakeswap-dev.netlify.app/

PR #2925, opened 2021-12-21
perf(ifo): Reduce ifo pool network request

chef-jojo:
Add pool action for specify pool zero

### DIFF
--- a/src/state/pools/index.ts
+++ b/src/state/pools/index.ts
@@ -1,31 +1,33 @@
 import { createAsyncThunk, createSlice, PayloadAction } from '@reduxjs/toolkit'
 import BigNumber from 'bignumber.js'
 import poolsConfig from 'config/constants/pools'
-import { BIG_ZERO } from 'utils/bigNumber'
 import {
+  AppThunk,
+  CakeVault,
+  IfoCakeVault,
+  IfoVaultUser,
   PoolsState,
   SerializedPool,
-  CakeVault,
   VaultFees,
   VaultUser,
-  AppThunk,
-  IfoVaultUser,
-  IfoCakeVault,
 } from 'state/types'
+import { getAddress } from 'utils/addressHelpers'
 import { getPoolApr } from 'utils/apr'
+import { BIG_ZERO } from 'utils/bigNumber'
+import { getCakeContract, getMasterchefContract } from 'utils/contractHelpers'
 import { getBalanceNumber } from 'utils/formatBalance'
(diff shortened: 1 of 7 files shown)


src/state/pools/index.ts
```diff
@@ -205,6 +251,17 @@ export const PoolsSlice = createSlice({
   name: 'Pools',
   initialState,
   reducers: {
+    setPoolZeroPublicData: (state, action) => {
```
memoyil, line 254: Shouldn't it be better to make this generic to set any preferred pool data to state?

memoyil:
Most of the network calls aside from farm calls for pools apy calculation (which is already done in root) is done for staking limit. When we apply https://github.com/pancakeswap/pancake-frontend/pull/2860 it will reduce the most of the calls.

Imho pr is good but it creates another place to maintain if something needs to be changed in pools public fetch logic.

chef-jojo:
Even reducing the network calls, we are still looping through all the pools from config here, which is not needed for ifo pages. That is the part of reason our pool page is too slow to navigate to it.

src/state/pools/index.ts
```diff
@@ -58,6 +60,50 @@ const initialState: PoolsState = {
...
+
+export const fetchPoolZeroUserDataAsync = (account: string) => async (dispatch) => {
+  const allowance = await cakeContract.allowance(account, poolZeroAddress)
+  const stakingTokenBalance = await cakeContract.balanceOf(account)
+  const masterChefContract = getMasterchefContract()
+  const pendingReward = await masterChefContract.pendingCake('0', account)
+  const { amount: masterPoolAmount } = await masterChefContract.userInfo('0', account)
+
+  dispatch(
+    setPoolZeroUserData({
```
memoyil, line 101: We can make this generic too

src/state/pools/index.ts
```diff
@@ -59,6 +61,56 @@ const initialState: PoolsState = {
 }
 
 // Thunks
+const poolZero = poolsConfig.find((pool) => pool.sousId === 0)
```
Chef-Cheems, line 64: Suggestion - let's rename "pool zero" to "cake pool" or something. Otherwise it might be confused with IFO pool 0 (basic sale). At least I for a second thought about two IFO pools. 🤔

chef-jojo: I thought pool zero name is a more catchy for me. but cake pool is better for context 👍

Chef-Cheems (approved):

PR #2973, opened 2022-01-02
feat(swap-warning-modals): Update SAFEMOON warning

Aosiepcs:
Update warning for protect degens


chef-jojo:
see https://github.com/pancakeswap/pancake-frontend/pull/2993

PR #2975, opened 2022-01-03
fix: (nft): Mobile nft modal cause a crash when nft metadata loading

memoyil:
To reproduce:

1. Go to nft activity page or profile->activity in mobile
2. Click row when nft metada loading
3. See page crash


src/views/Nft/market/components/Activity/ActivityPrice.tsx
```diff
@@ -14,7 +14,7 @@ const ActivityPrice = ({ bnbBusdPrice, price }) => {
             <Text maxWidth="80px" bold>
               {price.toLocaleString(undefined, {
                 minimumFractionDigits: 0,
-                maximumFractionDigits: 5,
+                maximumFractionDigits: price < 1 ? 5 : 2,
```
memoyil, line 17: Row not visible fully in iphone 8 width devices when price fraction is above 1 and has fraction

memoyil:
Closed since https://github.com/pancakeswap/pancake-frontend/pull/3014 fixed the issue

PR #2995, opened 2022-01-05
feat: Snapshot metadata deprecated

chef-jojo:


.yarnrc
```diff
@@ -0,0 +1 @@
+--install.ignore-engines true
```
chef-jojo, line 1: snapshot js require node > 14. Our netlify node version is 12.x

PR #2997, opened 2022-01-05
feat: List auction farm 12

chef-jojo:


memoyil:
Can we prioritize this https://github.com/pancakeswap/pancake-frontend/pull/2851? Otherwise each new farm puts extra load on nodes :(

Chef-Cheems (approved):

PR #3013, opened 2022-01-07
Update Tokens.md

crackaf:
Updated the file extension name in the doc. From `md` to `ts` and this is the typescript file.

### DIFF
--- a/doc/Tokens.md
+++ b/doc/Tokens.md
@@ -1,6 +1,6 @@
 # Tokens management
 
-All the tokens are in `/config/constans/tokens.md`. They are instances of tthe `Token` class defined in the SDK.
+All the tokens are in `/config/constans/tokens.ts`. They are instances of tthe `Token` class defined in the SDK.
 Before adding a new **farm** or **pool** you need to make sure the Tokens are in this file.
 To add a Token to the exchange lists:
 


doc/Tokens.md
```diff
@@ -1,6 +1,6 @@
 # Tokens management
 
-All the tokens are in `/config/constans/tokens.md`. They are instances of tthe `Token` class defined in the SDK.
+All the tokens are in `/config/constans/tokens.ts`. They are instances of tthe `Token` class defined in the SDK.
```
memoyil, line 3: Can you fix the typo as well

chef-jojo (approved):

crackaf:
@chef-jojo @memoyil  Can I work on the issue regarding testnet 97? I think the solution is to deploy the contracts which are not on 97 and add their address to the src/config/constants/contracts.ts. Will this resolve the errors which we encounter when we change the env from 56 to 97 with their respective RPC? Please guide if this solution isn't complete/correct. Thanks.

chef-jojo:
We have plan to implement testnet support but it's low priority right now. 
I believe you can do it on FE to check chain id and address at least can skip the error 😅
swap sdk needs to be updated to support testnet INIT_CODE_HASH as well

crackaf:
Ok, I'll look into it..

PR #3021, opened 2022-01-10
feat: Remove nft user state and achievement state

memoyil:
It is hard to maintain Nft user state if user got a nft transfer or when nft sold. Therefore it is better to have fresh data always.

It also fixes the issue where catbread nft owners (root cause: we didn't call owner function of that contract and rely on user nft list) cannot see it as owner or put on sale of their nfts.

To review:

https://deploy-preview-3021--pancakeswap-dev.netlify.app/


src/state/nftMarket/helpers.ts
```diff
@@ -940,7 +940,9 @@ export const getCompleteAccountNftData = async (
     })
     .map((nft) => nft.tokenId)
 
-  const marketDataForSaleNfts = await getNftsMarketData({ currentSeller: account.toLowerCase() })
```
memoyil, line 943: no need to fetch again

src/state/profile/hooks.ts
```diff
@@ -71,10 +71,10 @@ export const useAchievementsForAddress = (address: string) => {
         setState({ achievements: [], isFetching: false })
       }
     }
-    if (hasAddressChanged || (!state.isFetching && !state.achievements)) {
```
chef-jojo, line 74: removing this causing loop

src/views/Nft/market/Profile/ConnectedProfile.tsx
```diff
@@ -32,13 +29,11 @@ const TabMenuWrapper = styled(Box)`
...
-  const { profile, isLoading: isProfileLoading } = useProfile()
-  const { achievements, achievementFetchStatus } = useAchievements()
   const { account } = useWeb3React()
-  const { userNftsInitializationState, nfts: userNfts } = useUserNfts()
-
-  useFetchAchievements()
-  useFetchUserNfts()
+  const { isLoading: isProfileLoading, profile } = useProfile()
+  const { lastUpdated, setLastUpdated: refresh } = useLastUpdated()
+  const { achievements, isFetching: isAchievementsFetching } = useAchievementsForAddress(account, lastUpdated)
```
chef-jojo, line 35: can you take a look on https://github.com/pancakeswap/pancake-frontend/commit/05605c0da344cb6ebdfdab8514bb979e435a738a
you can get `mutate` as refresh in swr
```jsx
const { data, status, mutate } = useSWRImmutable([address, 'achievements'], () => getAchievements(address))
return {
 mutate: refresh
}
```

memoyil: Done

src/views/Nft/market/Profile/components/UserNfts.tsx
```diff
@@ -21,11 +24,16 @@ interface SellNftProps {
 }
 
 const UserNfts = () => {
-  const { nfts, userNftsInitializationState } = useUserNfts()
+  const { account } = useWeb3React()
+  const { isLoading: isProfileLoading, profile } = useProfile()
+  const { lastUpdated, setLastUpdated: refresh } = useLastUpdated()
+  const { nfts, isLoading: isNftLoading } = useNftsForAddress(account, profile, isProfileLoading, lastUpdated)
```
chef-jojo, line 30: `useNftsForAddress` is used in ConnectedProfile/UnconnectedProfile and here
This will fetch twice and have two same state

chef-jojo:
I really like the direction not putting all state into redux abusively. Not sure if I can rebase those changes on next js branch 😄

src/views/Nft/market/Profile/components/UserNfts.tsx
memoyil: Done

src/views/Nft/market/Collection/IndividualNFTPage/OneOfAKindNftPage/MainNFTCard.tsx
```diff
@@ -15,19 +15,20 @@ interface MainNFTCardProps {
...
 
   const currentAskPriceAsNumber = nft.marketData?.currentAskPrice ? parseFloat(nft.marketData.currentAskPrice) : 0
   const priceInUsd = multiplyPriceByAmount(bnbBusdPrice, currentAskPriceAsNumber)
   const [onPresentBuyModal] = useModal(<BuyModal nftToBuy={nft} />)
   const [onPresentSellModal] = useModal(
-    <SellModal variant={nft.marketData?.isTradable ? 'edit' : 'sell'} nftToSell={nft} />,
+    <SellModal variant={nft.marketData?.isTradable ? 'edit' : 'sell'} nftToSell={nft} onSuccess={onSuccess} />,
   )
-  const [onEditProfileModal] = useModal(<EditProfileModal />, false)
+  const [onEditProfileModal] = useModal(<EditProfileModal onSuccess={onSuccess} />, false)
```
chefilip, line 31: Could we use Context to avoid Prop Drilling ?

src/views/Nft/market/Profile/ConnectedProfile.tsx
```diff
@@ -32,13 +29,15 @@ const TabMenuWrapper = styled(Box)`
...
-  const { userNftsInitializationState, nfts: userNfts } = useUserNfts()
-
-  useFetchAchievements()
-  useFetchUserNfts()
+  const { isLoading: isProfileLoading, profile } = useProfile()
+  const {
+    achievements,
+    isFetching: isAchievementsFetching,
+    refresh: refreshAchievement,
+  } = useAchievementsForAddress(account)
```
chefilip, line 38: At this moment,  both `ConnectedProfile` and `UnconnectedProfile` use this hook.

Could we level `useAchievementsForAddress` up to the parent ?

chefilip (changes requested):

src/views/Nft/market/Collection/IndividualNFTPage/OneOfAKindNftPage/MainNFTCard.tsx
memoyil: We have state change events out of control of user such as incoming nft transfer, achievement (when claiming current ifo it rewards achievement automatically) or selling nft therefore we should not hold state to show user always correct data without forcing user to refresh webpage completely.

src/views/Nft/market/Profile/ConnectedProfile.tsx
memoyil: If we put one level up and if address of that account is invalid we redundantly send achievement requests to node.

chefilip: I see. We could create a separate component for this line `<>{isConnectedProfile ? <ConnectedProfile /> : <UnconnectedProfile />}</>`. 

It would be a reasonable thing to do.

src/views/Nft/market/Collection/IndividualNFTPage/OneOfAKindNftPage/MainNFTCard.tsx
chefilip: Hmm. Don't get me wrong. I don't suggest global state. I think we just need to add Context where the parent is used. It would be a good call if we try to avoid Prop Drilling in this case.

src/views/Nft/market/hooks/useCompleteNft.tsx
```diff
@@ -0,0 +1,81 @@
...
+        attributes: metadata.attributes,
+        marketData,
+      })
+      setIsLoading(false)
+    }
+
+    const getNftWithLocation = async () => {
+      const metadata = await getNftApi(collectionAddress, tokenId)
+      const [marketData] = await getNftsMarketData({ collection: collectionAddress.toLowerCase(), tokenId }, 1)
+      const tokenOwner = await collectionContract.ownerOf(tokenId)
```
chef-jojo, line 39: 👍

chef-jojo (approved):
Apart from the props drilling, I think it's good to merge. We can figure out a better way later.

memoyil:
> Apart from the props drilling, I think it's good to merge. We can figure out a better way later.

Yes, I will do that in next iteration.

memoyil:
@chef-jojo Can you check my last commit (https://github.com/pancakeswap/pancake-frontend/pull/3021/commits/281b25a3986306d9a1b1c90a0230a100b73ff29d). I did this to not fetch twice

chef-jojo:
> @chef-jojo Can you check my last commit ([281b25a](https://github.com/pancakeswap/pancake-frontend/commit/281b25a3986306d9a1b1c90a0230a100b73ff29d))?. I did this to not fetch twice, although we now fetch more when user accessing another profile pages after first access, it will be less calls for the initial access to the profile pages and to avoid data inconsistency between header and rest of the page when data updated via user action (claim achievement or transfer in/out the nft ).

I doubt it is better 🤔 moving the hooks out of the layout and put it on each pages will lose all states when navigate between pages. You will see nft loading when you swich the tabs. The only hook fetched twice is `useNftsForAddress`, let's convert it to swr, and you can use it on layout and pages without fetching twice

memoyil:
> > @chef-jojo Can you check my last commit ([281b25a](https://github.com/pancakeswap/pancake-frontend/commit/281b25a3986306d9a1b1c90a0230a100b73ff29d))?. I did this to not fetch twice, although we now fetch more when user accessing another profile pages after first access, it will be less calls for the initial access to the profile pages and to avoid data inconsistency between header and rest of the page when data updated via user action (claim achievement or transfer in/out the nft ).
> 
> I doubt it is better 🤔 moving the hooks out of the layout and put it on each pages will lose all states when navigate between pages. You will see nft loading when you swich the tabs. The only hook fetched twice is `useNftsForAddress`, let's convert it to swr, and you can use it on layout and pages without fetching twice

Done, can you review again?

src/views/Nft/market/Profile/index.tsx
```diff
@@ -2,24 +2,47 @@ import React, { FC } from 'react'
...
 import { useRouter } from 'next/router'
 import { isAddress } from 'utils'
-import { Flex, Text } from '@pancakeswap/uikit'
+import { Box, Flex, Text } from '@pancakeswap/uikit'
 import Page from 'components/Layout/Page'
 import { useTranslation } from 'contexts/Localization'
-import ConnectedProfile from './ConnectedProfile'
-import UnconnectedProfile from './UnconnectedProfile'
+import styled from 'styled-components'
+import useLastUpdated from 'hooks/useLastUpdated'
```
chef-jojo, line 9: remove this

memoyil: Done

chef-jojo (approved):

PR #3027, opened 2022-01-12
fix(ifo): Show zero balance instead of loading

chefilip:
When user has zero balance, showing 0,00$ instead of loading.

### DIFF
--- a/src/views/Pools/components/CakeVaultCard/VaultCardActions/index.tsx
+++ b/src/views/Pools/components/CakeVaultCard/VaultCardActions/index.tsx
@@ -24,7 +24,7 @@ export const IfoVaultCardAvgBalance = () => {
   const credit = useIfoPoolCredit()
 
   const cakeAsNumberBalance = getBalanceNumber(credit)
-  const creditsDollarValue = useBUSDCakeAmount(cakeAsNumberBalance)
+  const creditsDollarValue: number | undefined = useBUSDCakeAmount(cakeAsNumberBalance)
 
   return (
     <>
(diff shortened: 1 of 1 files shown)


src/views/Pools/components/CakeVaultCard/VaultCardActions/index.tsx
```diff
@@ -55,7 +55,7 @@ export const IfoVaultCardAvgBalance = () => {
       <Flex flexDirection="column" pb="16px">
         <Balance fontSize="20px" bold value={cakeAsNumberBalance} decimals={5} />
         <Text fontSize="12px" color="textSubtle" display="flex">
-          {creditsDollarValue ? (
+          {creditsDollarValue !== undefined ? (
             <Balance
```
memoyil, line 59: Imho Number.isFinite(creditsDollarValue) is more descriptive

chefilip: Screen Shot 2022-01-12 at 14 59 01

Thanks. Using type check in this case would be enough.

memoyil: True, but if logic of that method changes in the future we would be on the safe side if we handle more cases

chefilip: At this moment, we can use type check as a contract. In the future, if the method is changed, the contract would be broken. Type Check could notify that. I think it would be safe.

chef-jojo (approved):

PR #3030, opened 2022-01-12
fix(ifo): Avoid resize ifo section when loading

chefilip:
when `grid` div contains `margin-x: auto`, it would resize the section somehow which leads to unexpected resize section.


memoyil:
Can you put padding to container? Otherwise bunny will look like this

image

chefilip:
> Can you put padding to container? Otherwise bunny will look like this
> 
> image

For 320px width, we might have to accept the limitation of the current design.

chef-jojo (approved):

Issue #3029, opened 2022-01-12
pid: 0, points to CAKE Token Contract


jamel-86:
The first farm 0 is actually not  a LP pair contract, it points to Cake Token contract, how can I create this pid 0 with the factory contract ?

shopglobal:
PID 0 is a staking pool for CAKE(create this pool at the masterchef). So the meaning is to reward single stakers for staking CAKE with SYRUP issued through Masterchef which then gets tracked back on frontend. From there users can harvest SYRUP for CAKE and that process burns the SYRUP and mints new CAKE. Hope this helps

PR #3031, opened 2022-01-12
fix: (ifo): Ifo card width is narrower than it should be in mobile devices

memoyil:
Iphone 8 size 

Before:
image

After: 
image


memoyil:
Closed, it is fixed already in #3029 3030

PR #3032, opened 2022-01-12
refactor: (pool): Don't show max stake info when pool is finished

memoyil:


memoyil:
It is already not fetching when pool is finished. My development state still had the staking info therefore it showed

PR #3034, opened 2022-01-12
feat: Make scrolling smooth

memoyil:
Smooth scrolling for better user experience

### DIFF
--- a/src/views/Info/Tokens/TokenPage.tsx
+++ b/src/views/Info/Tokens/TokenPage.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable no-nested-ternary */
-import React, { useEffect, useMemo } from 'react'
+import React, { useMemo } from 'react'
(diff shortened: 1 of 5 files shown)


src/views/Info/Tokens/TokenPage.tsx
```diff
@@ -72,7 +72,11 @@ const TokenPage: React.FC<RouteComponentProps<{ address: string }>> = ({
 
   // Needed to scroll up if user comes to this page by clicking on entry in the table
   useEffect(() => {
-    window.scrollTo(0, 0)
+    window.scrollTo({
```
chefilip, line 75: Nice. Could you do me a favor by creating a common hook `useScrollTop`  in the `hooks` folder ? It would avoid copy and paste in the future.

memoyil:
Closed since next-js will handle it as @chef-jojo mentioned

PR #3046, opened 2022-01-14
fix: (intersection): Intersection observer cannot detect new mounted element after re-render

memoyil:
Issue mainly happens in https://github.com/pancakeswap/pancake-frontend/pull/2865 where nftmedia with video mounted after first render.

### DIFF
--- a/src/hooks/useIntersectionObserver.ts
+++ b/src/hooks/useIntersectionObserver.ts
@@ -1,8 +1,8 @@
 import { useEffect, useRef, useState } from 'react'
 
 const useIntersectionObserver = () => {
-  const [observerRef] = useState(() => (element) => setObserverRefElement(element))
   const [observerRefElement, setObserverRefElement] = useState(null)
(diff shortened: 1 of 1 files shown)


src/hooks/useIntersectionObserver.ts
```diff
@@ -1,7 +1,8 @@
 import { useEffect, useRef, useState } from 'react'
 
 const useIntersectionObserver = () => {
-  const observerRef = useRef<HTMLDivElement>(null)
+  const [observerRef] = useState(() => (element) => setObserverRefElement(element))
+  const [observerRefElement, setObserverRefElement] = useState(null)
```
Chef-Cheems, line 5:
```suggestion
  const [observerRefElement, setObserverRefElement] = useState(null)
  const [observerRef] = useState(() => (element) => setObserverRefElement(element))
```

Chef-Cheems (approved):

PR #3063, opened 2022-01-17
refactor: Import ethers per package

chef-jojo:
Reduce bundle size

before
CleanShot 2022-01-17 at 14 12 24@2x

after
CleanShot 2022-01-17 at 14 28 27@2x


src/App.tsx
```diff
@@ -79,7 +78,6 @@ const App: React.FC = () => {
     <Router history={history}>
       <ResetCSS />
       <GlobalStyle />
-      <GlobalCheckClaimStatus excludeLocations={[]} />
```
chef-jojo, line 82: AnniversaryAchievement no longer used, right? @Chef-Cheems

Chef-Cheems (approved):

PR #3065, opened 2022-01-17
fix: (pool): Unstake tx wait undefined error

memoyil:
Fixes:
TypeError: b.wait is not a function


chef-jojo (approved):
nit: should add types for these, we have typechain now

PR #3079, opened 2022-01-18
fix(ifo): check approval before showing stake in

chefilip:
- Add Enable Pool Button in IFO step
- Apply SWR for `useCheckVaultApprovalStatus` hook

### DIFF
--- a/src/views/Pools/hooks/useApprove.tsx
+++ b/src/views/Pools/hooks/useApprove.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useState } from 'react'
+import React, { useCallback, useMemo, useState } from 'react'
 import { useWeb3React } from '@web3-react/core'
 import { ethers, Contract } from 'ethers'
 import { useAppDispatch } from 'state'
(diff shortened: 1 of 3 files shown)


src/views/Pools/hooks/useApprove.tsx
```diff
@@ -95,23 +95,25 @@ export const useVaultApprove = (vaultKey: VaultKey, setLastUpdated: () => void)
...
+            methodName: 'allowance',
+            params: [account, vaultPoolContract.address],
+          }
+        : null,
+    [account, cakeContract, vaultPoolContract.address],
+  )
+
+  const mutate = useMutateSWRContract(key)
+
+  const { data } = useSWRContract(key)
```
chef-jojo, line 116: what's the purpose of  `useMutateSWRContract`?
We can get mutate from useSWRContract

```suggestion
  const { data, mutate } = useSWRContract(key)
```

chefilip: Don't know it already has the `mutate`. Will remove it.

chef-jojo (approved):

PR #3098, opened 2022-01-20
fix: (profile): Handle edit profile update properly

memoyil:
To reproduce:

1. Change profile picture
2. See picture not changed and user nfts not updated in profile page

We only need refresh profile trigger when in user profile page since profile data is fetched by useProfileByAddress hook instead of redux.

### DIFF
--- a/src/pages/nfts/profile/[accountAddress].tsx
+++ b/src/pages/nfts/profile/[accountAddress].tsx
@@ -12,15 +12,31 @@ const NftProfilePage = () => {
   const { account } = useWeb3React()
   const accountAddress = useRouter().query.accountAddress as string
   const isConnectedProfile = account?.toLowerCase() === accountAddress?.toLowerCase()
-  const { profile: profileHookState, isFetching: isProfileFetching } = useProfileForAddress(accountAddress)
+  const {
+    profile: profileHookState,
+    isFetching: isProfileFetching,
+    refresh: refreshProfile,
+  } = useProfileForAddress(accountAddress)
   const { profile } = profileHookState || {}
-  const { nfts, isLoading: isNftLoading, refresh } = useNftsForAddress(accountAddress, profile, isProfileFetching)
+  const {
+    nfts,
+    isLoading: isNftLoading,
+    refresh: refreshUserNfts,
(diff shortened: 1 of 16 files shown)


src/pages/nfts/profile/[accountAddress].tsx
```diff
@@ -12,15 +12,31 @@ const NftProfilePage = () => {
...
   return (
     <>
       <SubMenu />
       {isConnectedProfile ? (
-        <UserNfts nfts={nfts} isLoading={isNftLoading} onSuccess={refresh} />
+        <UserNfts
+          nfts={nfts}
+          isLoading={isNftLoading}
+          onSuccessSale={refreshUserNfts}
+          onSuccessEditProfile={async () => {
```
Comment (memoyil) [inline src/pages/nfts/profile/[accountAddress].tsx:35]: we need to fetch both

src/views/Nft/market/Collection/IndividualNFTPage/OneOfAKindNftPage/MainNFTCard.tsx
```diff
@@ -27,9 +27,9 @@ const MainNFTCard: React.FC<MainNFTCardProps> = ({ nft, isOwnNft, nftIsProfilePi
   const priceInUsd = multiplyPriceByAmount(bnbBusdPrice, currentAskPriceAsNumber)
   const [onPresentBuyModal] = useModal(<BuyModal nftToBuy={nft} />)
   const [onPresentSellModal] = useModal(
-    <SellModal variant={nft.marketData?.isTradable ? 'edit' : 'sell'} nftToSell={nft} onSuccess={onSuccess} />,
+    <SellModal variant={nft.marketData?.isTradable ? 'edit' : 'sell'} nftToSell={nft} onSuccessSale={onSuccess} />,
   )
-  const [onEditProfileModal] = useModal(<EditProfileModal onSuccess={onSuccess} />, false)
+  const [onEditProfileModal] = useModal(<EditProfileModal />, false)
```
memoyil, line 32: We don't need to trigger onSuccess here, since useProfile hook in useCompleteNft will be triggered when data changed

chef-jojo (approved):

PR #3099, opened 2022-01-20
fix(profile): Avoid re-enable for approval nfts

chefilip:
Avoid re-enabling NFT when re-selecting profile pic.

### DIFF
--- a/src/hooks/useApproveConfirmTransaction.tsx
+++ b/src/hooks/useApproveConfirmTransaction.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useReducer, useRef } from 'react'
+import React, { useEffect, useReducer, useRef, useCallback } from 'react'
 import noop from 'lodash/noop'
 import { useWeb3React } from '@web3-react/core'
 import { ethers } from 'ethers'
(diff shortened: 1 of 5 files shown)


src/state/nftMarket/hooks.ts
```diff
@@ -105,6 +109,35 @@ export const useUserNfts = (): UserNftsState => {
...
+    [nftsInWallet],
+  )
+
+  const { data } = useSWRMulticall(erc721Abi, nftApprovalCalls)
+
+  const approvedTokenIds = Array.isArray(data)
+    ? data
+        .flat()
+        .reduce(
+          (acc, address, index) => ({ ...acc, [nftsInWallet[index].tokenId]: getPancakeProfileAddress() === address }),
```
memoyil, line 133: This should check against user address no?

chefilip: `nftsInWallet` is already owned by user address. This check is for NFTs enabled to be used in Profile contract.

memoyil: This i agree, but contract returns the address that approves the token not the pancake profile address afaik. Correct me if i am wrong

chefilip: Screen Shot 2022-01-20 at 20 32 47

The approve logic is `approvalList[tokenId] = profileAddress`. Then the `getApprove` function would return `profileAddress` by `tokenId` 

You could check the contract for more detail. https://bscscan.com/address/0xDf7952B35f24aCF7fC0487D01c8d5690a60DBa07#code

memoyil: Sorry my bad :) I confused my address with profile address. :)

src/state/nftMarket/hooks.ts
```diff
@@ -99,6 +103,35 @@ export const useGetNFTInitializationState = () => {
...
+        return {
+          address: collectionAddress,
+          name: 'getApproved',
+          params: [tokenId],
+        }
+      }),
+    [nftsInWallet],
+  )
+
+  const { data } = useSWRMulticall(erc721Abi, nftApprovalCalls)
```
memoyil, line 121: What is the advantage of using swr here?

src/hooks/useApproveConfirmTransaction.tsx
```diff
@@ -1,5 +1,5 @@
-import React, { useEffect, useReducer, useRef } from 'react'
-import noop from 'lodash/noop'
+import React, { useEffect, useReducer, useRef, useCallback } from 'react'
+import { noop } from 'lodash'
```
chef-jojo, line 2: Import path

src/hooks/useSWRContract.ts
```diff
@@ -120,6 +121,10 @@ export function useSWRContract<
   )
 }
 
+export function useSWRMulticall(abi: any[], calls: Call[], options: MulticallOptions = { requireSuccess: true }) {
+  return useSWR(calls, () => multicallv2(abi, calls, options))
```
chef-jojo, line 125: Use this in caution with multicall. By default, it refetch when on focus or mount.

src/state/nftMarket/hooks.ts
chef-jojo: Will fetch by selectedNft affect user experience when switching selected? Just wondering if it is good enough to not use multicall here

src/hooks/useApproveConfirmTransaction.tsx
chefilip: Oops. It seems prettier auto updates this part. Will check

chefilip: Oops. Noted

src/hooks/useSWRContract.ts
chefilip: Noted

src/state/nftMarket/hooks.ts
chefilip: Without multicall, whenever user selects a pic, a loading status is shown, which is not a good UX. For this feature, it's reasonable to call it in one batch.

chef-jojo (approved):

PR #3115, opened 2022-01-23
fix: (nft): Missing items in Collection listing

memoyil:
On sale collections listing ordered by token id shows less results than it should: 

1. Go to https://pancakeswap.finance/nfts/collections/0x4bd2a30435e6624CcDee4C60229250A84a2E4cD6
2. Order by token id with show only on sale filter
3. See it shows less results

Missing nfts when select filter by all: 

1. Go to https://pancakeswap.finance/nfts/collections/0x8aE5c7b7946B7bEc7670c9F5b89203a32B05E25d
2. Select All
3. You will still see 59 results instead of 84

With new approach we will not use redux for fetching nfts (which we shouldn't, because storing nfts in redux eventually leads to stale data)


src/views/Nft/market/hooks/useCollectionNfts.tsx
```diff
@@ -0,0 +1,163 @@
...
+  const [itemListingSettings, setItemListingSettings] = useState({
+    page: 1,
+    field,
+    direction,
+    showOnlyNftsOnSale,
+    nftFilters,
+  })
+
+  const itemListingSettingsJson = JSON.stringify(itemListingSettings)
+  const filtersJson = JSON.stringify(nftFilters)
```
chef-jojo, line 31: is this stringify and parse to avoid useEffect necessary? I thought it won't re-run if the state is not changed

memoyil: Filters can trigger infinite rerender if there is no data stored in redux see other pr for that (https://github.com/pancakeswap/pancake-frontend/pull/3119)

Stringfy Item listings to avoid trigger fetching useEffect multiple times on mount, since all other effects triggered and setting a new object on mount causes useEffect that fetching nfts triggered multiple times.

src/views/Nft/market/hooks/useCollectionNfts.tsx
```diff
@@ -0,0 +1,161 @@
...
+    if (collectionNftsResponse?.data) {
+      tokenIds = Object.values(collectionNftsResponse.data).map((nft) => nft.tokenId)
+    }
+  }
+
+  if (tokenIds) {
+    const nftsMarket = await getMarketDataForTokenIds(address, tokenIds)
+
+    const responsesPromises = tokenIds.map(async (id) => {
+      const apiMetadata = collectionNftsResponse ? collectionNftsResponse.data[id] : await getNftApi(address, id)
```
chefilip, line 88: `apiMetadata` type should be explicit because we don't know what results are returned. It might lead to TypeError

src/views/Nft/market/hooks/useCollectionNfts.tsx
```diff
@@ -0,0 +1,161 @@
...
+      ...accum,
+      [attr.traitType]: attr.value,
+    }),
+    {},
+  )
+  const attrFilters = !isEmpty(attrParams) ? await fetchNftsFiltered(address, attrParams) : null
+  return attrFilters ? Object.values(attrFilters.data).map((apiToken) => apiToken.tokenId) : null
+}
+
+const fetchMarketDataNfts = async (address: string, settings: ItemListingSettings, page: number) => {
```
chefilip, line 37: define return type for `fetchMarketDataNfts` and `fetchAllNfts` for consistency.

memoyil: Done

src/views/Nft/market/hooks/useCollectionNfts.tsx
memoyil: Done

src/views/Nft/market/hooks/useCollectionNfts.tsx
```diff
@@ -0,0 +1,167 @@
...
+  page: number,
+): Promise<NftToken[]> => {
+  const tokenIdsFromFilter = await fetchTokenIdsFromFilter(address, settings)
+  const whereClause = tokenIdsFromFilter
+    ? {
+        collection: address.toLowerCase(),
+        isTradable: true,
+        tokenId_in: tokenIdsFromFilter,
+      }
+    : { collection: address.toLowerCase(), isTradable: true }
```
chef-jojo, line 49: Should reflect to `showOnlyNftsOnSale`

src/views/Nft/market/hooks/useCollectionNfts.tsx
```diff
@@ -0,0 +1,167 @@
...
+      field,
+      direction,
+      showOnlyNftsOnSale,
+      nftFilters: JSON.parse(filtersJson),
+    }))
+  }, [field, direction, showOnlyNftsOnSale, filtersJson])
+
+  const {
+    data: nfts,
+    isValidating,
```
chef-jojo, line 140: Showing loading when validating is not a good idea

memoyil: Fixed

src/views/Nft/market/hooks/useCollectionNfts.tsx
memoyil: fetchMarketDataNfts is called when showOnlyNftsOnSale is true, should we put that condition inside that method?

chef-jojo: umm.. weird nfts are missing if you select filters

memoyil: Fixed 👍

src/views/Nft/market/hooks/useCollectionNfts.tsx
```diff
@@ -0,0 +1,268 @@
...
+      }
+      if (newNfts.length < REQUEST_SIZE) {
+        isLastPage.current = true
+      }
+      return newNfts
+    },
+    { revalidateFirstPage: false },
+  )
+
+  const uniqueNftList: NftToken[] = nfts ? uniqBy(nfts.flat(), 'tokenId') : []
```
chefilip, line 257: I think we should `useMemo` this, otherwise it will keep returning new array reference

src/views/Nft/market/hooks/useCollectionNfts.tsx
```diff
@@ -0,0 +1,268 @@
...
+    data: nfts,
+    status,
+    size,
+    setSize,
+  } = useSWRInfinite(
+    (pageIndex, previousPageData) => {
+      if (pageIndex !== 0 && previousPageData && !previousPageData.length) return null
+      return [collectionAddress, itemListingSettingsJson, pageIndex, 'collectionNfts']
+    },
+    async (address, settingsJson, page) => {
```
chefilip, line 232: it seems `address` not used in this function

src/views/Nft/market/hooks/useCollectionNfts.tsx
```diff
@@ -0,0 +1,268 @@
...
+      const settings: ItemListingSettings = JSON.parse(settingsJson)
+      const tokenIdsFromFilter = await fetchTokenIdsFromFilter(collection.address, settings)
```
chefilip, line 234: `collection` might be `undefined`. Should add a gaurd here `collection?.address` ?

src/views/Nft/market/hooks/useCollectionNfts.tsx
```diff
@@ -0,0 +1,268 @@
...
+    field,
+    direction,
+    showOnlyNftsOnSale,
+    nftFilters,
+  })
+
+  // We don't know the amount in advance if nft filters exist
+  const resultSize = !Object.keys(nftFilters).length
+    ? showOnlyNftsOnSale || field !== 'tokenId'
+      ? collection.numberTokensListed
```
chefilip, line 202: `collection` might be `undefined`

src/views/Nft/market/hooks/useCollectionNfts.tsx
```diff
@@ -0,0 +1,268 @@
...
+  page: number,
+  tokenIdsFromFilter: string[],
+  fetchedNfts: NftToken[],
+  fallbackMode: React.MutableRefObject<boolean>,
+  fallbackPage: React.MutableRefObject<number>,
+): Promise<NftToken[]> => {
+  const newNfts: NftToken[] = []
+  let tokenIds: string[] = []
+  let collectionNftsResponse: ApiResponseCollectionTokens = null
+  if (settings.field !== 'tokenId' && !fallbackMode.current) {
```
chefilip, line 125: hmm. Allowing `fetchAllNfts` to know about `useRef` is quite strange to me. The function handles more responsibilities than it should be, I think.

src/views/Nft/market/hooks/useCollectionNfts.tsx
memoyil: Done

src/views/Nft/market/hooks/useCollectionNfts.tsx
memoyil: Done

src/views/Nft/market/hooks/useCollectionNfts.tsx
memoyil: Done

src/views/Nft/market/hooks/useCollectionNfts.tsx
memoyil: Yes, but it should be there to keep the order

src/views/Nft/market/hooks/useCollectionNfts.tsx
memoyil: Done

chefilip (approved):

PR #3127, opened 2022-01-24
fix: (nft): Sort and order options defaulted when clicking filter clear button

memoyil:
To reproduce:

1. Go collection items page
2. Change filter by all and order by
3. Filter by trait
4. Click clear
5. See sort and order options defaulted although it is not visible to user on options


chef-jojo:
I think the clear all button is only for traits. It only show when traits filter is selected

memoyil:
> I think the clear all button is only for traits. It only show when traits filter is selected

Done 👍

chef-jojo (approved):

PR #3130, opened 2022-01-25
feat(pool): Add user requirement

chef-jojo:


src/state/user/hooks/index.tsx
```diff
@@ -357,7 +358,7 @@ export function useRemoveUserAddedToken(): (chainId: number, address: string) =>
 }
 
 export function useGasPrice(): string {
-  const chainId = process.env.NEXT_PUBLIC_CHAIN_ID
+  const chainId = CHIAN_ID
```
memoyil, line 361: Typo

src/utils/getGasPrice.ts
```diff
@@ -1,12 +1,13 @@
...
+import { CHIAN_ID } from 'config/constants/networks'
 import store from 'state'
 import { GAS_PRICE_GWEI } from 'state/user/hooks/helpers'
 
 /**
  * Function to return gasPrice outwith a react component
  */
 const getGasPrice = (): string => {
-  const chainId = process.env.NEXT_PUBLIC_CHAIN_ID
+  const chainId = CHIAN_ID
```
memoyil, line 10: Typo

src/state/user/hooks/index.tsx
chef-jojo: thanks!

src/state/pools/fetchPools.ts
```diff
@@ -51,26 +52,25 @@ export const fetchPoolsBlockLimits = async () => {
...
-  const bnbPool = poolsConfig.filter((p) => p.stakingToken.symbol === 'BNB')
-
-  const callsNonBnbPools = nonBnbPools.map((poolConfig) => {
-    return {
-      address: poolConfig.stakingToken.address,
-      name: 'balanceOf',
-      params: [getAddress(poolConfig.contractAddress)],
-    }
-  })
+const nonBnbPools = poolsConfig.filter((p) => p.stakingToken.symbol !== 'BNB')
```
memoyil, line 55: we can use lodash partition, something like this

`
 const [bnbPool, nonBnbPools] = partition(pools, (p) => p.stakingToken.symbol === 'BNB')
`

memoyil, line 55: we can use lodash partition, something like this

`
 const [bnbPool, nonBnbPools] = partition(pools, (p) => p.stakingToken.symbol === 'BNB')
`

memoyil, line 55: we can use lodash partition, something like this

`
 const [bnbPool, nonBnbPools] = partition(pools, (p) => p.stakingToken.symbol === 'BNB')
`

Chef-Cheems: @memoyil @chef-jojo we might also just ignore the split into 2 arrays and just use whole thing since later on:
1. bnb.address is actually configured as wbnb.address anyway
2. wbnbABI and cakeABI are the same ERC20 stuff, at least here we just need `balanceOf`. The ABI entry for it is slightly different but mostly just regarding name and other non-important stuff. 🤔 

We perhaps should use erc20.json here to begin with 🤷

src/hooks/useContract.ts
```diff
@@ -21,7 +21,6 @@ import {
   getIfoPoolContract,
   getPredictionsContract,
   getChainlinkOracleContract,
-  getSouschefV2Contract,
```
Chef-Cheems, line 24: @chef-jojo it looks like `getSouschefV2Contract` is not used anywhere anymore.

src/state/pools/fetchPools.ts
chef-jojo: Good point! I will refactor it following this. Actually I just moved this out of runtime function 😅

src/hooks/useContract.ts
chef-jojo: export for testing. removed getSouschefV3Contract

Chef-Cheems (approved):

PR #3133, opened 2022-01-26
feat(swap): Add hard refresh button

chefilip:
CleanShot 2022-01-26 at 12 46 22


src/state/block/hooks.ts
```diff
@@ -8,6 +9,35 @@ import { State } from '../types'
...
+
+  const dispatch = useAppDispatch()
+
+  const fetchBlockNumber = useCallback(() => {
+    if (!isLoading) {
+      setFetch(true)
+    }
+  }, [isLoading])
+
+  const { isValidating, data: blockNumber } = useSWR(isLoading ? ['blockNumber'] : null, () =>
```
chef-jojo, line 23: How about use `mutate`, and `const isLoading = useDebounce(isValidating)`  to get the delay loading?

src/views/Swap/index.tsx
```diff
@@ -12,13 +12,15 @@ import {
...
   ArrowUpDownIcon,
+  Skeleton,
 } from '@pancakeswap/uikit'
 import { useIsTransactionUnsupported } from 'hooks/Trades'
 import UnsupportedCurrencyFooter from 'components/UnsupportedCurrencyFooter'
 import Footer from 'components/Menu/Footer'
 import { useRouter } from 'next/router'
 import { useTranslation } from 'contexts/Localization'
 import SwapWarningTokens from 'config/constants/swapWarningTokens'
+import { useFetchBlockNumber } from 'state/block/hooks'
```
chef-jojo, line 23: this hook with delay loading is only for swap page only, and `useFetchBlockNumber` naming is quite misleading. 
I recommend move it to swap views folder. state related code is not easy to tree shaking base on pages right now.

chef-jojo (approved):

PR #3136, opened 2022-01-26
fix(swap): Remove SAFEMOON from frontend default list

Aosiepcs:
we yeeted from /toolkit https://github.com/pancakeswap/pancake-toolkit/pull/388

but still in default token list, please yeet from also here

Screen Shot 2022-01-26 at 2 49 28 PM

from preview:

no safemoon finally

Screen Shot 2022-01-26 at 2 55 28 PM

if you import, you will see 2 warnin (anyone create and special safemoon warnn)
Screen Shot 2022-01-26 at 2 55 37 PM

seems fine.


chef-jojo (approved):
👍

PR #3142, opened 2022-01-28
chore: Update Lottery How to Play wording

ChefMomota:


chefilip:
lgtm

jiyuunin:
As mentioned in [this comment on Reddit](https://www.reddit.com/r/pancakeswap/comments/ske0tj/comment/hvks6wm/?utm_source=reddit&utm_medium=web2x&context=3), rounds are not happening every 36 hours but alternating between 36 hours and 12 hours so I would suggest to rephrase the sentence to something like
```
There is one draw every day alternating between 0 AM UTC and 12 PM UTC.
```
By the way, the space between `every` and `day` should be kept.

chef-jojo (approved):

PR #3159, opened 2022-01-31
feat: (profile): Remove profile state from redux

memoyil:
One of the issues that is seen when using redux states is that if other users changes its avatar, that change cannot be seen until full refresh. So if user buys other user nft which is the profile picture of them, it will still see its freshly owned nft as profile picture of the seller.


src/pages/nfts/profile/[accountAddress].tsx
```diff
@@ -14,9 +14,9 @@ const NftProfilePage = () => {
   const isConnectedProfile = account?.toLowerCase() === accountAddress?.toLowerCase()
   const {
     profile: profileHookState,
-    isFetching: isProfileFetching,
+    isValidating: isProfileFetching,
```
Comment (memoyil) [inline src/pages/nfts/profile/[accountAddress].tsx:17]: We have to use isValidating directly here, since isFetching always comes false after first fetch

src/views/Nft/market/Profile/index.tsx
```diff
@@ -33,9 +33,9 @@ const NftProfile: FC = ({ children }) => {
 
   const {
     profile: profileHookState,
-    isFetching: isProfileFetching,
+    isValidating: isProfileFetching,
```
memoyil, line 36: See comment https://github.com/pancakeswap/pancake-frontend/pull/3159/files#r795953381

src/views/Predictions/Leaderboard/components/Results/ConnectedWalletResult.tsx
```diff
@@ -19,7 +18,6 @@ const ConnectedWalletResult = () => {
 
   useEffect(() => {
     if (account) {
-      dispatch(fetchProfileAvatar(account))
```
memoyil, line 22: No need to fetch again since already fetched

src/views/Predictions/Leaderboard/components/Results/RankingCard.tsx
```diff
@@ -51,7 +51,8 @@ const getRankingColor = (rank: number) => {
 const RankingCard: React.FC<RankingCardProps> = ({ rank, user }) => {
   const { t } = useTranslation()
   const rankColor = getRankingColor(rank)
-  const profileAvatar = useGetProfileAvatar(user.id)
```
memoyil, line 54: There is not much diff between fetching profile via useGetProfileAvatar and useProfileForAddress in request aspect. Only difference is team data fetching from onchain which can be negligible.

src/pages/nfts/profile/[accountAddress].tsx
```diff
@@ -14,9 +14,13 @@ const NftProfilePage = () => {
...
     refresh: refreshProfile,
-  } = useProfileForAddress(accountAddress)
+  } = useProfileForAddress(accountAddress, {
```
Comment (memoyil) [inline src/pages/nfts/profile/[accountAddress].tsx:19]: Have to revalidate since data (profile) that used in useNftsForAddress uses useSwr

src/state/index.ts
```diff
@@ -35,7 +34,6 @@ const persistedReducer = persistReducer(
     farms: farmsReducer,
     pools: poolsReducer,
     predictions: predictionsReducer,
-    profile: profileReducer,
```
chef-jojo, line 38: Please remove `profile` from `PERSISTED_KEYS`

src/pages/nfts/profile/[accountAddress].tsx
Comment (chef-jojo) [reply src/pages/nfts/profile/[accountAddress].tsx:17]: Not sure if I understand it correctly. You use `isValidating` to trigger refetch on `useNftsForAddress` ?

src/views/Nft/market/hooks/useNftsForAddress.tsx
```diff
@@ -26,12 +26,10 @@ const useNftsForAddress = (account: string, profile: Profile, isProfileFetching:
...
 
-  const { status, data, mutate } = useSWR([account, collections, isProfileFetching, 'userNfts'], async () => {
-    if (!isProfileFetching && !isEmpty(collections) && isAddress(account)) {
-      return getCompleteAccountNftData(account, collections, profileNftWithCollectionAddress)
-    }
-    return []
-  })
+  const { status, data, mutate } = useSWR(
+    !isProfileFetching && !isEmpty(collections) && isAddress(account) ? [account, 'userNfts'] : null,
+    async () => getCompleteAccountNftData(account, collections, profileNftWithCollectionAddress),
```
chef-jojo, line 31: should we consider `[account, collections, profileNftWithCollectionAddress, 'userNfts']` as key to revalidate everytime those 4 variables change?

src/state/index.ts
memoyil: Done

src/pages/nfts/profile/[accountAddress].tsx
Comment (memoyil) [reply src/pages/nfts/profile/[accountAddress].tsx:17]: It is not about triggering fetch actually, if we use isFetching that comes always true after inital fetch, when revalidating usernfts it will fetch twice (first with old profile data, second with after profile data is fetched again)

chef-jojo:
seems like the profile page keep refreshing

memoyil:
> seems like the profile page keep refreshing

I couldn't reproduce it, what are your steps to reproduce?

chef-jojo:
> > seems like the profile page keep refreshing
> 
> I couldn't reproduce it, what are your steps to reproduce?

revalidating: data will becomes undefined and show placeholder. I'm not sure if it's related to swr key changes, Although it shouldn't for profile address? 🤔
if keys are changes, we will need this middleware https://swr.vercel.app/docs/middleware#keep-previous-result

memoyil:
> > > seems like the profile page keep refreshing
> > 
> > 
> > I couldn't reproduce it, what are your steps to reproduce?
> 
> Try switching tab revalidating: data will becomes undefined and show placeholder. I'm not sure if it's related to swr key changes, Although it shouldn't for profile address? 🤔
> 
> if keys are changes, we will need this middleware https://swr.vercel.app/docs/middleware#keep-previous-result

`  const { status, data, mutate } = useSWR(
    !isProfileFetching && !isEmpty(collections) && isAddress(account) ? [account, 'userNfts'] : null,
    async () => getCompleteAccountNftData(account, collections, profileNftWithCollectionAddress),
  )

  return { nfts: data ?? [], isLoading: status !== FetchStatus.Fetched, refresh: mutate }`
  
  This issue happens here when we revalidate the profile and it switched to isProfileFetching true that makes key null therefore data returns undefined and status returns idle which makes isLoading true.

memoyil:
> > > seems like the profile page keep refreshing
> > 
> > 
> > I couldn't reproduce it, what are your steps to reproduce?
> 
> Try switching tab revalidating: data will becomes undefined and show placeholder. I'm not sure if it's related to swr key changes, Although it shouldn't for profile address? 🤔
> 
> if keys are changes, we will need this middleware https://swr.vercel.app/docs/middleware#keep-previous-result

Done 👍

chefilip:
In general, it looks great. One high-level concern is that using `mutate` like `dispatch` seems not fit well with the `swr` paradigm.

memoyil:
> In general, it looks great. One high-level concern is that using `mutate` like `dispatch` seems not fit well with the `swr` paradigm.

We need to refresh the cache when user interacts with their own profile, Imho mutate is the safest option. What should we apply instead?

src/state/index.ts
chef-jojo: Looks like we'll have error when deleted key. Need to add `blacklist` or add migration for it.
https://github.com/rt2zz/redux-persist/issues/618

src/state/profile/hooks.ts
```diff
@@ -44,34 +40,17 @@ export const useAchievementsForAddress = (address: string) => {
...
 
 export const useProfile = () => {
-  const { isInitialized, isLoading, data, hasRegistered }: ProfileState = useSelector((state: State) => state.profile)
-
+  const { account } = useWeb3React()
+  const {
+    data: { profile, hasRegistered } = { profile: null, hasRegistered: false } as GetProfileResponse,
+    status,
+    mutate,
+  } = useSWRImmutable(account ? [account, 'profile'] : null, () => getProfile(account))
```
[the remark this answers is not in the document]
memoyil: Done

src/state/index.ts
memoyil: Done 👍 , added to blacklist

chef-jojo:
LGTM. Only one thing ProfileHeader points keep showing skeleton when validating

memoyil:
> LGTM. Only one thing ProfileHeader points keep showing skeleton when validating

Done 👍

chefilip:
> > In general, it looks great. One high-level concern is that using `mutate` like `dispatch` seems not fit well with the `swr` paradigm.
> 
> @chefilip We need to refresh the cache when user interacts with their own profile, Imho mutate is the safest option. What should we apply instead?

@memoyil  

Don't get me wrong. The way you do it is good. It's just about the whole data flow is quite similar to Redux but in the mutable way.

memoyil:
> > > In general, it looks great. One high-level concern is that using `mutate` like `dispatch` seems not fit well with the `swr` paradigm.
> > 
> > 
> > @chefilip We need to refresh the cache when user interacts with their own profile, Imho mutate is the safest option. What should we apply instead?
> 
> @memoyil
> 
> Don't get me wrong. The way you do it is good. It's just about the whole data flow is quite similar to Redux but in the mutable way.

I agree 👍 , we could also wait for auto revalidation when user changes something on profile but it is better for user to see the changes immediately on ui

chef-jojo (approved):

PR #3171, opened 2022-02-03
refactor: (bounty): Use props instead of hooks inside modal to reduce calculation and memory

memoyil:


src/views/Pools/components/BountyCard.tsx
```diff
@@ -60,7 +61,17 @@ const BountyCard = () => {
...
+  const [onPresentBountyModal] = useModal(
+    <BountyModal
+      estimatedCakeBountyReward={estimatedCakeBountyReward}
+      totalPendingCakeHarvest={totalPendingCakeHarvest}
+      callFee={callFee}
+      cakePriceBusd={cakePriceBusd}
+      TooltipComponent={TooltipComponent}
+    />,
+    true,
+    true,
```
memoyil, line 73: It will rerender the modal when props change

chef-jojo (approved):

PR #3178, opened 2022-02-06
fix: (prediction): Bump up swiper and reduce memory and use same amount of swiper cards

### LINKED ISSUE #2783: [BUG] Firefox / TOR extreme memory leaks

### Product

Prediction

### Current Behavior

Extreme memory leaks on clean Firefox / TOR anywhere from 1.2 to 20 Gb of RAM, renders whole thing completely unusable, can happen anywhere from 10 minutes to 24 hours of use.

After 5-10 minutes Predictions page becomes very hard to use, constantly just hangs browser.

Also happens for swap page as well.

### Expected Behavior

I want such service to work like a clock.

### Steps To Reproduce

1. Open Firefox / TOR with just Metamask and Predictions tab opened
2. Wait. You can just leave it for 10 minutes - 24 hours.

### Environment

```markdown
- Device: Any
- OS: Arch Linux
- Browser:Firefox / TOR
- Wallet: Metamask
```

### Anything else?

Please fix it, it drives me insane :upside_down_face: 
Perhaps it could be caused by constant `console.log`s?

memoyil:
Attempt to fix #2783


src/views/Nft/market/Collection/IndividualNFTPage/shared/MoreFromThisCollection.tsx
```diff
@@ -1,7 +1,10 @@
 import React, { useState, useEffect, useMemo } from 'react'
 import shuffle from 'lodash/shuffle'
 import styled from 'styled-components'
+// eslint-disable-next-line import/no-unresolved
 import { Swiper, SwiperSlide } from 'swiper/react'
+// eslint-disable-next-line import/no-unresolved
```
memoyil, line 6: See https://github.com/import-js/eslint-plugin-import/issues/2266

chefilip:
Thanks for great work @memoyil. 

I know it is a hard question. Do we know the root cause of this issue? It's quite difficult if we provide a solution without understanding the problem.

chef-jojo:
It's difficult to say. 
From the time I encountered the issue (hangs on prod) most likely fixed by https://github.com/pancakeswap/pancake-frontend/pull/3081
And there are couples possible issues
1. Crash when no current round https://github.com/pancakeswap/pancake-frontend/pull/3232/files#diff-a0654c309ce6273881749b35204235f8d88cbe8e7d55c461295b30591bf7e9b8R138-R141
2. Too many cards: which should be fixed in this PR

memoyil:
> Thanks for great work @memoyil.
> 
> I know it is a hard question. Do we know the root cause of this issue? It's quite difficult to measure the result if we provide a solution without understanding the problem.

No, I cannot reproduce on my machine, but i tried to mitigate the memory usage as much as possible

memoyil:
> Screen Shot 2022-03-02 at 20 17 13
> 
> I got this error when trying to swipe on mobile ( chrome simulation )

Fixed

src/state/predictions/index.ts
```diff
@@ -498,7 +498,15 @@ export const predictionsSlice = createSlice({
 
     // Get multiple rounds
     builder.addCase(fetchRounds.fulfilled, (state, action) => {
-      state.rounds = merge({}, state.rounds, action.payload)
+      const allRoundData = merge({}, state.rounds, action.payload)
+      // Take always last PAST_ROUND_COUNT items
+      state.rounds = Object.entries(allRoundData)
```
chefilip, line 503: You could try `pickBy` lodash for this case

src/views/Predictions/hooks/useOnNextRound.ts
```diff
@@ -2,29 +2,43 @@ import { useEffect } from 'react'
...
   useEffect(() => {
-    if (swiper && currentEpoch !== undefined && previousEpoch !== undefined && currentEpoch !== previousEpoch) {
-      const currentEpochIndex = rounds.findIndex((round) => round.epoch === currentEpoch)
+    if (
+      swiper &&
+      currentEpoch !== undefined &&
+      previousEpoch !== undefined &&
+      (currentEpoch !== previousEpoch || roundsEpochsString !== previousRoundsEpochsString)
+    ) {
+      const currentEpochIndex = orderBy(Object.values(rounds), ['epoch'], ['asc']).findIndex(
```
chefilip, line 33: Am I missing something here ?

I don't see the optimization value when adding more logic in this hook.

src/views/Predictions/components/RoundCard/LiveRoundPrice.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { Skeleton, TooltipText } from '@pancakeswap/uikit'
+import { NodeRound } from 'state/types'
+import { formatBigNumberToFixed } from 'utils/formatBalance'
+import usePollOraclePrice from '../../hooks/usePollOraclePrice'
+
+interface LiveRoundPriceProps {
+  round: NodeRound
+}
+
+const LiveRoundPrice: React.FC<LiveRoundPriceProps> = ({ round }) => {
```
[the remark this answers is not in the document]
memoyil: Done

src/views/Predictions/hooks/useOnNextRound.ts
memoyil: If we don't put that extra roundsEpochsString check, swiper cannot jump on the new current card after the oldest one is removed. (Current epoch is changed before card removal and there would be no extra trigger to run this effect again)

src/views/Predictions/components/RoundCard/LiveRoundPrice.tsx
```diff
@@ -0,0 +1,38 @@
+import React, { useEffect, useRef } from 'react'
+import { useCountUp } from 'react-countup'
+import { Skeleton, TooltipText } from '@pancakeswap/uikit'
+import { NodeRound } from 'state/types'
+import { formatBigNumberToFixed } from 'utils/formatBalance'
```
chefilip, line 5: eslint

src/views/Predictions/components/RoundCard/LiveRoundPrice.tsx
```diff
@@ -0,0 +1,38 @@
...
+  }, [priceAsNumber, updateRef])
+
+  return (
+    <TooltipText bold color={priceColor} fontSize="24px" style={{ minHeight: '36px' }}>
+      {price.gt(0) ? `$${countUp}` : <Skeleton height="36px" width="94px" />}
+    </TooltipText>
+  )
+}
+
+export default LiveRoundPrice
```
chefilip, line 38: I think we could `memo` this component

src/views/Predictions/components/RoundCard/LiveRoundPrice.tsx
memoyil: Done

src/views/Predictions/components/RoundCard/LiveRoundPrice.tsx
memoyil: Done

chefilip:
Hi @memoyil !

Could you test this preview link on mobile ? It seems not swiping smoothly like before. 

I tested on metamask browser. The production version works smoother than this PR.

PR #3184, opened 2022-02-08
feat: PAN-128 country warning vercel

chefilip:


src/components/BlockCountry/index.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from 'react'
+import { useModal } from '@pancakeswap/uikit'
+import Cookies from 'js-cookie'
+import { BLOCK_COUNTRIES_COOKIE_NAME } from 'config/constants/cookie-names'
+import BlockCountryModal from './BlockCountryModal'
+
+function BlockCountry() {
+  const [showModal] = useModal(<BlockCountryModal />, false)
+  const showModalRef = React.useRef(showModal)
```
Chef-Cheems, line 9: Why use ref here and not just showModal? 🤔

package.json
```diff
@@ -33,22 +33,23 @@
...
+    "@ethersproject/providers": "^5.5.2",
+    "@ethersproject/strings": "^5.5.0",
     "@ethersproject/units": "^5.5.0",
+    "@next/bundle-analyzer": "^12.0.7",
     "@pancakeswap/sdk": "^2.4.0",
     "@pancakeswap/uikit": "^0.63.1",
     "@reduxjs/toolkit": "^1.5.0",
     "@sentry/nextjs": "^6.17.4",
     "@snapshot-labs/snapshot.js": "^0.3.25",
+    "@types/js-cookie": "^3.0.1",
```
Chef-Cheems, line 52: Please install it to dev dependencies :)

src/components/BlockCountry/index.tsx
chefilip: Oh because the `showModal` from `useModal` is not memoized, it would cause the `useEffect` below unnecessarily re-executed.

By `useRef`, the `useEffect` below would only be executed once.

Chef-Cheems (approved):

PR #3187, opened 2022-02-08
refactor: (contract): Use simple rpc provider for contract calls that making no user transaction

memoyil:


src/views/Nft/market/components/BuySellModals/BuyModal/index.tsx
```diff
@@ -44,7 +44,8 @@ const BuyModal: React.FC<BuyModalProps> = ({ nftToBuy, onDismiss }) => {
   const { callWithGasPrice } = useCallWithGasPrice()
 
   const { account } = useWeb3React()
-  const wbnbContract = useERC20(tokens.wbnb.address)
+  const wbnbContractReader = useERC20(tokens.wbnb.address, false)
```
Chef-Cheems, line 47: Not sure if there is much point of doing it here, since only connected users see this.
makes code a bit harder to read 🤷 
I think it also doesn't have any performance improvements 🤔 
maybe let's roll back this change?

memoyil: My intention was to make allowance check is done through pcs nodes, and approval is done by wallet rpc nodes. It might bring some ux performance if wallet rpc node is unstable/slow.

Chef-Cheems: hm, didn't think about it, good stuff 👍

Chef-Cheems (approved):

PR #3209, opened 2022-02-09
refactor: (bounty): Use props instead of hooks inside modal to reduce calculation and memory

memoyil:


src/views/Pools/components/BountyCard.tsx
```diff
@@ -60,7 +61,18 @@ const BountyCard = () => {
...
+    <BountyModal
+      estimatedCakeBountyReward={estimatedCakeBountyReward}
+      totalPendingCakeHarvest={totalPendingCakeHarvest}
+      callFee={callFee}
+      cakePriceBusd={cakePriceBusd}
+      TooltipComponent={TooltipComponent}
+    />,
+    true,
+    true,
+    'bountyModal',
```
memoyil, line 74: This is the missing part that cause refresh the other modal instances with this instance component. Thanks @chef-jojo

chef-jojo (approved):

PR #3215, opened 2022-02-10
fix: (nft): Nft Media with video not shown when low power mode enabled in mobile

memoyil:
To reproduce: 

1. Set your phone to low power mode (iphone preferred)
2. Go to nft activity
3. Set collection to pancake bunnies (this is where the most video media )
4. See thumbnails are not shown

### DIFF
--- a/src/views/Nft/market/components/NFTMedia.tsx
+++ b/src/views/Nft/market/components/NFTMedia.tsx
@@ -3,6 +3,9 @@ import { FC, useEffect, useRef } from 'react'
 import useIntersectionObserver from 'hooks/useIntersectionObserver'
 import { NftToken } from 'state/nftMarket/types'
 import styled from 'styled-components'
+import { useTryVideoNftMedia } from 'state/nftMarket/hooks'
+import { useAppDispatch } from 'state'
+import { setTryVideoNftMedia } from 'state/nftMarket/reducer'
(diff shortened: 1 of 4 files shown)


src/views/Nft/market/components/NFTMedia.tsx
```diff
@@ -26,18 +26,21 @@ const NFTMedia: FC<
 > = ({ width, height, nft, borderRadius = 'default', as, ...props }) => {
   const { observerRef, isIntersecting } = useIntersectionObserver()
   const vidRef = useRef(null)
+  const [fallbackToImage, setFallbackToImage] = useState(false)
 
   useEffect(() => {
     if (vidRef.current) {
       if (isIntersecting) {
```
chef-jojo, line 33: better add `&& fallbackToImage` ?

chef-jojo:
Can we have a global variable to store this error and skip retry play again? 
Not sure if we can catch error that is only not allowed. I think normal users won't toggle low power mode within our session

memoyil:
> Can we have a global variable to store this error and skip retry play again? Not sure if we can catch error that is only not allowed. I think normal users won't toggle low power mode within our session

Done

> Can we have a global variable to store this error and skip retry play again? Not sure if we can catch error that is only not allowed. I think normal users won't toggle low power mode within our session

Done

chef-jojo (approved):

PR #3220, opened 2022-02-10
fix: Voting cache on not closed state

chef-jojo:


memoyil:
Can proposal metadata be changed after published?

Chef-Cheems (approved):

memoyil:
> Can proposal metadata be changed after published?

Yes we needed for the state 👍

PR #3230, opened 2022-02-11
refactor: Remove block state from redux

memoyil:


chef-jojo:
I thought about it too. We have some actions get block number directly from `getState().block`. Need to handle it as well

src/state/pools/hooks.ts
```diff
@@ -23,21 +23,24 @@ import {
...
       await dispatch(fetchFarmsPublicDataAsync(activeFarms.map((farm) => farm.pid)))
       batch(() => {
-        dispatch(fetchPoolsPublicDataAsync())
+        dispatch(fetchPoolsPublicDataAsync(currentBlock))
         dispatch(fetchPoolsStakingLimitsAsync())
       })
     }
 
     fetchPoolsDataWithFarms()
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
memoyil, line 43: We have to use to not trigger this effect by block number

memoyil:
> I thought about it too. We have some actions get block number directly from `getState().block`. Need to handle it as well

Done

src/state/pools/hooks.ts
chef-jojo: I had a thought about avoiding this disable line. Haven't test it yet maybe broken :P  
```ts
// useRefreshEffect.ts
type BlockEffectCallback = (blockNumber: number) => ReturnType<EffectCallback>

export function useFastRefreshEffect(effect: BlockEffectCallback, deps?: DependencyList) {
  const { data = 0 } = useSWR<number>([FAST_INTERVAL, 'blockNumber'])
  const depsMemo = useMemo(() => [data, ...(deps || [])], [data, deps])

  // eslint-disable-next-line react-hooks/exhaustive-deps
  useEffect(effect.bind(null, data), depsMemo)
}

useFastRefreshEffect((blockNumber) => {}, [])
```

chef-jojo (approved):

src/hooks/useRefreshEffect.ts
```diff
@@ -2,18 +2,20 @@ import { FAST_INTERVAL, SLOW_INTERVAL } from 'config/constants'
...
-export function useFastRefreshEffect(effect: EffectCallback, deps?: DependencyList) {
+type BlockEffectCallback = (blockNumber: number) => ReturnType<EffectCallback>
+
+export function useFastRefreshEffect(effect: BlockEffectCallback, deps?: DependencyList) {
   const { data = 0 } = useSWR([FAST_INTERVAL, 'blockNumber'])
   const depsMemo = useMemo(() => [data, ...(deps || [])], [data, deps])
 
   // eslint-disable-next-line react-hooks/exhaustive-deps
-  useEffect(effect, depsMemo)
+  useEffect(effect.bind(null, data), depsMemo)
```
chefilip, line 12: Hmm. This current memo isn't right. For `useFastRefreshEffect`, `effect` should only be called when `blockNumber` is changed. However, with the current logic, `effect` would be called if `deps` are changed even though `data` might not be changed.

chefilip: I think it should be 
```
const callback = useCallback((data) => effect(data), [deps])

useEffect(() => callback(data), [data])
```

chef-jojo: If `deps` changed, the effect should be called again, some effect depends on `account`

chefilip: I think it should be 
```
const callback = useCallback((data) => effect(data), [deps])

useEffect(() => callback(data), [data])
```

src/views/Home/hooks/useGetTopPoolsByApr.tsx
```diff
@@ -60,6 +62,7 @@ const useGetTopPoolsByApr = (isIntersecting: boolean) => {
     if (isIntersecting && fetchStatus === FetchStatus.Idle) {
       fetchPoolsPublicData()
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [dispatch, setFetchStatus, fetchStatus, topPools, isIntersecting])
```
chefilip, line 66: add `currentBlock` in deps

chefilip, line 66: add `currentBlock` in deps

src/hooks/useRefreshEffect.ts
chef-jojo: It wont't run effect when `callback` changes, no?
Maybe a good chance to introduce some react hook testing there 😅

chefilip: It won't because deps of `useEffect` don't have `callback`. The example isn't good as well because it omits the `callback` in dep while using it. But the logic should be similar to this. 

Totally agreed. we should have Unit Test on hooks

src/views/Home/hooks/useGetTopPoolsByApr.tsx
chef-jojo: it's intentional. Let's use `intitalBlock`

memoyil: Done

src/hooks/useRefreshEffect.ts
memoyil: I think we should stringify the dependencies before use it in memo, otherwise it will create a new array on each render which cause refresheffects to be triggered. I did the fix seems working, can you review @chef-jojo @chefilip ?

chef-jojo: IMO, I don't prefer stringify especially in this hook, we don't know what kind of `deps` pass into. It won't work for function and circular ref

memoyil: True

memoyil: So we want effect to be called when one of its dependencies changes right?

src/state/pools/hooks.ts
```diff
@@ -23,22 +23,28 @@ import {
...
+        await dispatch(fetchFarmsPublicDataAsync(activeFarms.map((farm) => farm.pid)))
+        batch(() => {
+          dispatch(fetchPoolsPublicDataAsync(currentBlock))
+          dispatch(fetchPoolsStakingLimitsAsync())
+        })
+      }
 
-    fetchPoolsDataWithFarms()
-  }, [dispatch])
+      if (currentBlock > 0) {
```
chef-jojo, line 42: If it is depends on currentBlock, which will be `0` at first and wait for slow interval. We should handle it

src/state/pools/index.ts
```diff
@@ -134,13 +134,12 @@ export const fetchCakePoolUserDataAsync = (account: string) => async (dispatch)
...
+export const fetchPoolsPublicDataAsync = (currentBlockNumber: number) => async (dispatch, getState) => {
   try {
     const blockLimits = await fetchPoolsBlockLimits()
     const totalStakings = await fetchPoolsTotalStaking()
     const profileRequirements = await fetchPoolsProfileRequirement()
-    let currentBlock = getState().block?.currentBlock
-
+    let currentBlock = currentBlockNumber
     if (!currentBlock) {
       currentBlock = await simpleRpcProvider.getBlockNumber()
```
chefilip, line 144: Ideally, we should only use `currentBlock` from swr as a source of truth. I think we could remove `initialBlock` by using `currentBlock` as condition. For example
```
if (!currentBlock) {
  fetchPoolsPublicDataAsync(currentBlock)
}
```

chef-jojo: How do you get `currentBlock` if no `currentBlock` 😆

chefilip: Fixed it. Wrong logic. :)

chefilip: Got it. Thanks

memoyil: But I can remove that fallback and rely on the parameter. What do you think ?

src/hooks/useSubgraphHealth.ts
```diff
@@ -54,8 +54,6 @@ const useSubgraphHealth = () => {
           `,
         )
 
-        const currentBlock = await simpleRpcProvider.getBlockNumber()
-
```
memoyil, line 58: we don't need to check current block, after @chef-jojo 's branch merged to this branch

src/state/pools/index.ts
chef-jojo: Let's keep it. The block number is `0` at first time

chefilip (approved):

chef-jojo (approved):

PR #3232, opened 2022-02-11
feat(prediction): Chainlink chart

chef-jojo:
Add chainlink data feed chart
CleanShot 2022-02-14 at 11 01 37@2x
CleanShot 2022-02-14 at 11 02 18@2x


src/views/Predictions/MobileChart.tsx
```diff
@@ -0,0 +1,59 @@
...
+    <Flex flexDirection="column" height="100%">
+      <MenuWrapper>
+        <Menu />
+      </MenuWrapper>
+      <TabToggleGroup>
+        <TabToggle
+          isActive={chartView === PredictionsChartView.TradingView}
+          onClick={() => dispatch(setChartView(PredictionsChartView.TradingView))}
+        >
+          TradingView {t('Chart')}
```
memoyil, line 39: Nice to have 'Chart %symbol%' translation

src/views/Predictions/hooks/usePollOraclePrice.ts
```diff
@@ -1,28 +1,16 @@
...
-import { setLastOraclePrice } from 'state/predictions'
-import useGetLatestOraclePrice from './useGetLatestOraclePrice'
+import { useChainlinkOracleContract } from 'hooks/useContract'
+import { useSWRContract } from 'hooks/useSWRContract'
+import { Zero } from '@ethersproject/constants'
 
 const usePollOraclePrice = (seconds = 10) => {
-  const { price, refresh } = useGetLatestOraclePrice()
-  const dispatch = useAppDispatch()
+  const chainlinkOracleContract = useChainlinkOracleContract()
```
memoyil, line 6: Should be without signer

src/views/Predictions/components/Label.tsx
```diff
@@ -89,7 +90,7 @@ const Label = styled(Flex)<{ dir: 'left' | 'right' }>`
 `
 
 export const PricePairLabel: React.FC = () => {
-  const price = useGetLastOraclePrice()
+  const price = usePollOraclePrice()
   const priceAsNumber = parseFloat(formatBigNumberToFixed(price, 3, 8))
   const { countUp, update } = useCountUp({
```
memoyil, line 95: this hook makes this component render too much when price changes, which calls usePollOraclePrice() hook too many times in one second. Although it probably will not make a onchain request in short time, retrieving data from cache make performance suffer?

chef-jojo: I prefer not to split fetch hook and get cache hook here.I will add `dedupingInterval` to prevent some edge cases.

Actually I thought we could use event listener to do that, but it seems like the event is not implement in contract side

src/views/Predictions/Desktop.tsx
```diff
@@ -155,28 +156,57 @@ const Desktop: React.FC = () => {
...
+                  toggleChartPane()
+                  dispatch(setChartView(PredictionsChartView.Chainlink))
+                  e.stopPropagation()
+                  e.preventDefault()
+                }}
+              >
+                {chartView === PredictionsChartView.Chainlink && <ChartIcon mr="10px" />} Chainlink {t('Chart')}
+              </TabToggle>
+            </ExpandButtonGroup>
+            <ChartByLabel
```
chefilip, line 198: Remove the label if there is no link.

chefilip (commented):
Great work !

Chef-Cheems (approved):

PR #3235, opened 2022-02-11
chore: Remove sentry txn trace

chef-jojo:
Remove sentry performance check


chefilip:
It makes sense because we already have Vercel to check that

Chef-Cheems (approved):

PR #3244, opened 2022-02-11
feat: (nft): New look for collections page

memoyil:
To review:

1. Go to /nfts/collections page 
2. It uses collections cards in grid instead of table


chef-jojo:
@memoyil I think we can shuffle the collection only one time within a session.

memoyil:
> @memoyil I think we can shuffle the collection only one time within a session.

Then I would wait redux collection removal pr to be merged to store shuffled collection in swr

memoyil:
> @memoyil I think we can shuffle the collection only one time within a session.

Done, I will change the implementation when we merge collection removal pr.

chef-jojo (approved):

PR #3259, opened 2022-02-14
fix: (ifo): Ifo private sale tooltip li position

memoyil:
Before: 

After:


chefilip:
Thanks for your git. LGTM

chef-jojo (approved):

PR #3260, opened 2022-02-14
feat: (pool): Add auto earnings interest breakdown to pool cards

memoyil:
To review:

1. Go to pool cards
2. See tooltip of recent cake profit value  has auto earnings interest breakdown like pool table rows

### DIFF
--- a/src/views/Pools/components/CakeVaultCard/index.tsx
+++ b/src/views/Pools/components/CakeVaultCard/index.tsx
@@ -124,7 +124,7 @@ const CakeVaultCard: React.FC<CakeVaultProps> = ({ pool, showStakedOnly, default
         <FlexGap mt="16px" gap="24px" flexDirection={accountHasSharesStaked ? 'column-reverse' : 'column'}>
           <Box>
             <Box mt="24px">
-              <RecentCakeProfitRow vaultKey={pool.vaultKey} />
+              <RecentCakeProfitRow pool={pool} />
             </Box>
             <Box mt="8px">
               <UnstakingFeeCountdownRow vaultKey={pool.vaultKey} />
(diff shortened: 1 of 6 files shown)


src/views/Pools/components/CakeVaultCard/index.tsx
```diff
@@ -125,7 +125,7 @@ const CakeVaultCard: React.FC<CakeVaultProps> = ({ pool, showStakedOnly, default
         <FlexGap mt="16px" gap="24px" flexDirection={accountHasSharesStaked ? 'column-reverse' : 'column'}>
           <Box>
             <Box mt="24px">
-              <RecentCakeProfitRow vaultKey={pool.vaultKey} />
+              <RecentCakeProfitRow pool={pool} />
```
chefilip, line 128: Hmm. I understand that why we need to pass `pool`. However, the more components use this pattern, the more re-re-render it needs.

I don't have a solution yet but it would be good when we are aware of this.

memoyil: We only have 2 cakevault card on runtime if user doesn't enable staked only pools. So there should be no additional rerender if we use pool as props. Am i missing something?

chefilip: Oh. passing the `pool` object is convenient but it's hard to manage re-rendering. We don't know which properties are consumed by component. For example, If other devs add a new attribute which changes frequently in the pool prop, which leads to components using the `pool` prop re-rendered frequently even though components don't use the new attribute.

Nothing wrong here. I will think about how we could mitigate this risk.

chefilip (approved):

src/views/Pools/components/CakeVaultCard/index.tsx
memoyil: This I agree 👍  but afaik unless we use react memo it will be rerendered each time even if we pass the same referenced props. So if pool update frequency increased in the future it will affect all components even if vaultKey used in props only.

PR #3179, opened 2022-02-07
feat: add wrong network and pending message to menu

ChefMomota:

PR #3270, opened 2022-02-16
feat: Only attempt to setup network once

chef-jojo:
Since we already have Wrong Network Indicator https://github.com/pancakeswap/pancake-frontend/pull/3179
We don't need to keep asking for `setupNetwork` each time when navigating the pages.

### DIFF
--- a/src/contexts/ToastsContext/Provider.tsx
+++ b/src/contexts/ToastsContext/Provider.tsx
@@ -30,22 +30,38 @@ export const ToastsProvider: React.FC = ({ children }) => {
     [setToasts],
   )
 
-  const toastError = (title: Toast['title'], description?: Toast['description']) => {
-    return toast({ title, description, type: toastTypes.DANGER })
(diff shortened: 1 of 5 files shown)


src/contexts/ToastsContext/Provider.tsx
```diff
@@ -30,22 +30,38 @@ export const ToastsProvider: React.FC = ({ children }) => {
...
+      return toast({ title, description, type: toastTypes.WARNING })
+    },
+    [toast],
+  )
+
+  const clear = useCallback(() => setToasts([]), [])
+  const remove = useCallback((id: string) => {
     setToasts((prevToasts) => prevToasts.filter((prevToast) => prevToast.id !== id))
-  }
+  }, [])
```
chef-jojo, line 64: This helps a lot when we have `toastError` or etc in useEffect deps. And this is the reason why we keep asking for request network change

PR #3271, opened 2022-02-16
fix(nft): Individual page crash

chef-jojo:


src/views/Nft/market/hooks/useCompleteNft.tsx
```diff
@@ -78,7 +83,7 @@ export const useCompleteNft = (collectionAddress: string, tokenId: string) => {
   }, [mutate, refetchNftMarketData, refetchNftOwn])
 
   return {
-    combinedNft: nft ? { ...nft, marketData } : undefined,
+    combinedNft: nft ? { ...nft, marketData, location: nftOwn?.location } : undefined,
```
memoyil, line 86: with this approach location will be undefined at first, may be it is better to set with default data?

memoyil (approved):

Chef-Cheems (approved):

PR #3274, opened 2022-02-16
fix(ifo 3.1): Fix bugs in Live and Finished phases

chefilip:
Fix following bugs: 
- Should show Claim button when ineligible users need to claim
- Should not minus committed credit in Private Sale from IFO credits
- Should show Commit button in Live mode when user has not enabled the pool yet

### DIFF
--- a/src/views/Ifos/components/IfoFoldableCard/IfoPoolCard/IfoCardActions.tsx
+++ b/src/views/Ifos/components/IfoFoldableCard/IfoPoolCard/IfoCardActions.tsx
@@ -53,6 +53,16 @@ const IfoCardActions: React.FC<Props> = ({
     )
   }
 
+  const needClaim =
+    publicIfoData.status === 'finished' &&
+    !userPoolCharacteristics.hasClaimed &&
+    (userPoolCharacteristics.offeringAmountInToken.isGreaterThan(0) ||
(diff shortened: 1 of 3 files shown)


src/views/Ifos/components/IfoFoldableCard/IfoPoolCard/IfoCardActions.tsx
```diff
@@ -53,7 +53,20 @@ const IfoCardActions: React.FC<Props> = ({
     )
   }
 
-  if (enableStatus !== EnableStatus.ENABLED || (ifo.version === 3.1 && poolId === PoolIds.poolBasic && !isEligible)) {
+  const needClaim =
```
memoyil, line 56: Maybe use 

` if (publicIfoData.status === 'coming_soon') { ... `

which is easier to read and reduce calculation when ifo status is not finished

chefilip: Good idea. I will create another PR with Unit Test to refactor this after the IFO 3.1 is released

memoyil: 👍

chef-jojo (approved):

PR #3275, opened 2022-02-16
chore: Remove import react

memoyil:


src/components/EasterEgg/EasterEgg.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useState, useCallback } from 'react'
+import { useState, useCallback, memo } from 'react'
 import { FallingBunnies, FallingBunniesProps, useKonamiCheatCode } from '@pancakeswap/uikit'
 
 const EasterEgg: React.FC<FallingBunniesProps> = (props) => {
```
chef-jojo, line 4: I didn't know we can do this 🤯 maybe types import globally

memoyil: React namespace is imported thats why we can use without import

chef-jojo (changes requested):
fix build

memoyil:
> fix build

Done

chef-jojo (approved):

PR #3279, opened 2022-02-16
fix: (profile): No Nft warning in Change profile modal

memoyil:
Fixed issues:

Fixes no nft warning in modal is shown after updating the profile picture 
Fixes no nft warning is shown in modal when inital load instead of skeleton
Fixes no nft warning not shown when nft wallet doesnt have nfts but has on sale or as profile picture

### DIFF
--- a/src/views/Nft/market/Profile/components/EditProfileModal/ChangeProfilePicView.tsx
+++ b/src/views/Nft/market/Profile/components/EditProfileModal/ChangeProfilePicView.tsx
@@ -1,5 +1,5 @@
 import { useState, useMemo } from 'react'
-import { Button, Box, InjectedModalProps, Text } from '@pancakeswap/uikit'
+import { Button, Box, InjectedModalProps, Text, Skeleton } from '@pancakeswap/uikit'
 import { useWeb3React } from '@web3-react/core'
 import { useProfile } from 'state/profile/hooks'
 import { useTranslation } from 'contexts/Localization'
@@ -28,7 +28,7 @@ const ChangeProfilePicPage: React.FC<ChangeProfilePicPageProps> = ({ onDismiss, 
   const { t } = useTranslation()
   const { account, library } = useWeb3React()
   const { isLoading: isProfileLoading, profile, refresh: refreshProfile } = useProfile()
-  const { nfts } = useNftsForAddress(account, profile, isProfileLoading)
+  const { nfts, isLoading } = useNftsForAddress(account, profile, isProfileLoading)
   const profileContract = useProfileContract()
   const { toastSuccess } = useToast()
   const { callWithGasPrice } = useCallWithGasPrice()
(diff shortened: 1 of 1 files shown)


src/views/Nft/market/Profile/components/EditProfileModal/ChangeProfilePicView.tsx
```diff
@@ -61,13 +61,13 @@ const ChangeProfilePicPage: React.FC<ChangeProfilePicPageProps> = ({ onDismiss,
         return callWithGasPrice(profileContract, 'updateProfile', [selectedNft.collectionAddress, selectedNft.tokenId])
       },
       onSuccess: async ({ receipt }) => {
-        // Re-fetch profile
-        await dispatch(fetchProfile(account))
```
memoyil, line 65: No need to wait for the result of fetchProfile

src/views/Nft/market/Profile/components/EditProfileModal/ChangeProfilePicView.tsx
```diff
@@ -61,13 +61,13 @@ const ChangeProfilePicPage: React.FC<ChangeProfilePicPageProps> = ({ onDismiss,
...
+        onDismiss()
```
memoyil, line 64: We should dismiss the modal first from ui otherwise user would see skeleton in this version or warning in current deployed version (since nfts will first return empty array when retriggered)

src/views/Nft/market/Profile/components/EditProfileModal/ChangeProfilePicView.tsx
```diff
@@ -78,30 +78,33 @@ const ChangeProfilePicPage: React.FC<ChangeProfilePicPageProps> = ({ onDismiss,
...
-              disabled={isApproving || isConfirming || isConfirmed}
-            >
-              <Text bold>{walletNft.name}</Text>
-            </SelectionCard>
-          )
-        })}
-      </Box>
-      {nfts.length === 0 && (
+      {isLoading ? (
+        <Skeleton width="100%" height="80px" mb="16px" />
```
memoyil, line 82: We should add skeleton first otherwise user sees warning about no eligible nfts when nfts are loading

src/views/Nft/market/Profile/components/EditProfileModal/ChangeProfilePicView.tsx
```diff
@@ -78,30 +78,33 @@ const ChangeProfilePicPage: React.FC<ChangeProfilePicPageProps> = ({ onDismiss,
...
-              isChecked={walletNft.tokenId === selectedNft.tokenId}
-              onChange={handleChange}
-              disabled={isApproving || isConfirming || isConfirmed}
-            >
-              <Text bold>{walletNft.name}</Text>
-            </SelectionCard>
-          )
-        })}
-      </Box>
-      {nfts.length === 0 && (
```
memoyil, line 104: Root cause of the issue is here, it uses nfts instead of nftsInWallet to determine showing warnings

src/views/Nft/market/hooks/useNftsForAddress.tsx
```diff
@@ -26,13 +27,11 @@ const useNftsForAddress = (account: string, profile: Profile, isProfileFetching:
...
-    if (!isProfileFetching && !isEmpty(collections) && isAddress(account)) {
-      return getCompleteAccountNftData(account, collections, profileNftWithCollectionAddress)
-    }
-    return []
-  })
-
+  const { status, data, mutate } = useSWR(
+    !isProfileFetching && !isEmpty(collections) && isAddress(account) ? [account, 'userNfts'] : null,
+    async () => getCompleteAccountNftData(account, collections, profileNftWithCollectionAddress),
+    { use: [laggyMiddleware] },
```
chef-jojo, line 33: I think the reason we need laggy here is because `isProfileFetching` toggle key changes. Can't think of any better way yet but sure it's a problem

chef-jojo (approved):

PR #3286, opened 2022-02-17
fix: Add null check to onDismiss before calling it

memoyil:
onDismiss is nullable so we have to check it before calling


chef-jojo:
Feeling little verbose. Let's use onDismiss.?()

memoyil:
> Feeling little verbose. Let's use onDismiss.?()

Done

PR #3307, opened 2022-02-19
fix(ifo): Fix how does it work buttons

AGregorc:
**Steps to reproduce:**

1. Go to IFO Finished page `/ifo/history` 
2. Press on `How does it work?` button
3. The button doesn't do anything

Another link that doesn't work is in `/ifo/history` page on "Sale Finished!" card detail when any warning message is shown. By that, I mean in the `IfoCardTokens` component where the link was not working as well. 

This PR attempts to fix these link issues.


chef-jojo (approved):
Finally someone fix it. Thanks!

PR #3312, opened 2022-02-20
feat: Use next-themes

memoyil:


src/state/user/reducer.ts
```diff
@@ -69,6 +70,7 @@ export interface UserState {
   timestamp: number
   audioPlay: boolean
   isDark: boolean
```
memoyil, line 72: although it will not be used, keeping this to be ensure having not problems when migrating

src/state/user/reducer.ts
```diff
@@ -100,6 +102,7 @@ export const initialState: UserState = {
   timestamp: currentTimestamp(),
   audioPlay: true,
   isDark: false,
+  theme: null,
```
chef-jojo, line 105: Let's use `undefined` here. or else we have to add redux persist migration for any existing users

memoyil: Done

chefilip:
Could I ask what is the benefits of using theme system in redux state ?

memoyil:
> Could I ask what is the benefits of using theme system in redux state ?

Afaik, it is persisted to browsers local storage, we can do it manually though but redux makes it kind of easier i think

chefilip:
Because we replace a `boolean` state with an `object` Redux state. As you know,  with object state, there is a risk of accidentally introducing unnecessarily re-rendering if we don't use it carefully. I think the trade-off should be worth enough to replace.

memoyil:
> Because we replace a `boolean` state with an `object` Redux state. As you know, with object state, there is a risk of accidentally introducing unnecessarily re-rendering if we don't use it carefully. I think the trade-off should be worth enough to replace.

This I agree we should be careful, but since we are using enums to compare based on string and the only way to change that state is direct user interaction, I think we are safe. Also If we made changes on isDark property instead of introducing a new one, most of the users would not be able to see that feature unless they clear the cache.

src/state/user/hooks/index.tsx
```diff
@@ -139,15 +140,25 @@ export function useExpertModeManager(): [boolean, () => void] {
...
+export function useThemeManager(): [boolean, (theme: Theme) => void] {
   const dispatch = useDispatch<AppDispatch>()
-  const isDark = useSelector<AppState, AppState['user']['isDark']>((state) => state.user.isDark)
+  const theme = useSelector<AppState, AppState['user']['theme']>((state) => state.user.theme)
 
-  const toggleTheme = useCallback(() => {
-    dispatch(toggleThemeAction())
-  }, [dispatch])
+  const isDark = useMemo(() => {
+    if (typeof window !== 'undefined' && theme === null) {
```
chef-jojo, line 148: `!theme`

chef-jojo:
Since we will introduce a breaking change, theme setting will be gone after this move.
I want to bring this up and we could discuss about using this approach or we simply just use [next-themes](https://www.npmjs.com/package/next-themes):

So I'm targeting to remove redux persist gate if possible, the blocker of it is because of the first paint theme changing, which is styled component limitation.

To avoid the first paint flashing, we have to do 2 things:
1. Inject a script on the head to get theme from localStorage as soon as possible https://github.com/chef-jojo/pancake-frontend/compare/4fb7ff68bb08a5db75f5a2dba1a7b3709f41a95b..e9b053c1311fbdbb336a7f8e05d9cbd82514a1ba#diff-1c97cd72a45115615dc67dac74783fe3956b4ab2219db23115322a0ff114daffR34-R44
which is same as their upgraded version https://github.com/pacocoursey/next-themes/blob/master/src/index.tsx#L252-L256
2. replace styled-component or replace styled-component variable to CSS variable

The current approach is okay to me, it's less code change but it's harder to deal with when we want to inject script because the theme variable lives deep inside of a redux persist object.

src/state/user/hooks/index.tsx
memoyil: Done

chef-jojo:
Umm let's make a redux persist migration for it.

If isDark is true, we can safely set the new theme to dark

src/state/index.ts
```diff
@@ -18,13 +29,27 @@ import swap from './swap/reducer'
 import transactions from './transactions/reducer'
 import user from './user/reducer'
 
+const migrations = {
+  0: (state) => {
+    return {
+      ...state,
+      user: {
+        ...state.user,
+        theme: state.user.isDark ? Theme.DARK : Theme.LIGHT,
```
chef-jojo, line 38: Don't set to light, instead let default to be `undefined`, or you can change it back to `null`.
I think it's safe to take out isDark and no need to save it in next version 
`const { isDark, ...user } = state.user`

memoyil: Done

chef-jojo:
sorry looks like my suggestion has format error 😅 
other than that, all looks good to me

memoyil:
> sorry looks like my suggestion has format error 😅 other than that, all looks good to me

Your commit message has that, I fixed it but your commit author is gone even if i added to the message :(

chef-jojo:
> > sorry looks like my suggestion has format error 😅 other than that, all looks good to me
> 
> 
> 
> Your commit message has that, I fixed it but your commit author is gone even if i added to the message :(

I see. We may get rid of commitlint and in favor of only pr title lint only. Was looking at some good bot to help merge with PR description 👀

chef-jojo (dismissed):

memoyil:
> Since we will introduce a breaking change, theme setting will be gone after this move. I want to bring this up and we could discuss about using this approach or we simply just use [next-themes](https://www.npmjs.com/package/next-themes):
> 
> So I'm targeting to remove redux persist gate if possible, the blocker of it is because of the first paint theme changing, which is styled component limitation.
> 
> To avoid the first paint flashing, we have to do 2 things:
> 
>     1. Inject a script on the head to get theme from localStorage as soon as possible https://github.com/chef-jojo/pancake-frontend/compare/4fb7ff68bb08a5db75f5a2dba1a7b3709f41a95b..e9b053c1311fbdbb336a7f8e05d9cbd82514a1ba#diff-1c97cd72a45115615dc67dac74783fe3956b4ab2219db23115322a0ff114daffR34-R44
>        which is same as their upgraded version https://github.com/pacocoursey/next-themes/blob/master/src/index.tsx#L252-L256
> 
>     2. replace styled-component or replace styled-component variable to CSS variable
> 
> 
> The current approach is okay to me, it's less code change but it's harder to deal with when we want to inject script because the theme variable lives deep inside of a redux persist object.



> Since we will introduce a breaking change, theme setting will be gone after this move. I want to bring this up and we could discuss about using this approach or we simply just use [next-themes](https://www.npmjs.com/package/next-themes):
> 
> So I'm targeting to remove redux persist gate if possible, the blocker of it is because of the first paint theme changing, which is styled component limitation.
> 
> To avoid the first paint flashing, we have to do 2 things:
> 
>     1. Inject a script on the head to get theme from localStorage as soon as possible https://github.com/chef-jojo/pancake-frontend/compare/4fb7ff68bb08a5db75f5a2dba1a7b3709f41a95b..e9b053c1311fbdbb336a7f8e05d9cbd82514a1ba#diff-1c97cd72a45115615dc67dac74783fe3956b4ab2219db23115322a0ff114daffR34-R44
>        which is same as their upgraded version https://github.com/pacocoursey/next-themes/blob/master/src/index.tsx#L252-L256
> 
>     2. replace styled-component or replace styled-component variable to CSS variable
> 
> 
> The current approach is okay to me, it's less code change but it's harder to deal with when we want to inject script because the theme variable lives deep inside of a redux persist object.

Imho next-themes would be the most compact solution (i think it also has built-in system theme option) until we apply that we can use this solution 👍

chef-jojo:
Maybe we just do it? Should be pretty straight forward.
I'm not feeling comfortable to maintain the persist migration

memoyil:
> Maybe we just do it? Should be pretty straight forward. I'm not feeling comfortable to maintain the persist migration

Sure 👍 it is ready already right?

memoyil:
> Maybe we just do it? Should be pretty straight forward. I'm not feeling comfortable to maintain the persist migration

Done

chefilip:
CleanShot 2022-03-02 at 06 44 20

1. It seems the default theme is dark
2. For current users already set up dark theme, they would have to toggle it again, right ?
    - I'm wondering if it's worth to maintain the theme state for them. Or just ignore it

memoyil:
> CleanShot 2022-03-02 at 06 44 20
> 
> 1. It seems the default theme is dark
> 2. For current users already set up dark theme, they would have to toggle it again, right ?
>    
>    * I'm wondering if it's worth to maintain the theme state for them. Or just ignore it
> 
> Great work !

Default theme is system theme, I think it would be overkill to write a migration for the users who uses light theme in operating system and has dark theme in pcs. Wdyt?

chefilip:
Hi @memoyil  !

Could you help me resolve conflicts ? Thanks

src/pages/_app.tsx
```diff
@@ -11,7 +11,7 @@ import useSentryUser from 'hooks/useSentryUser'
 import useUserAgent from 'hooks/useUserAgent'
 import type { AppProps } from 'next/app'
 import Head from 'next/head'
-import { Fragment } from 'react'
+import { Fragment, useState, useEffect } from 'react'
```
chefilip, line 14: eslint

chef-jojo (dismissed):

src/pages/_app.tsx
memoyil: Done

PR #3328, opened 2022-02-22
feat: Limit orders

Chef-Cheems:
👁️


src/utils/localStorageOrders.ts
```diff
@@ -0,0 +1,121 @@
+import { Order } from '@gelatonetwork/limit-orders-lib'
+import { get, set, clear } from 'local-storage'
+
+const LS_ORDERS = 'gorders_'
```
memoyil, line 4: Can't we use usePersistState ?

src/hooks/limitOrders/useGelatoLimitOrdersHistory.ts
```diff
@@ -0,0 +1,168 @@
...
+    pending: Order[]
+    confirmed: Order[]
+  }>({ pending: [], confirmed: [] })
+  const [cancelledOrders, setCancelledOrders] = useState<{
+    pending: Order[]
+    confirmed: Order[]
+  }>({ pending: [], confirmed: [] })
+  const [executedOrders, setExecutedOrders] = useState<Order[]>([])
+
+  const transactions = useAllTransactions()
```
chef-jojo, line 33: can return useMemo in this hook, and better filter only limit orders tx

src/utils/localStorageOrders.ts
Chef-Cheems: yea I'll rewrite some of the localStorage logic today, some stuff is not working properly yet, its a draft PR 😅

src/hooks/limitOrders/useGelatoLimitOrdersLib.ts
```diff
@@ -0,0 +1,22 @@
...
+const useGelatoLimitOrdersLib = (): GelatoLimitOrders | undefined => {
+  const { chainId, library } = useActiveWeb3React()
+
+  return useMemo(() => {
+    if (!chainId || !library) {
+      console.error('Could not instantiate GelatoLimitOrders: missing chainId or library')
+      return undefined
+    }
+    try {
+      return new GelatoLimitOrders(chainId as ChainId, library?.getSigner(), 'pancakeswap', false)
```
chef-jojo, line 14: nit: Can we do this?
```tsx
const gelatoLimitOrdersRef = useRef<GelatoLimitOrders>()

useEffect(() => {
    if (!chainId || !library) {
      console.error('Could not instantiate GelatoLimitOrders: missing chainId or library')
      gelatoLimitOrdersRef.current = undefined
    }
    gelatoLimitOrdersRef.current = new GelatoLimitOrders(chainId as ChainId, library?.getSigner(), 'pancakeswap', false)

return () => {
      gelatoLimitOrdersRef.current = undefined
  }
}, [chainId, library])

return gelatoLimitOrdersRef.current
```

chef-jojo: maybe not worth to change nothing from here 🤭

src/state/limitOrders/reducer.ts
```diff
@@ -0,0 +1,70 @@
+import { createReducer } from '@reduxjs/toolkit'
+import { replaceLimitOrdersState, selectCurrency, typeInput, switchCurrencies, setRateType } from './actions'
+import { Field, Rate, OrderState } from './types'
+
+export const initialState: OrderState = {
+  independentField: Field.INPUT,
+  typedValue: '',
+  inputValue: '',
+  [Field.INPUT]: {
+    currencyId: '',
```
chefilip, line 10: I think `currencyAddress` would be more informative.

src/views/LimitOrders/components/LimitOrderTable/DetailLimitOrderModal.tsx
```diff
@@ -0,0 +1,252 @@
...
+            </Tag>
+          )}
+          {isCancellationPending && (
+            <Tag outline scale="sm" p="8px" mb="16px" variant="warning">
+              {t('Cancelling')}
+            </Tag>
+          )}
+          {isCancelled && !isCancellationPending && (
+            <Tag outline scale="sm" p="8px" mb="16px" variant="failure">
+              {t('Cancelled')}
```
chefilip, line 219: Currently we have `Canceled` word in translation. Should we use the new one with double `l` or use the existing one ?

src/state/transactions/reducer.ts
```diff
@@ -32,14 +37,18 @@ export const initialState: TransactionState = {}
...
+    .addCase(
+      addTransaction,
+      (transactions, { payload: { chainId, from, hash, approval, summary, claim, type, order } }) => {
+        if (transactions[chainId]?.[hash]) {
+          throw Error('Attempted to add existing transaction.')
+        }
+        const txs = transactions[chainId] ?? {}
+        txs[hash] = { hash, approval, summary, claim, from, addedTime: now(), type, order }
+        transactions[chainId] = txs
+        if (order) saveOrder(chainId, from, order, true)
```
chefilip, line 49: `saveOrder` is side-effect function. I think it shouldn't be in the reducer.

chef-jojo:
Can we merge input state with swap?
I think it's better ux when switching swap and limit order

src/views/LimitOrders/index.tsx
```diff
@@ -0,0 +1,428 @@
...
+
+  const showApproveFlow =
+    !inputError && (approvalState === ApprovalState.NOT_APPROVED || approvalState === ApprovalState.PENDING)
+
+  const inputCurrencyId =
+    currencies.input instanceof Token ? currencies.input.address : currencies.input === ETHER ? 'BNB' : ''
+  const outputCurrencyId =
+    currencies.output instanceof Token ? currencies.output.address : currencies.output === ETHER ? 'BNB' : ''
+
+  const userHasOrders = useMemo(
```
chefilip, line 255: if we keep this logic, the performance issue isn't easy to solve. The LimitOrder will keep re-rendering because of the interval. 

I would suggest we show the open table even when user has no order

src/views/LimitOrders/utils/getPriceForOneToken.ts
```diff
@@ -0,0 +1,25 @@
...
+  const fractionPerOneToken = outputAmount.asFraction
+    .divide(inputAmount.asFraction)
+    .divide(inputDecimals)
+    .multiply(outputDecimals)
+  return new Price(
+    inputAmount.currency,
+    outputAmount.currency,
+    fractionPerOneToken.denominator,
+    fractionPerOneToken.numerator,
+  )
```
chef-jojo, line 22: any different between this and `new Price(inputAmount, outputAmount, inputAmount.raw, outputAmount.raw)` ?

src/views/LimitOrders/components/LimitOrderTable/DetailLimitOrderModal.tsx
Chef-Cheems: Let's fix it later. I think in predictions some "canceled" strings might come from the status field and not just used as plain text 🥴

src/views/LimitOrders/index.tsx
Chef-Cheems: Thanks for the fix @chefilip 🙌

src/views/LimitOrders/utils/getPriceForOneToken.ts
Chef-Cheems: you're right 🙇

chef-jojo (approved):

chef-jojo (approved):

PR #3332, opened 2022-02-23
chore: Handle Error on getting vote power

Chef-Yogi:
demo

demo2

### DIFF
--- a/src/views/Voting/components/CastVoteModal/MainView.tsx
+++ b/src/views/Voting/components/CastVoteModal/MainView.tsx
@@ -1,4 +1,4 @@
-import { IconButton, Text, Skeleton, Button, AutoRenewIcon, ChevronRightIcon } from '@pancakeswap/uikit'
+import { IconButton, Text, Skeleton, Button, AutoRenewIcon, ChevronRightIcon, Message } from '@pancakeswap/uikit'
 import { useTranslation } from 'contexts/Localization'
 import { formatNumber } from 'utils/formatBalance'
 import TextEllipsis from '../TextEllipsis'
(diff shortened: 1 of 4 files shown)


src/views/Voting/hooks/useGetVotingPower.tsx
```diff
@@ -27,8 +29,10 @@ const initialState: State = {
 
 const useGetVotingPower = (block?: number, isActive = true): State & { isLoading: boolean } => {
   const { account } = useWeb3React()
+  const { toastError } = useToast()
   const [votingPower, setVotingPower] = useState(initialState)
   const [isLoading, setIsLoading] = useState(!!account)
+  const { t } = useTranslation()
 
   useEffect(() => {
     const fetchVotingPower = async () => {
```
memoyil, line 38: We can use swr instead

memoyil:
It should not be shown voting power 0 when there is an error, instead of toast error maybe it can be switched to error modal (since there is a visibility timeout on toast it can be missed by user)

chefilip:
fix eslint

src/views/Voting/components/CastVoteModal/MainView.tsx
```diff
@@ -41,8 +42,12 @@ const MainView: React.FC<MainViewProps> = ({
         <Text color="secondary" mb="8px" textTransform="uppercase" fontSize="12px" bold>
           {t('Your Voting Power')}
         </Text>
-        {isLoading ? (
+        {isLoading && !isError ? (
           <Skeleton height="64px" mb="24px" />
+        ) : isError ? (
+          <Text mb="12px" color="red">
+            {t('Error fetching Voting Power')}
```
chefilip, line 49: Hmm. The error is a good indicator but it's not informative. As a user, I don't know what the error I'm facing and what the next action I should take to resolve it.

Chef-Yogi: Yeah, useSWR will retry to fetch automatically. (maybe we can tell the user will retry to fetch automatically?)
And about the wording. I am discussing this with Chef Yuna, and she said she wants to reword this message.

PR #3333, opened 2022-02-23
feat(ifo): Add ComingSoon Section

chefilip:
Screen Shot 2022-02-23 at 15 27 08


src/views/Ifos/Ifo.tsx
```diff
@@ -0,0 +1,14 @@
+import { ifosConfig } from 'config/constants'
+import CurrentIfo from './CurrentIfo'
+import SoonIfo from './SoonIfo'
+
+/**
+ * Note: currently there should be only 1 active IFO at a time
+ */
+const activeIfo = ifosConfig.find((ifo) => ifo.isActive)
```
memoyil, line 8: We need to have extra deployment after the ifo to make it false, maybe it is better to calculate the end with blocks automatically wdyt?

chefilip: It's a good idea but the problem is we don't know when business chefs want to close the current 'Finished' IFO. I will ask and apply it later.

PR #3337, opened 2022-02-24
fix(voting): Handle voting error when api is not stable

chef-jojo:
When snapshot api is not stable, some CORS issue. we should handle it properly

### DIFF
--- a/src/pages/voting/proposal/[id].tsx
+++ b/src/pages/voting/proposal/[id].tsx
@@ -5,7 +5,7 @@ import { getProposal } from 'state/voting/helpers'
 import { ProposalState } from 'state/types'
 import ProposalPageRouter from 'views/Voting/Proposal'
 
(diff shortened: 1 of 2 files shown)


src/pages/voting/proposal/[id].tsx
```diff
@@ -5,7 +5,7 @@ import { getProposal } from 'state/voting/helpers'
 import { ProposalState } from 'state/types'
 import ProposalPageRouter from 'views/Voting/Proposal'
 
-const ProposalPage = ({ fallback }: InferGetStaticPropsType<typeof getStaticProps>) => {
+const ProposalPage = ({ fallback = {} }: InferGetStaticPropsType<typeof getStaticProps>) => {
```
Comment (chef-jojo) [inline src/pages/voting/proposal/[id].tsx:8]: Not sure why fallback can be `undefined` here.

Chef-Cheems (approved):

PR #3360, opened 2022-02-28
feat: Monorepo turbo part 1

chef-jojo:
### What changes
Only add uikit and eslint into packages. 
Use tsconfig path to map uikit, consume it by src.
FE code as workspace root now. Can be changed follow by https://github.com/pancakeswap/pancake-frontend/pull/3359 later

### Note
Turbo doesn't do anything at this moment.
Keep Lerna only use for packages version and changelog


packages/.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
+  "semi": true,
+  "printWidth": 120
```
chef-jojo, line 3: This file prevent different prettier format on packages. We can remove this later

tsconfig.json
```diff
@@ -19,7 +19,10 @@
     "noImplicitAny": false,
     "noFallthroughCasesInSwitch": true,
     "downlevelIteration": true,
-    "incremental": true
+    "incremental": true,
+    "paths": {
+      "@pancakeswap/uikit": ["../packages/uikit/src/index.ts"]
+    }
```
chef-jojo, line 25: If you are looking for something deeper. This is where magic happens. 
mapping src instead of as node module `dist` and avoid parallel `dev` scripts

packages/uikit/package.json
```diff
@@ -0,0 +1,56 @@
...
+  "files": [
+    "dist"
+  ],
+  "repository": "https://github.com/pancakeswap/pancake-toolkit/tree/master/packages/pancake-uikit",
+  "license": "MIT",
+  "scripts": {
+    "start": "yarn storybook",
+    "build": "rm -rf ./dist && rollup -c && tsc -d --emitDeclarationOnly --declarationDir dist",
+    "watch": "rollup -c -w",
+    "lint": "eslint 'src/**/*.{js,jsx,ts,tsx}'",
```
chefilip, line 17: those commands should be used in root ?

.gitignore
```diff
@@ -18,6 +18,8 @@ cypress/fixtures/example.json
 
 # production
 /build
+dist
+storybook-static
```
chefilip, line 23: add `.turbo` ignore

chef-jojo: good catch

chefilip:
Sync the latest uikit before merge

Chef-Cheems (approved):

PR #3367, opened 2022-03-01
fix: Intl object not found

chefilip:


src/utils/formatBalance.ts
```diff
@@ -64,7 +64,12 @@ export const formatFixedNumber = (number: FixedNumber, displayDecimals = 18, dec
 
 export const formatLocalisedCompactNumber = (number: number): string => {
   const codeFromStorage = getLanguageCodeFromLS()
-  return new Intl.NumberFormat(codeFromStorage, {
+
+  const isClient = typeof window === 'object'
+
+  if (!isClient || !window?.Intl) return ''
```
Chef-Cheems, line 70: @chefilip Is this for SSR? Is there some bug currently regarding this?
I wonder if we should install / write some polyfil instead of returning `''`

chefilip: Oh. It's on client-side. Some browsers don't support it.

chef-jojo: I think we have to handle Intl not supported browser. I don't wan to add polyfill here just for 10 users.
can we just return the `codeFromStorage` when isClient and no Intl support?

chefilip: Done

chef-jojo: sorry, should be `number` instead 😅 
Intl should works on node js.. umm
CleanShot 2022-03-01 at 19 20 41@2x

chefilip: Done again. :)

chef-jojo (approved):

PR #3368, opened 2022-03-01
fix: Check IntersectionObserver 

chefilip:
Screen Shot 2022-03-01 at 15 12 41

### DIFF
--- a/src/hooks/useIntersectionObserver.ts
+++ b/src/hooks/useIntersectionObserver.ts
@@ -1,4 +1,4 @@
-import { useEffect, useRef, useState } from 'react'
+import { useLayoutEffect, useRef, useState } from 'react'
 
 const useIntersectionObserver = () => {
   const [observerRefElement, setObserverRefElement] = useState(null)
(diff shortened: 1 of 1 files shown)


memoyil:
Does this happen on serverside rendering?

chefilip:
> Does this happen on serverside rendering?

The error is currently triggered on client side. But it would be a good practice to put `IntersectionObserver` in `useLayoutEffect` instead of `useEffect`

src/hooks/useIntersectionObserver.ts
```diff
@@ -1,18 +1,20 @@
...
   const observerRef = useRef((element) => setObserverRefElement(element))
   const intersectionObserverRef = useRef<IntersectionObserver>(null)
   const [isIntersecting, setIsIntersecting] = useState(false)
 
-  useEffect(() => {
-    if (!intersectionObserverRef.current && observerRefElement) {
+  useLayoutEffect(() => {
+    const isSupported = typeof window === 'object' && window.IntersectionObserver
+
+    if (isSupported && !intersectionObserverRef.current && observerRefElement) {
```
chef-jojo, line 12: if not isSupported we should just say it intersecting?

chefilip: Hmm. I think it would be fine to do that. Will add it

chef-jojo (approved):

PR #3392, opened 2022-03-02
refactor: Remove global constant objects

memoyil:
Using global constant objects can be dangerous if the variable is changed somewhere else


src/views/Nft/market/Collection/IndividualNFTPage/PancakeBunnyPage/index.tsx
```diff
@@ -121,7 +120,10 @@ const IndividualPancakeBunnyPageBase: React.FC<IndividualPancakeBunnyPageProps>
     }
   }, [cheapestBunny, bunnyId])
 
-  const sortedNfts = useMemo(() => allBunnies.sort(sortNFTsByPriceBuilder({ priceSort })), [allBunnies, priceSort])
```
memoyil, line 124: This sorts the same array which readonly type doesn't have that function

chefilip (dismissed):

PR #3395, opened 2022-03-03
perf: Prevent rerun toCallState when currentBlock changes

chef-jojo:


src/state/multicall/hooks.ts
```diff
@@ -225,11 +225,11 @@ export function useMultipleContractSingleData(
...
   const results = useCallsData(calls, options)
 
-  const currentBlock = useCurrentBlock()
+  const { cache } = useSWRConfig()
 
   return useMemo(() => {
-    return results.map((result) => toCallState(result, contractInterface, fragment, currentBlock))
-  }, [fragment, results, contractInterface, currentBlock])
+    return results.map((result) => toCallState(result, contractInterface, fragment, cache.get('blockNumber')))
+  }, [fragment, results, contractInterface, cache])
```
chefilip, line 232: Should we just depend on cache.get("blockNumber") instead of the whole cache

chef-jojo: It's quite interesting the `cache` is not causing re-render, here we *only get the blockNumber* when other dependents `fragment, results, contractInterface` change
if we depend on `cache.get("blockNumber")` it will be the same as original currentBlock hook, everytime blocknumber changes will cause the whole toCallState rerurn.

The blockNumber we are using here is only use the check the result is syncing
`const syncing = (blockNumber ?? 0) < latestBlockNumber`

chefilip: Interesting. 🤔  Minor optimized.

```
{
 const currentBlockNumber = cache.get('blockNumber')
 
 return results.map((result) => toCallState(result, contractInterface, fragment, currentBlockNumber))
}
```

chef-jojo: Yup better

chefilip (dismissed):

PR #3401, opened 2022-03-03
fix: Remove user data from farms and pools after logout

memoyil:
After logout redux still contains userdata from previous account

To reproduce:

1. Go to farms
2. Connect account
3. See harvest amount on farms
4. Logout
5. Harvest amount still visible

### DIFF
--- a/src/state/index.ts
+++ b/src/state/index.ts
@@ -1,4 +1,4 @@
-import { combineReducers, configureStore } from '@reduxjs/toolkit'
+import { combineReducers, configureStore, createAction } from '@reduxjs/toolkit'
 import { useMemo } from 'react'
 import { useDispatch } from 'react-redux'
 import {
@@ -22,13 +22,15 @@ import lotteryReducer from './lottery'
 import mint from './mint/reducer'
 import multicall from './multicall/reducer'
 import nftMarketReducer from './nftMarket/reducer'
-import poolsReducer from './pools'
+import poolsReducer, { initialPoolVaultState } from './pools'
 import predictionsReducer from './predictions'
 import swap from './swap/reducer'
 import transactions from './transactions/reducer'
 import user from './user/reducer'
 
 const PERSISTED_KEYS: string[] = ['user', 'transactions', 'lists']
+
+export const resetUserState = createAction<void>('global/resetUserState')
 
 const migrations = {
   0: (state) => {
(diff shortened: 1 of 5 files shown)


src/utils/clearUserStates.ts
```diff
@@ -1,10 +1,14 @@
...
+import { resetFarmsUserData } from 'state/farms'
+import { resetPoolsUserData } from 'state/pools'
+import { clearAllTransactions } from 'state/transactions/actions'
 import { connectorLocalStorageKey } from '@pancakeswap/uikit'
 import { connectorsByName } from './web3React'
-import { clearAllTransactions } from '../state/transactions/actions'
 
 export const clearUserStates = (dispatch: Dispatch<any>, chainId: number) => {
+  dispatch(resetFarmsUserData())
+  dispatch(resetPoolsUserData())
```
chefilip, line 11: I think instead of cleaning redux state separately, we could centralize the logic by capturing the logout action and cleaning everything we need. We could do it in root reducer or add a higher order reducer "cleanupReducer" to achieve this

memoyil: Done

src/state/index.ts
```diff
@@ -52,26 +54,59 @@ const persistConfig = {
...
+  mint,
+  burn,
+  multicall,
+  lists,
+})
+
+const rootReducer = (state, action) => {
+  if (action.type === 'global/resetUserState') {
+    // eslint-disable-next-line no-param-reassign
+    state = {
```
chefilip, line 78: I think after cleaning the state, we would not run further reducers. Returning state here is good enough.

memoyil: Done

memoyil:
@chefilip Build fixed, you can review

src/state/index.ts
chefilip: It seems the Build Fix Commit removed the return keyword ? 😀

memoyil: Fixed :D

src/state/index.ts
```diff
@@ -52,26 +57,73 @@ const persistConfig = {
...
+    // eslint-disable-next-line no-param-reassign
+    return {
+      ...state,
+      farms: {
+        ...state.farms,
+        data: state.farms.data.map((farm) => {
+          return {
+            ...farm,
+            userData: {
+              allowance: '0',
```
chefilip, line 89: Hmm. When I looked at this code, I realized my idea is bad. Exposing many detail of sub-reducers in the root reducer is quite hard to maintain in the future. For example, if we add more state in the store, we have to maintain this code as well. 

I will think another cleaner solution. 😅

memoyil: We can revert back to the run subreducers after action logic

memoyil: ah with that we still have to assign type :(

memoyil: I think if we let reducers run by assigning state only instead of return, we don't have to assign a type

chefilip: I think we could listen the reset action in each relevant sub-reducer to make it more precise. 

I think this way would fit better in redux paradigm. 

What do you think ? @memoyil

memoyil: Done

src/state/pools/index.ts
```diff
@@ -343,6 +344,15 @@ export const PoolsSlice = createSlice({
     },
   },
   extraReducers: (builder) => {
+    builder.addCase(resetUserState, (state) => {
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      state.data = state.data.map(({ userData, ...pool }) => {
+        return { ...pool }
```
chefilip, line 350: Great work ! @memoyil  

One minor thing: Do you think it's dangerous to leave `userData` as undefined ? 

If other places use `pool.userData.xyz` uncarefully, it might throw an error. Right ?

memoyil: True but also the initial state value is undefined, therefore currently all the places that we use, we check its availability beforehand so we are safe

chefilip: Yeah. I check its usages. We are safe for now. Thanks !!! Great work !!!

chefilip (approved):

PR #3418, opened 2022-03-07
fix(wallet): Local injected eager connect

chef-jojo:
Not gonna dive in why web3-react has stale connection on local. just setTimeout to throw login to event loop
https://github.com/NoahZinsmeister/web3-react/blob/v6/packages/core/src/manager.ts#L141-L143


chefilip (approved):
lgtm

PR #3420, opened 2022-03-07
chore: Add and initialize style-lint

Chef-Yogi:
run 
```yarn lint:style```
and will get some hints and suggestions like this

demo

we can discuss the rules together
and it's not a force check right now

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -14,6 +14,7 @@
     "test:config": "jest src/config/__tests__/ --testPathIgnorePatterns",
     "integration-test": "start-server-and-test start http://localhost:3000 'cypress run'",
     "lint": "eslint 'src/**/*.{js,jsx,ts,tsx}'",
+    "lint:style": "npx stylelint '**/*.{js,jsx,ts,tsx}'",
```
chefilip, line 17: II think we only need to lint `tsx` ?

Chef-Yogi: Yeah, we don't need js and jsx
but, I think we need ts either because styled-components can be in ts file also( export and const only )

chefilip: Yeah. You're right.

chefilip (dismissed):

PR #3421, opened 2022-03-07
feat: (farms): Show token amounts inside LP

memoyil:
Screenshot 2022-03-07 at 10 07 08
Screenshot 2022-03-07 at 11 07 29


Chef-Cheems (approved):
Didn't even realize the typings were that bad in Farms, thanks for fixing those 👍

PR #3422, opened 2022-03-07
perf: Use selectors for farms

memoyil:
Thanks @chefilip that gave me the idea on pools, now I applied that to farms

### DIFF
--- a/src/state/farms/selectors.ts
+++ b/src/state/farms/selectors.ts
@@ -38,21 +38,20 @@ const deserializeFarm = (farm: SerializedFarm): DeserializedFarm => {
 }
 
 const selectCakeFarm = (state: State) => state.farms.data.find((f) => f.pid === 251)
-const selectFarmFromPid = (pid: number) => (state: State) => state.farms.data.find((f) => f.pid === pid)
-const selectFarmFromLpSymbol = (lpSymbol: string) => (state: State) =>
-  state.farms.data.find((f) => f.lpSymbol === lpSymbol)
+const selectFarmByKey = (key: string, value: string | number) => (state: State) =>
+  state.farms.data.find((f) => f[key] === value)
 
 export const makeFarmFromPidSelector = (pid: number) =>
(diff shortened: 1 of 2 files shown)


src/state/farms/selectors.ts
```diff
@@ -0,0 +1,106 @@
...
+    tokenAmountTotal: farm.tokenAmountTotal ? new BigNumber(farm.tokenAmountTotal) : BIG_ZERO,
+    lpTotalInQuoteToken: farm.lpTotalInQuoteToken ? new BigNumber(farm.lpTotalInQuoteToken) : BIG_ZERO,
+    lpTotalSupply: farm.lpTotalSupply ? new BigNumber(farm.lpTotalSupply) : BIG_ZERO,
+    tokenPriceVsQuote: farm.tokenPriceVsQuote ? new BigNumber(farm.tokenPriceVsQuote) : BIG_ZERO,
+    poolWeight: farm.poolWeight ? new BigNumber(farm.poolWeight) : BIG_ZERO,
+  }
+}
+
+const selectCakeFarm = (state: State) => state.farms.data.find((f) => f.pid === 251)
+const selectFarmFromPid = (pid: number) => (state: State) => state.farms.data.find((f) => f.pid === pid)
```
chefilip, line 41: I think we could consolidate `selectFarmFromPid` and `selectFarmFromLpSymbol` into one function.
```
const selectFarmByKey = (key: string, value: string|number) => data.find(f => f[key] === value)
```

chefilip: Also, should we add `deserializeFarm` in this function as well ?

memoyil: > `const selectFarmByKey = (key: string, value: string|number) => data.find(f => f[key] === value)`

Done

memoyil: > Also, should we add `deserializeFarm` in this function as well ?

We cannot put deserializeFarm inside the selector, it will break memoization because it creates a new object

chefilip (approved):

PR #3427, opened 2022-03-07
perf: Bottom Drawer animation optimization

Chef-Yogi:


chefilip:
Good work ! 👍

chef-jojo:
Maybe not that efficient because local doesn't has api call. Great work though 👍🏽

chef-jojo (approved):

PR #3433, opened 2022-03-08
fix: Collection listing order by token id load more

memoyil:
To reproduce: 

1. Go to any collection items
2. Select filter by all
3. Select order by token id
4. Go to bottom and load more
5. At first click nothing happens, user has to click twice to get more items

### DIFF
--- a/src/views/Nft/market/hooks/useCollectionNfts.tsx
+++ b/src/views/Nft/market/hooks/useCollectionNfts.tsx
@@ -144,7 +144,7 @@ const fetchAllNfts = async (
   } else if (tokenIdsFromFilter) {
     tokenIds = tokenIdsFromFilter.slice(page * REQUEST_SIZE, (page + 1) * REQUEST_SIZE)
   } else {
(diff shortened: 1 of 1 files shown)


src/views/Nft/market/hooks/useCollectionNfts.tsx
```diff
@@ -144,7 +144,7 @@ const fetchAllNfts = async (
   } else if (tokenIdsFromFilter) {
     tokenIds = tokenIdsFromFilter.slice(page * REQUEST_SIZE, (page + 1) * REQUEST_SIZE)
   } else {
-    collectionNftsResponse = await getNftsFromCollectionApi(collection.address, REQUEST_SIZE, page)
+    collectionNftsResponse = await getNftsFromCollectionApi(collection.address, REQUEST_SIZE, page + 1)
```
memoyil, line 147: Nft api root page is 1 instead of 0

chefilip (approved):

PR #3435, opened 2022-03-08
fix: (limit-orders): Order still visible after executed in open orders tab

memoyil:
Currently when open orders from gelato returns empty array, orders in local storage don't get updated

### DIFF
--- a/src/views/LimitOrders/hooks/useGelatoLimitOrdersHistory.ts
+++ b/src/views/LimitOrders/hooks/useGelatoLimitOrdersHistory.ts
@@ -4,7 +4,7 @@ import useSWR from 'swr'
 import { SLOW_INTERVAL } from 'config/constants'
 import { useMemo } from 'react'
 
-import { getLSOrders, saveOrder, removeOrder } from 'utils/localStorageOrders'
+import { getLSOrders, saveOrder, removeOrder, saveOrders } from 'utils/localStorageOrders'
 import useGelatoLimitOrdersLib from 'hooks/limitOrders/useGelatoLimitOrdersLib'
 
 import { ORDER_CATEGORY } from '../types'
@@ -13,15 +13,13 @@ function newOrdersFirst(a: Order, b: Order) {
   return Number(b.updatedAt) - Number(a.updatedAt)
 }
 
-function syncOrderToLocalStorage({ chainId, account, orders }) {
+function syncOrderToLocalStorage({ chainId, account, orders }: { chainId: number; account: string; orders: Order[] }) {
   const ordersLS = getLSOrders(chainId, account)
 
-  orders.forEach((order: Order) => {
-    const hasLSOrder = ordersLS.some((confOrder) => confOrder.id.toLowerCase() === order.id.toLowerCase())
-    if (!hasLSOrder) {
-      saveOrder(chainId, account, order)
-    }
-  })
+  const newOrders = orders.filter(
+    (order: Order) => !ordersLS.some((confOrder) => confOrder.id.toLowerCase() === order.id.toLowerCase()),
+  )
+  saveOrders(chainId, account, newOrders)
 
   ordersLS.forEach((confOrder: Order) => {
     const updatedOrder = orders.find((order) => confOrder.id.toLowerCase() === order.id.toLowerCase())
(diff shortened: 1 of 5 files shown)


src/views/LimitOrders/hooks/useGelatoLimitOrdersHistory.ts
```diff
@@ -17,12 +17,21 @@ function syncOrderToLocalStorage({ chainId, account, orders }) {
   const ordersLS = getLSOrders(chainId, account)
 
   orders.forEach((order: Order) => {
```
chefilip, line 19: It seems we could combiine 2 logics by simply updating the latest orders if latest orders and cache orders are different.
 
 ```
 const ordersLS = getLSOrders(chainId, account)
 
 if (stringify(ordersLS) !== stringify(orders) {
   saveOrders(chainId, account, orders)
 }
 ```

memoyil: I am a bit reluctant to do this because stringify can become a heavy operation if account has too many orders

chefilip: Yeah. Agreed. It's just a simple way to do it. I will think about another optimal diff algorithm and override the current orders with the latest order.

Currently, we loop and store order one by one in local storage.

memoyil: I added saveOrders and removeOrders methods

src/views/LimitOrders/hooks/useGelatoLimitOrdersHistory.ts
```diff
@@ -13,14 +13,21 @@ function newOrdersFirst(a: Order, b: Order) {
...
   const ordersLS = getLSOrders(chainId, account)
 
-  orders.forEach((order: Order) => {
-    const orderExists = ordersLS.find((confOrder) => confOrder.id.toLowerCase() === order.id.toLowerCase())
+  const newOrders = orders.filter(
+    (order: Order) => !ordersLS.some((confOrder) => confOrder.id.toLowerCase() === order.id.toLowerCase()),
+  )
+  saveOrders(chainId, account, newOrders)
+
+  ordersLS.forEach((confOrder: Order) => {
```
memoyil, line 24: We can't do batch update here since there updateAt check as well as removing the order when it is status changed

chefilip: Hmm. Got it. Thanks

src/views/LimitOrders/hooks/useGelatoLimitOrdersHistory.ts
```diff
@@ -13,14 +13,20 @@ function newOrdersFirst(a: Order, b: Order) {
...
 
-  orders.forEach((order: Order) => {
-    const orderExists = ordersLS.find((confOrder) => confOrder.id.toLowerCase() === order.id.toLowerCase())
+  const ordersLSHashSet = hashOrderSet(ordersLS)
+  const newOrders = orders.filter((order: Order) => !ordersLSHashSet.has(hashOrder(order)))
+  saveOrders(chainId, account, newOrders)
 
-    if (!orderExists || (orderExists && Number(orderExists.updatedAt) < Number(order.updatedAt))) {
-      saveOrder(chainId, account, order)
+  ordersLS.forEach((confOrder: Order) => {
```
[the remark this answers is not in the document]
memoyil: No, we don't need to because orders that we add above is not in the local storage therefore we don't need to compare them with already saved ones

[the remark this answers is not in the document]
chefilip: Got it. Thanks

chefilip:
Hi @memoyil  !

It seems the order cache behaviour is changed. When user switches between tabs, the cache order lists update accordingly. 

The order cache list should keep both orders of both tabs, right ?

memoyil:
> Hi @memoyil !
> 
> It seems the order cache behaviour is changed. When user switches between tabs, the cache order lists update accordingly.
> 
> The order cache list should keep both orders of both tabs, right ?

Fixed

PR #3437, opened 2022-03-09
feat: Provide more token lists options

ChefMomota:

PR #3442, opened 2022-03-09
perf: Token list performance for huge data

chef-jojo:


memoyil:
This will affect remove liquidity pairs not be able to found if it is not included in default list (currently it contains small set of pairs). Users will not be able to see their liquidities in remove liquidity component

chef-jojo:
> This will affect remove liquidity pairs not be able to found if it is not included in default list (currently it contains small set of tokens). Users will not be able to see their lp's in remove liquidity component. This computation imho it is necessary and only be done in remove liquidity afaik.

alright, picked changes that is confirmed #3443

About the `includeDefaultLists` removal, it still includes user added tokens and enabled list. I don't see why we keep default list if we are adding more lists to it. See #3437

memoyil:
> > This will affect remove liquidity pairs not be able to found if it is not included in default list (currently it contains small set of tokens). Users will not be able to see their lp's in remove liquidity component. This computation imho it is necessary and only be done in remove liquidity afaik.
> 
> alright, picked changes that is confirmed #3443
> 
> About the `includeDefaultLists` removal, it still includes user added tokens and enabled list. I don't see why we keep default list if we are adding more lists to it. See #3437

If user hasn't yet added corresponding tokens or enabled list that contains those tokens on the device (can happen after cleared cache or new device or mobile wallet apps usually clear cache after exit), they will not see their lp's on remove liquidity. In order to see it it has to add that missing token or enable the list or import lp (except the last option those are quite ambiguous to novice users) and many users doesn't know to use import lp properly thus it created a problem in the past that lots user complained lp's are not visible after removed from farm. I think at least we should include pancakeswap extended lists for checking user lp's.

chef-jojo:
I can see benefits including extended list. Fetch all tokens by enabled list to try get user lp feels kind of dumb to me
Maybe we can use subgraph instead. 
```graphql
 mints(orderBy: timestamp, orderDirection: desc, where: { to: $to })
```

src/state/lists/hooks.ts
```diff
@@ -136,14 +136,11 @@ function useCombinedTokenMapFromUrls(urls: string[] | undefined): TokenAddressMa
...
-export function useActiveListUrls(includeDefaultLists = false): string[] | undefined {
-  const defaultList = includeDefaultLists
-    ? DEFAULT_LIST_OF_LISTS.filter((url) => !UNSUPPORTED_LIST_URLS.includes(url))
-    : []
+export function useActiveListUrls(): string[] | undefined {
   const stateList = useSelector<AppState, AppState['lists']['activeListUrls']>(
     (state) => state.lists.activeListUrls,
   )?.filter((url) => !UNSUPPORTED_LIST_URLS.includes(url))
-  return [...new Set(defaultList.concat(stateList ?? []))]
+  return useMemo(() => stateList, [stateList])
```
chefilip, line 143: Could we add the logic inside`useSelector` to utilize selector memoization.

```
const activeListUrlsSelector = (state) => {
  return state.lists.activeListUrls?.filter((url => !UNSUPPORTED_LIST_URLS.includes(url)))
}

function useCombinedActiveList() {
 return useSelector(activeListUrlsSelector)
}

const stateList = useSelector(
```

src/state/lists/hooks.ts
```diff
@@ -136,14 +136,11 @@ function useCombinedTokenMapFromUrls(urls: string[] | undefined): TokenAddressMa
 }
 
 // filter out unsupported lists
-export function useActiveListUrls(includeDefaultLists = false): string[] | undefined {
-  const defaultList = includeDefaultLists
-    ? DEFAULT_LIST_OF_LISTS.filter((url) => !UNSUPPORTED_LIST_URLS.includes(url))
-    : []
+export function useActiveListUrls(): string[] | undefined {
```
chefilip, line 139: How about creating a list selector file `list/selectors.ts` ?

```
  const activeListUrlsSelector = (state) => state.lists.activeListUrls
  
  const combinedActiveListUrlsSelector = createSelector(activeListUrlsSelector, (urls) => urls.filter(...))
```

In `hooks.ts`: 
```

useActiveListUrls() { return useSelector(combinedActiveListUrlsSelector))
```

memoyil:
> I can see benefits including extended list. Fetch all tokens by enabled list to try get user lp feels kind of dumb to me Maybe we can use subgraph instead.
> 
> ```graphql
>  mints(orderBy: timestamp, orderDirection: desc, where: { to: $to })
> ```

I agree with you, I am checking it

memoyil:
@chef-jojo I checked the subgraph data for mints, it is pretty old (30 days). We can't use it :(

Chef-Cheems (approved):

src/state/user/hooks/index.tsx
```diff
@@ -417,10 +417,7 @@ export function toV2LiquidityToken([tokenA, tokenB]: [Token, Token]): Token {
  */
 export function useTrackedTokenPairs(): [Token, Token][] {
   const { chainId } = useActiveWeb3React()
-  const tokens = useAllTokens(true)
-  const filterTokens = Object.keys(tokens)
-    .slice(0, 1000)
-    .map((key) => tokens[key])
+  const tokens = useOfficialsAndUserAddedTokens()
```
memoyil, line 420: 👍

memoyil:
@chef-jojo I have one quick question other than that LGTM

memoyil:
User added tokens should be on top of the list after balanced tokens, but current behavior is ordered alphabetically.

src/state/wallet/hooks.ts
```diff
@@ -57,7 +57,12 @@ export function useTokenBalancesWithLoadingIndicator(
 
   const validatedTokenAddresses = useMemo(() => validatedTokens.map((vt) => vt.address), [validatedTokens])
 
-  const balances = useMultipleContractSingleData(validatedTokenAddresses, ERC20_INTERFACE, 'balanceOf', [address])
+  const balances = useMultipleContractSingleData(
+    validatedTokenAddresses,
+    ERC20_INTERFACE,
+    'balanceOf',
+    useMemo(() => [address], [address]),
```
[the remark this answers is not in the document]
chef-jojo: it's array in deps, `[[address]]`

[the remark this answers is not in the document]
memoyil: 👍

memoyil:
LGTM

PR #3444, opened 2022-03-09
fix: Multiple banner still shows fade effect when only one item

memoyil:
https://user-images.githubusercontent.com/2213635/157502567-480cdae1-e130-4a1a-97a1-b60b269bcb18.mov


Chef-Yogi:
Thanks for fixing this 🙏!

chef-jojo (approved):

PR #3451, opened 2022-03-10
feat: Dynamic status tags for menu items

memoyil:
Feature for Menu item tags (currently for ifo only) to be placed automatically based on component status

### DIFF
--- a/src/components/Menu/index.tsx
+++ b/src/components/Menu/index.tsx
@@ -6,10 +6,9 @@ import { languageList } from 'config/localization/languages'
 import { useTranslation } from 'contexts/Localization'
 import PhishingWarningBanner from 'components/PhishingWarningBanner'
 import useTheme from 'hooks/useTheme'
-import { useMenuItemsStatus } from 'hooks/useMenuItemsStatus'
+import { useMenuItems } from 'hooks/useMenuItems'
 import { usePriceCakeBusd } from 'state/farms/hooks'
 import { usePhishingBannerManager } from 'state/user/hooks'
-import config from './config/config'
 import UserMenu from './UserMenu'
 import GlobalSettings from './GlobalSettings'
 import { getActiveMenuItem, getActiveSubMenuItem } from './utils'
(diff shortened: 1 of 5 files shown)


src/views/Home/components/Banners/hooks/useIsRenderIFOBanner.tsx
```diff
@@ -1,17 +1,12 @@
...
 import { useActiveIfoWithBlocks } from 'hooks/useActiveIfoWithBlocks'
-import { getStatus } from 'views/Ifos/hooks/helpers'
 
 const useIsRenderIfoBanner = () => {
   const currentBlock = useCurrentBlock()
 
   const activeIfoWithBlocks = useActiveIfoWithBlocks()
 
-  const isIfoAlive = !!(currentBlock && activeIfoWithBlocks && activeIfoWithBlocks.endBlock > currentBlock)
-  const status = isIfoAlive
```
memoyil, line 11: isIfoAlive is enough to determine rendering

src/components/Menu/index.tsx
```diff
@@ -42,7 +45,7 @@ const Menu = (props) => {
       langs={languageList}
       setLang={setLanguage}
       cakePriceUsd={cakePriceUsd.toNumber()}
-      links={config(t)}
+      links={config(t, menuItemsStatus)}
```
chefilip, line 48: I think we could turn config into a hook to avoid re-rendering

chefilip (commented):
Fantastic. Thanks for doing this ❤️

src/hooks/useMenuItemsStatus.ts
```diff
@@ -0,0 +1,19 @@
...
+  const ifoStatus =
+    currentBlock && activeIfo && activeIfo.endBlock > currentBlock
+      ? getStatus(currentBlock, activeIfo.startBlock, activeIfo.endBlock)
+      : null
+
+  return ifoStatus
+    ? {
+        '/ifo': ifoStatus === 'coming_soon' ? 'soon' : ifoStatus,
+      }
+    : null
```
chef-jojo, line 18: let's use memo

src/components/Menu/index.tsx
```diff
@@ -24,6 +25,8 @@ const Menu = (props) => {
   const activeMenuItem = getActiveMenuItem({ menuConfig: config(t), pathname })
```
chef-jojo, line 25: here can lift it to useMenuConfig or something

chef-jojo:
@memoyil, thanks for the PR.
I have an idea to inject hook inside the menu items. Having a currentBlock hook on top of menu looks a bit scary, but I think it's fine for now.

src/hooks/useMenuItemsStatus.ts
memoyil: Done

src/components/Menu/index.tsx
memoyil: Done

src/components/Menu/index.tsx
memoyil: Done

memoyil:
> @memoyil, thanks for the PR. I have an idea to inject hook inside the menu items. Having a currentBlock hook on top of menu looks a bit scary, but I think it's fine for now.

I think so too, we have cakePriceUsd in that menu that is updated with fast refresh same as block number and there is no/negligible performance impact

PR #3465, opened 2022-03-11
fix: (limit-orders): Open orders still visible after executed

memoyil:


src/views/LimitOrders/hooks/useGelatoLimitOrdersHistory.ts
```diff
@@ -52,13 +55,16 @@ const useOpenOrders = (turnOn: boolean): Order[] => {
...
       try {
-        const orders = await gelatoLimitOrders.getOpenOrders(account.toLowerCase(), false)
+        const [openOrders, exeOrders] = await Promise.all([
+          gelatoLimitOrders.getCancelledOrders(account.toLowerCase(), false),
+          gelatoLimitOrders.getExecutedOrders(account.toLowerCase(), false),
+        ])
 
         syncOrderToLocalStorage({
-          orders,
+          orders: [...openOrders, ...exeOrders],
```
memoyil, line 64: we should both fetch open and executed orders to determine order status switched to executed

src/views/LimitOrders/hooks/useGelatoLimitOrdersHistory.ts
```diff
@@ -4,23 +4,42 @@ import useSWR from 'swr'
...
+  account,
+  orders,
+  syncTypes,
+}: {
+  chainId: number
+  account: string
+  orders: Order[]
+  syncTypes: LimitOrderType[]
+}) {
+  const allOrdersLS = getLSOrders(chainId, account)
```
memoyil, line 27: we should fetch all orders to check to filter out locally updated order

src/views/LimitOrders/hooks/useGelatoLimitOrdersHistory.ts
```diff
@@ -4,23 +4,42 @@ import useSWR from 'swr'
...
+
+  const typeOrders = orders.filter((order) => syncTypes.some((type) => type === order.status))
+  const otherOrders = orders.filter((order) => syncTypes.some((type) => type !== order.status))
+  const typeOrdersLS = allOrdersLS.filter((order) => syncTypes.some((type) => type === order.status))
+  typeOrdersLS.forEach((confOrder: Order) => {
+    const updatedOrder = typeOrders.find((order) => confOrder.id.toLowerCase() === order.id.toLowerCase())
+
+    if (updatedOrder && Number(confOrder.updatedAt) < Number(updatedOrder.updatedAt)) {
+      saveOrder(chainId, account, updatedOrder)
+    } else if (!updatedOrder && otherOrders.some((order) => confOrder.id.toLowerCase() === order.id.toLowerCase())) {
```
memoyil, line 41: If the order is not found desired status but updated in graph, remove it from that type of orders

src/views/LimitOrders/hooks/useGelatoLimitOrdersHistory.ts
```diff
@@ -52,13 +55,17 @@ const useOpenOrders = (turnOn: boolean): Order[] => {
     startFetch ? ['gelato', 'openOrders'] : null,
     async () => {
       try {
-        const orders = await gelatoLimitOrders.getOpenOrders(account.toLowerCase(), false)
+        const [openOrders, canOrders, exeOrders] = await Promise.all([
```
memoyil, line 58: We should fetch all the orders to determine order status updated to cancelled or executed. (Cancelled added as wel to possibility to cancel order from another device)

src/views/LimitOrders/hooks/useGelatoLimitOrdersHistory.ts
```diff
@@ -36,29 +55,30 @@ const useOpenOrders = (turnOn: boolean): Order[] => {
     startFetch ? ['gelato', 'openOrders'] : null,
     async () => {
       try {
-        const orders = await gelatoLimitOrders.getOpenOrders(account.toLowerCase(), false)
+        const orders = await gelatoLimitOrders.getOrders(account.toLowerCase(), false)
```
Chef-Cheems, line 58: Any reason to use `getOrders` instead of `getOpenOrders` here?

memoyil: This is for to detect open order is executed, otherwise it will stay on open orders on the same sevice if user doesnt change the tab to order history. But i have a optimised solution which we dont have to fetch all orders, i will apply soon

src/views/LimitOrders/hooks/useGelatoLimitOrdersHistory.ts
```diff
@@ -39,26 +70,28 @@ const useOpenOrders = (turnOn: boolean): Order[] => {
         const orders = await gelatoLimitOrders.getOpenOrders(account.toLowerCase(), false)
 
         syncOrderToLocalStorage({
```
chefilip, line 70: should be `await` ?

memoyil: Done

src/views/LimitOrders/hooks/useGelatoLimitOrdersHistory.ts
```diff
@@ -1,28 +1,57 @@
...
+}) {
+  const allOrdersLS = getLSOrders(chainId, account)
+
+  const lsOrdersHashSet = hashOrderSet(allOrdersLS)
+  const newOrders = orders.filter((order: Order) => !lsOrdersHashSet.has(hashOrder(order)))
+  saveOrders(chainId, account, newOrders)
+
+  const typeOrdersLS = allOrdersLS.filter((order) => syncTypes.some((type) => type === order.status))
+
+  for (let i = 0; i < typeOrdersLS.length; i++) {
```
chefilip, line 41: for History Order Tab, we don't have to take this update step, right ?

memoyil: Yes, fixed

chefilip (approved):

PR #3466, opened 2022-03-11
fix: Remove limit order local storage data after disconnect

memoyil:
To reproduce:

1. Connect account that has limit orders
2. Go limit order page
3. Go to order history
4. Disconnect account 
5. See data still available in web storage

### DIFF
--- a/src/utils/getLocalStorageItemKeys.ts
+++ b/src/utils/getLocalStorageItemKeys.ts
@@ -1,0 +1,11 @@
+const getLocalStorageItemKeys = (prefix: string) => {
+  const result = []
+  for (let i = 0; i < localStorage.length; i++) {
+    if (localStorage.key(i).startsWith(prefix)) {
+      result.push(localStorage.key(i))
+    }
+  }
+  return result
+}
+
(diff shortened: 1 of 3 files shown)


src/utils/getLocalStorageItemKeys.ts
```diff
@@ -0,0 +1,11 @@
+const getLocalStorageItemKeys = (prefix: string) => {
+  const result = []
+  for (let i = 0; i < localStorage.length; i++) {
+    if (localStorage.key(i).startsWith(prefix)) {
```
chefilip, line 4: Could we check if the key exists before calling starWith ?

chefilip (approved):

src/utils/getLocalStorageItemKeys.ts
memoyil: Afaik calling Key function with index always return value. Should we add an additional check?

chefilip: Hmm. If that is the case, I think it's safe to keep it this way

chefilip (approved):

PR #3490, opened 2022-03-14
fix: Token list sorting ref

chef-jojo:
sorting return same reference


chefilip (approved):

chefilip:
> This sort method is kind of error prone, magbe we should switch all sort calls to lodash sortBy wdyt?

Totally agree. Immutability is better

PR #3495, opened 2022-03-15
refactor: Use orderBy instead of sort

memoyil:
Orderby should be used where sort is applied on direct array references to not sort the array that is passed

### DIFF
--- a/src/utils/localStorageOrders.ts
+++ b/src/utils/localStorageOrders.ts
@@ -1,5 +1,6 @@
 import { Order } from '@gelatonetwork/limit-orders-lib'
 import { get, set, clear } from 'local-storage'
+import orderBy from 'lodash/orderBy'
 
 export const LS_ORDERS = 'gorders_'
 
@@ -130,9 +131,6 @@ export function confirmOrderSubmission(chainId: number, account: string, submiss
 
 export const getUniqueOrders = (allOrders: Order[]): Order[] => [
   ...new Map(
-    allOrders
-      // sort by `updatedAt` asc so that the most recent one will be used
-      .sort((a, b) => parseFloat(a.updatedAt) - parseFloat(b.updatedAt))
-      .map((order) => [order.id, order]),
+    orderBy(allOrders, (order) => parseFloat(order.updatedAt), 'desc').map((order) => [order.id, order]),
   ).values(),
 ]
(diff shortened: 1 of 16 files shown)


src/utils/localStorageOrders.ts
```diff
@@ -130,9 +131,6 @@ export function confirmOrderSubmission(chainId: number, account: string, submiss
 
 export const getUniqueOrders = (allOrders: Order[]): Order[] => [
   ...new Map(
-    allOrders
-      // sort by `updatedAt` asc so that the most recent one will be used
-      .sort((a, b) => parseFloat(a.updatedAt) - parseFloat(b.updatedAt))
-      .map((order) => [order.id, order]),
+    orderBy(allOrders, (order) => parseFloat(order.updatedAt), 'desc').map((order) => [order.id, order]),
```
memoyil, line 134: @Chef-Cheems To get the most recent one shouldn't we order by descending?

Chef-Cheems: yes, descending. idk why it was asc and why nobody noticed 🤔 
thanks!

src/views/FarmAuction/helpers.tsx
```diff
@@ -14,25 +15,15 @@ export const FORM_ADDRESS =
...
-    .map((bidder, index) => {
-      const bidderInfo = getBidderInfo(bidder.account)
-      return {
-        ...bidderInfo,
-        position: index + 1,
-        account: bidder.account,
-        amount: bidder.amount,
-      }
-    })
+  const sortedBidders = orderBy(bidders, (bidder) => bidder.amount, 'desc').map((bidder, index) => {
```
Chef-Cheems, line 18: @memoyil since `amount` is a BigNumber I think we should cast it `.toNumber()`

src/views/Nft/market/ActivityHistory/utils/sortActivity.tsx
```diff
@@ -50,11 +51,7 @@ export const sortActivity = ({
 
   const allActivity = [...transformAskOrders(askOrders), ...transformTransactions(transactions)]
   if (allActivity.length > 0) {
-    const sortedByMostRecent = allActivity.sort((activityItem1, activityItem2) => {
-      const timestamp1 = BigNumber.from(activityItem1.timestamp)
-      const timestamp2 = BigNumber.from(activityItem2.timestamp)
-      return timestamp2.sub(timestamp1).toNumber()
-    })
+    const sortedByMostRecent = orderBy(allActivity, (activity) => BigNumber.from(activity.timestamp), 'desc')
```
Chef-Cheems, line 54: I think orderBy can't handle BigNumber like that. There was a bug once in farms table where orderBy was passed BN in second argument and it looked like it worked but it was just doing random stuff. I believe it can only work with primitives (it could only work with custom classes if operator overloading was a thing in JS)
by the way why we even need BN for timestamp?

memoyil: Done, converted to number

src/views/FarmAuction/helpers.tsx
memoyil: Done

PR #3496, opened 2022-03-15
fix: Keep prediction buffer seconds to default 60 instead 30 from onchain

memoyil:
Buffer seconds in prediction contract is 30 which sometimes not enough to detect expired round is actually canceled or not therefore it shows round cancelled incorrectly for amount of time.


src/views/Predictions/hooks/usePollPredictions.ts
```diff
@@ -26,15 +26,12 @@ const usePollPredictions = () => {
       batch(() => {
         dispatch(fetchRounds(liveCurrentAndRecent))
         dispatch(fetchMarketData())
-      })
-
-      if (account) {
-        const epochRange = range(earliestEpoch, currentEpoch + 1)
-        batch(() => {
+        if (account) {
```
memoyil, line 29: This is a refactor not related to fix

chef-jojo (approved):

PR #3506, opened 2022-03-16
fix: (prediction) Wrong buffer phase let users to bet until the very end of the open round

memoyil:
Currently buffer phase is not working correctly which let users to bet until the very end of open round which results transaction failures on onchain side when submitting a transaction exceeds lock timestamp

@Chef-Cheems can you confirm ?

### DIFF
--- a/src/state/predictions/config.ts
+++ b/src/state/predictions/config.ts
@@ -1,7 +1,9 @@
+import { BSC_BLOCK_TIME } from 'config'
+
 export const REWARD_RATE = 0.97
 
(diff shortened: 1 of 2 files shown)


Chef-Cheems (approved):
Yep, extra second is good 👍 
Could just put >= in existing solution, but I like new way better, easier to read, thanks!

memoyil:
> Yep, extra second is good 👍
> Could just put >= in existing solution, but I like new way better, easier to read, thanks!

Actually previous approach was checking the buffer period by adding it with lock timestamp therefore buffer period never evaluated to true. We should create a buffer period before the lock timestamp like this right?

PR #3509, opened 2022-03-16
fix: (prediction): Swiper doesn't work after clicking back button on round card on mobile

memoyil:
To reproduce: 

1. Go to prediction on mobile
2. Connect account
3. Go to open round
4. Select up or down
5. Tap back button
6. See swiper doesn't work anymore


src/views/Predictions/components/RoundCard/SetPositionCard.tsx
```diff
@@ -253,15 +242,16 @@ const SetPositionCard: React.FC<SetPositionCardProps> = ({ position, togglePosit
           {account ? (
             <Button
               width="100%"
-              disabled={!account || disabled}
```
memoyil, line 256: we don't need to check account which is already checked outer block

PR #3517, opened 2022-03-17
chore: Try derfer next script to improve LCP & FCP

Chef-Yogi:
this is an experiment for improve FCP

before:
demo
after:
demo2

### DIFF
--- a/src/pages/_document.tsx
+++ b/src/pages/_document.tsx
@@ -56,7 +56,7 @@ class MyDocument extends Document {
             />
           </noscript>
           <Main />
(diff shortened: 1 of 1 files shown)


Chef-Yogi:
@chef-jojo @chefilip  
Not sure this improvement is a deviation or not,
but the screenshot is run lighthouse on desktop.
I think mobile doesn't improve at all, need more investigation 🤔

chef-jojo:
Prefer run lighthouse on cloud. Local lighthouse scores depends on your equipment

PR #3519, opened 2022-03-17
fix: Reset lottery user state when logout

memoyil:


src/state/lottery/index.ts
```diff
@@ -5,6 +5,8 @@ import { LotteryState, LotteryRoundGraphEntity, LotteryUserGraphEntity, LotteryR
 import { fetchLottery, fetchCurrentLotteryIdAndMaxBuy } from './helpers'
 import getLotteriesData from './getLotteriesData'
 import getUserLotteryData, { getGraphLotteryUser } from './getUserLotteryData'
+import { resetUserState } from '../global/actions'
+import { initialPoolVaultState } from '../pools'
```
chefilip, line 9: eslint

memoyil: Done

chefilip (approved):

PR #3533, opened 2022-03-18
fix: Avoid collection.address error

chefilip:
Screen Shot 2022-03-18 at 05 08 00


chef-jojo (approved):

memoyil:
This will push the error to useCollectionNfts we need to fix there as well

PR #3534, opened 2022-03-18
chore: Fix IntersectionObserver

chefilip:
Screen Shot 2022-03-18 at 04 59 15


memoyil:
Imho if intersection is not supported, we should handle the places where we rely on isIntersecting. For example on farms we should show all item on the list.

chefilip:
> Imho if intersection is not supported, we should handle the places where we rely on isIntersecting. For example on farms we should show all items of the list.

I already handle that level. This one is fixing the root cause of where component is used.

PR #3537, opened 2022-03-18
refactor: Move transaction cleanup on logout to reset user state action

memoyil:


src/state/transactions/reducer.ts
```diff
@@ -77,5 +79,12 @@ export default createReducer(initialState, (builder) =>
       } else if (transactions[chainId]?.[hash].type === 'limit-order-cancellation') {
         confirmOrderCancellation(chainId, receipt.from, hash, receipt.status !== 0)
       }
+    })
+    .addCase(resetUserState, (transactions, { payload: { chainId } }) => {
+      if (transactions[chainId]) {
+        transactions[chainId] = {}
+      }
+      const lsOrderKeys = getLocalStorageItemKeys(LS_ORDERS)
+      lsOrderKeys.forEach((lsOrderKey) => window.localStorage.removeItem(lsOrderKey))
```
chef-jojo, line 88: Why do we need to remove limit order localstorage? Isn't it store by account?

memoyil: Yes it is stored per account but we need to remove them to prevent other users to see the previous user’s actions on the same device by checking local storage

src/state/transactions/reducer.ts
```diff
@@ -77,5 +79,12 @@ export default createReducer(initialState, (builder) =>
       } else if (transactions[chainId]?.[hash].type === 'limit-order-cancellation') {
         confirmOrderCancellation(chainId, receipt.from, hash, receipt.status !== 0)
       }
+    })
+    .addCase(resetUserState, (transactions, { payload: { chainId } }) => {
+      if (transactions[chainId]) {
+        transactions[chainId] = {}
+      }
+      const lsOrderKeys = getLocalStorageItemKeys(LS_ORDERS)
+      lsOrderKeys.forEach((lsOrderKey) => window?.localStorage?.removeItem(lsOrderKey))
```
chefilip, line 88: Hmm. it's not recommended that we should have side-effect operation in the reducer. We should keep it pure.

Add/Remove localStorage should be done in middleware or action instead.

memoyil: Done

chefilip (approved):

PR #3540, opened 2022-03-18
fix: Avoid Object.fromEntries error

chefilip:
Screen Shot 2022-03-18 at 08 12 50


memoyil:
Strange, where this issue seen ?

chefilip:
> Strange, where this issue seen ? server side?
Screen Shot 2022-03-18 at 08 16 42

Here is the info @memoyil

chef-jojo (approved):

memoyil:
> > Strange, where this issue seen ? server side?
> 
> Screen Shot 2022-03-18 at 08 16 42
> 
> Here is the info @memoyil

https://caniuse.com/?search=Object.fromEntries

Looks like fromentries support started from 72. Thanks for the fix. Should we have kind of check that shows compatibility to old browsers?

PR #3546, opened 2022-03-18
fix: Lottery selectors not trigger rerender after user tickets fetch

memoyil:
To reproduce: 

1. Go to lottery
2. Buy ticket
3. See Your tickets section on main page not updated that you have to refresh the page to see it or wait for the next cycle of fetchcurrentround

### DIFF
--- a/src/state/lottery/index.ts
+++ b/src/state/lottery/index.ts
@@ -119,8 +119,10 @@ export const LotterySlice = createSlice({
     builder.addCase(
       fetchUserTicketsAndLotteries.fulfilled,
       (state, action: PayloadAction<{ userTickets: LotteryTicket[]; userLotteries: LotteryUserGraphEntity }>) => {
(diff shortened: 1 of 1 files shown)


src/state/lottery/index.ts
```diff
@@ -119,8 +119,10 @@ export const LotterySlice = createSlice({
     builder.addCase(
       fetchUserTicketsAndLotteries.fulfilled,
       (state, action: PayloadAction<{ userTickets: LotteryTicket[]; userLotteries: LotteryUserGraphEntity }>) => {
-        state.currentRound.userTickets.isLoading = false
-        state.currentRound.userTickets.tickets = action.payload.userTickets
+        state.currentRound = {
```
memoyil, line 122: We should pass a new object to trigger rerender

chefilip (approved):

PR #3552, opened 2022-03-20
chore: Add types for base in theme

zengxinhai:
Since types have been defined for theme values. We should enforce this by using these types

### DIFF
--- a/packages/uikit/src/components/Card/CardRibbon.tsx
+++ b/packages/uikit/src/components/Card/CardRibbon.tsx
@@ -7,7 +7,7 @@ interface StyledCardRibbonProps extends CardRibbonProps {
 }
 
 const StyledCardRibbon = styled.div<Partial<StyledCardRibbonProps>>`
-  z-index: 10;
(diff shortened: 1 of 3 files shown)


packages/uikit/src/components/Card/CardRibbon.tsx
```diff
@@ -7,7 +7,7 @@ interface StyledCardRibbonProps extends CardRibbonProps {
 }
 
 const StyledCardRibbon = styled.div<Partial<StyledCardRibbonProps>>`
-  z-index: 10;
+  z-index: ${({ theme }) => theme.zIndices.ribbon};
```
memoyil, line 10: z index was 10 now it is set to 9, is this intended?

zengxinhai: Cause, the zIndex of a Ribbon should be smaller than that of a drodown.
The zIndex of dropdown is set to 10

memoyil: 👍

chef-jojo (approved):

PR #3554, opened 2022-03-21
fix: Hot fix snapshot limit 8 strategies

chef-jojo:
super quick fix where snapshot api now limit to 8 strategies per request

### DIFF
--- a/src/views/Voting/helpers.ts
+++ b/src/views/Voting/helpers.ts
@@ -186,7 +186,7 @@ function calculateVotingPower(scoresList: GetScoresResponse, voters: string[], s
 
     const cakePoolBalance = new BigNumber(userStakeInCakePools[address])
     // calculate cakeBnbLpBalance
-    const cakeBnbLpBalance = new BigNumber(cakeBnbLpBalances[address])
+    const cakeBnbLpBalance = cakeBnbLpBalances[address]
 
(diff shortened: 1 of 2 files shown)


src/views/Voting/helpers.ts
```diff
@@ -181,34 +186,20 @@ function calculateVotingPower(scoresList: GetScoresResponse, voters: string[], s
...
 
     const total = cakeBalance
       .plus(cakeVaultBalance)
       .plus(cakePoolBalance)
       .plus(IFOPoolBalance)
-      .plus(cakeBnbLpBalance)
-      .plus(poolsBalance)
-      .div(TEN_POW_18)
-      .toFixed(18)
+      .plus(new BigNumber(cakeBnbLpBalance).times(TEN_POW_18))
```
memoyil, line 195: I think we should keep the div and tofixed here

chef-jojo: calculation on below
https://github.com/pancakeswap/pancake-frontend/blob/ce75a69b31bb369cc683141cfdbafbca474f237a/src/views/Voting/helpers.ts#L284-L294

memoyil: 👍

memoyil:
LGTM

chefilip:
eslint ?

Chef-Cheems (approved):

PR #3555, opened 2022-03-21
fix: (prediction): Mitigate discrepancy between last price of live round card and expired round card

memoyil:
Although rounds in here are just visual representation and it is not affecting predictions, this issue is complaint by users that seeing different price after round closes. To mitigate the price discrepancy, it will send price fetching right before round closes.

### DIFF
--- a/src/views/Predictions/components/RoundCard/LiveRoundCard.tsx
+++ b/src/views/Predictions/components/RoundCard/LiveRoundCard.tsx
@@ -1,3 +1,4 @@
+import { useEffect } from 'react'
 import { Card, CardBody, Flex, PlayCircleOutlineIcon, Text, useTooltip } from '@pancakeswap/uikit'
 import { useTranslation } from 'contexts/Localization'
 import { NodeRound, NodeLedger, BetPosition } from 'state/types'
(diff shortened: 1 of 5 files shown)


src/views/Predictions/components/RoundCard/LiveRoundCard.tsx
```diff
@@ -22,6 +23,8 @@ interface LiveRoundCardProps {
   bearMultiplier: string
 }
 
+const REFRESH_PRICE_BEFORE_SECONDS_TO_CLOSE = 2
```
memoyil, line 26: 2 seconds is enough to fetch new price

src/views/Predictions/hooks/usePollOraclePrice.ts
```diff
@@ -5,15 +5,15 @@ import { Zero } from '@ethersproject/constants'
 const usePollOraclePrice = (seconds = 10) => {
   const chainlinkOracleContract = useChainlinkOracleContract(false)
   // Can refactor to subscription later
-  const { data: price } = useSWRContract([chainlinkOracleContract, 'latestAnswer'], {
+  const { data: price, mutate } = useSWRContract([chainlinkOracleContract, 'latestAnswer'], {
```
memoyil, line 8: when mutate is called it will bypass deduping interval

chef-jojo (approved):

PR #3557, opened 2022-03-21
fix: (nft): Pancake bunny page still shows non tradable nft on cheapest to buy after it is sold and refactor to swr

memoyil:
To reproduce: 

1. Open any pancake bunny item detail page on two browsers with different wallets
2. Buy the cheapest item from one account (from the main card at the top)
3. Switch to another account
4. Wait for it is updated on graph
5. The item is updated (its current ask price is set to 0) which makes it the cheapest item again that gives an error when buying it

This pr is also includes using swr's to fetch the data instead of redux


memoyil:
@chef-jojo  @Chef-Cheems @chefilip I think it is finally ready :)

chefilip:
> @chef-jojo  @Chef-Cheems @chefilip I think it is finally ready :)

Awesome ! I'm going to review and merge this

chefilip (dismissed):

chefilip (approved):

PR #3564, opened 2022-03-22
fix: (prediction): Timer shows more time remaining than actual and use timeouts to trigger phase changes

memoyil:
To reproduce (Timer shows more time remaining than actual):

1. Go to prediction
2. See round ends before timer counts down to zero
3. Which causes user to think there is still time place prediction

Other Issues fixed:

1. Countdown timer uses next rounds lock timestamp instead of current round close timestamp. Most of the time they are the same but it is nice to use current rounds data instead of next round.
2. Currently timer is out of sync when it reaches at the end beacuse using pure timeout doesn't guarentee it triggers at the correct time. (every clockticking is off by certain milliseconds which results at end of the timer it is out of sync).  In order to solve it more accurate timer implementation is used.
3. Currently we are relying on a re-render to check phase changes which doesn't happen at the time which it should be. One result of this thing is Round buffer phase in open round card not always triggered at the same time and disables placing predictions variable times instead of strict. Calculating phase for live round is also triggered variably which results it switches to that phase not exactly when timer shows it. In order to solve it use timeouts to trigger phase changes.

To review:

Go to prediction see calculating phase, timer and open round placing bet disable buffer (currently 3 block time ~9 seconds) works correctly.


chefilip:
eslint ?

memoyil:
> eslint ?

Done

chefilip:
Hi @memoyil 

Thanks for a great PR. Could you separate a PR `accurateTimer` update so I could take a closer look later? The rest is LGTM.

memoyil:
> Hi @memoyil
> 
> Thanks for a great PR. Could you separate a PR `accurateTimer` update so I could take a closer look later? The rest is LGTM.

Done 👍  https://github.com/pancakeswap/pancake-frontend/pull/3766

chefilip:
Awesome. Thanks

chefilip (approved):

PR #3568, opened 2022-03-23
fix: Voting scores might return empty

chef-jojo:


src/views/Voting/helpers.ts
```diff
@@ -135,7 +135,7 @@ function calculateVotingPowerPools(scoresList: GetScoresResponse, voters: string
   let poolsBalance = new BigNumber(0)
   return voters.reduce<{ [key: string]: BigNumber }>((acc, cur) => {
     for (let i = 0; i < scoresList.length; i++) {
-      const currentPoolBalance = new BigNumber(scoresList[i][cur])
+      const currentPoolBalance = scoresList[i][cur] ? new BigNumber(scoresList[i][cur]) : new BigNumber(0)
```
memoyil, line 138: BIG_ZERO constant can be used

chef-jojo: i don't care now 😭

PR #3572, opened 2022-03-23
fix: (prediction): Round card transition on prediction

memoyil:
Before: 

https://user-images.githubusercontent.com/2213635/159758997-d4f1e562-bb8e-4896-b93c-9336f5f5c9f8.mov

After:

https://user-images.githubusercontent.com/2213635/159759096-333cdef6-7575-47a3-b900-8d97eb6f9066.mov

This fix will also reduce the calls, the size of swr cache of prediction and rerenders

### DIFF
--- a/src/state/predictions/index.ts
+++ b/src/state/predictions/index.ts
@@ -138,6 +138,56 @@ export const fetchRound = createAsyncThunk<ReduxNodeRound, number>('predictions/
   return serializePredictionsRoundsResponse(response)
 })
 
+export const fetchPredictionData = createAsyncThunk<PredictionInitialization, string>(
+  'predictions/fetchPredictionData',
+  async (account = null) => {
+    const { status, currentEpoch, intervalSeconds, minBetAmount } = await getPredictionData()
+    const liveCurrentAndRecent = [currentEpoch, currentEpoch - 1, currentEpoch - 2]
+
+    const roundsResponse = await getRoundsData(liveCurrentAndRecent)
+    const roundData = roundsResponse.reduce((accum, round) => {
+      if (!round) {
+        return accum
+      }
+
+      const reduxNodeRound = serializePredictionsRoundsResponse(round)
+
+      return {
+        ...accum,
+        [reduxNodeRound.epoch.toString()]: reduxNodeRound,
+      }
+    }, {})
+
+    const publicData = {
+      status,
+      currentEpoch,
+      intervalSeconds,
+      minBetAmount,
+      rounds: roundData,
+      ledgers: {},
+      claimableStatuses: {},
+    }
+
+    if (!account) {
+      return publicData
+    }
+
+    const epochs =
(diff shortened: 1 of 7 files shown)


src/state/predictions/index.ts
```diff
@@ -439,6 +489,35 @@ export const predictionsSlice = createSlice({
...
+    // Get static market data
+    builder.addCase(fetchPredictionData.fulfilled, (state, action) => {
+      const { status, currentEpoch, intervalSeconds, minBetAmount, rounds, claimableStatuses, ledgers } = action.payload
+
+      const allRoundData = merge({}, state.rounds, rounds)
+      let newRounds = pickBy(allRoundData, (value, key) => {
+        return Number(key) > state.currentEpoch - PAST_ROUND_COUNT
+      })
+
+      // If the round has change add a new future round
```
memoyil, line 501: Adding new future round after trimming makes the transition correctly. Next fetching cycle the earlist round will be trimmed and it will be handled in here https://github.com/pancakeswap/pancake-frontend/pull/3572/files#diff-f1462bf4b1ea4d9ef14f0ed608ff12ad755d8a05854aace98963662164cd3bb3R43

src/state/predictions/index.ts
```diff
@@ -154,14 +149,37 @@ export const fetchRounds = createAsyncThunk<{ [key: string]: ReduxNodeRound }, n
...
+
+    if (!account) {
+      return publicData
+    }
+
+    const epochs =
+      currentEpoch > PAST_ROUND_COUNT ? range(currentEpoch, currentEpoch - PAST_ROUND_COUNT) : [currentEpoch]
+
+    // Bet data
+    const ledgerResponses = await getLedgerData(account, epochs)
```
chefilip, line 171: shold we use `Promise.all` in this case ?

src/state/predictions/index.ts
```diff
@@ -154,14 +149,37 @@ export const fetchRounds = createAsyncThunk<{ [key: string]: ReduxNodeRound }, n
...
+      ledgers: {},
+      claimableStatuses: {},
+    }
+
+    if (!account) {
+      return publicData
+    }
+
+    const epochs =
+      currentEpoch > PAST_ROUND_COUNT ? range(currentEpoch, currentEpoch - PAST_ROUND_COUNT) : [currentEpoch]
```
chefilip, line 168: Screen Shot 2022-04-06 at 22 07 10
The old logic for range is `earliestEpoch -> currentEpoch + 1`. I'm wondering why they need to fetch future epoch here

src/state/predictions/index.ts
memoyil: No, we wait multicall response inside that function

src/state/predictions/index.ts
memoyil: Because we had an increasing range that was excluding the latest element that is why it was used like that. Now we have a decreasing range so we don't need it anymore.

chefilip: Got it. Thanks

PR #3578, opened 2022-03-25
Revert "fix: (prediction): Remove prediction user state in redux when logout"

Sergio8A1994:
Reverts pancakeswap/pancake-frontend#3573


memoyil:
Have you seen an issue about this fix?

PR #3579, opened 2022-03-25
feat: Adding migration v1 warning banner component

ChefMomota:


src/views/Farms/components/MigrationSticky.tsx
```diff
@@ -0,0 +1,126 @@
...
+            'PancakeSwap is undergoing MasterChef migration. You might see farms being moved to “Finished” and stop giving out rewards.',
+          )}
+        </Text>
+        <Flex flexDirection={isSmallScreen ? 'column' : 'row'}>
+          <StyledLink href="https://twitter.com/pancakeswap" external>
+            <Button width="100%" mr={isSmallScreen ? '0px' : '24px'}>
+              {t('Follow Twitter for updates')}
+            </Button>
+          </StyledLink>
+          <StyledLink href="https://docs.pancakeswap.finance/get-started/connection-guide" external>
```
ChefMomota, line 116: Link will be change before release.

src/views/Farms/components/MigrationSticky.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+const MigrationSticky: React.FC = () => {
+  const { t } = useTranslation()
+  const { isMobile, isXs, isSm, isMd } = useMatchBreakpoints()
+  const [stickPosition, setStickyPosition] = useState<number>(0)
+  const refPrevOffset = useRef(typeof window === 'undefined' ? 0 : window.pageYOffset)
+  const isSmallScreen = isXs || isSm || isMd
+
+  useEffect(() => {
+    const scrollEffect = () => {
```
memoyil, line 74: Might be an overkill, but we can use event throttling to avoid this gets called so many times for performance

src/views/Farms/components/MigrationSticky.tsx
```diff
@@ -0,0 +1,126 @@
...
+  margin-top: 8px;
+  &:hover {
+    text-decoration: initial;
+  }
+`
+
+const MigrationSticky: React.FC = () => {
+  const { t } = useTranslation()
+  const { isMobile, isXs, isSm, isMd } = useMatchBreakpoints()
+  const [stickPosition, setStickyPosition] = useState<number>(0)
```
memoyil, line 69: No need for number type cast

src/views/Farms/components/MigrationSticky.tsx
```diff
@@ -0,0 +1,126 @@
...
+        const topNavHeight = isMobile ? MENU_HEIGHT : MOBILE_MENU_HEIGHT
+        const totalTopMenuHeight = warningBannerHeight ? topNavHeight + topBannerHeight : topNavHeight
+        setStickyPosition(totalTopMenuHeight)
+      }
+      refPrevOffset.current = currentScroll
+    }
+
+    window.addEventListener('scroll', scrollEffect)
+    return () => window.removeEventListener('scroll', scrollEffect)
+  }, [])
```
memoyil, line 94: Add isMobile

src/views/Farms/components/MigrationSticky.tsx
```diff
@@ -0,0 +1,126 @@
...
+        const warningBannerHeight = document.querySelector('.warning-banner')
+        const topBannerHeight = isMobile ? TOP_BANNER_HEIGHT_MOBILE : TOP_BANNER_HEIGHT
+        const topNavHeight = isMobile ? MENU_HEIGHT : MOBILE_MENU_HEIGHT
+        const totalTopMenuHeight = warningBannerHeight ? topNavHeight + topBannerHeight : topNavHeight
+        setStickyPosition(totalTopMenuHeight)
+      }
+      refPrevOffset.current = currentScroll
+    }
+
+    window.addEventListener('scroll', scrollEffect)
```
memoyil, line 92: We need to check window is available (i think crawlers don't have window object right @chefilip ?) before calling addEventListener

src/views/Farms/components/MigrationSticky.tsx
ChefMomota: Why need add isMobile?
code refer to 
https://github.com/pancakeswap/pancake-frontend/blob/develop/src/components/ScrollToTopButton/index.tsx

memoyil: Yes but there is a warning here, "React Hook useEffect has a missing dependency: 'isMobile'. Either include it or remove the dependency array
"

ChefMomota: how about now?

memoyil: 👍

memoyil: Might be an edge case but we need to refresh the event listener when isMobile changes (can happen when orientation changes)

ChefMomota: @memoyil  updated

src/views/Farms/Farms.tsx
```diff
@@ -350,6 +351,7 @@ const Farms: React.FC = ({ children }) => {
           </Button>
         </NextLinkFromReactRouter>
       </PageHeader>
+      <MigrationSticky />
```
Chef-Cheems, line 354: Maybe add `showMigrationBanner` boolean here?

ChefMomota: @Chef-Cheems  updated

src/views/Farms/components/MigrationSticky.tsx
chefilip: Oh. I don't think we need to check. sometimes error throws `window.methodX` called undefined because some browsers don't support `methodX`. In this case `addEventListener` is safe because it's widely supported

src/views/Farms/components/MigrationSticky.tsx
```diff
@@ -0,0 +1,120 @@
...
+      const totalTopMenuHeight = warningBannerHeight ? MENU_HEIGHT + topBannerHeight : MENU_HEIGHT
+      setStickyPosition(totalTopMenuHeight)
+    }
+    refPrevOffset.current = currentScroll
+  }
+
+  useEffect(() => {
+    window.addEventListener('scroll', scrollEffect)
+    return () => window.removeEventListener('scroll', scrollEffect)
+  }, [isMobile])
```
chefilip, line 88: add dependency ?

ChefMomota: Might be an edge case but we need to refresh the event listener when isMobile changes (can happen when orientation changes)

chef-jojo: move the whole `scrollEffect` in to use effect

chef-jojo (approved):

PR #3590, opened 2022-03-27
chore: Avoid totalTickets undefined

chefilip:
Screen Shot 2022-03-27 at 14 43 11


memoyil:
I think we can remove totalTickets from that modal wdyt?

chefilip:
> I think we can remove totalTickets from that modal wdyt?

Done

memoyil:
LGTM

PR #3594, opened 2022-03-28
refactor: (lottery-banner): Use swr for to fetch current lottery without fetching graph data

memoyil:
Currently there is a graph data and additional onchain lottery requests being made which we don't need it when we show lottery banner

### DIFF
--- a/src/views/Home/components/Banners/LotteryBanner.tsx
+++ b/src/views/Home/components/Banners/LotteryBanner.tsx
@@ -4,16 +4,18 @@ import { LotteryStatus } from 'config/constants/types'
 import { useTranslation } from 'contexts/Localization'
 import Image from 'next/image'
 import { memo } from 'react'
+import useSWR from 'swr'
+import BigNumber from 'bignumber.js'
 import { usePriceCakeBusd } from 'state/farms/hooks'
-import { useLottery } from 'state/lottery/hooks'
+import { LotteryResponse } from 'state/types'
 import styled from 'styled-components'
 import { getBalanceNumber } from 'utils/formatBalance'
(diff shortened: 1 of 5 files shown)


src/views/Home/components/Banners/LotteryBanner.tsx
```diff
@@ -60,15 +62,13 @@ export const StyledSubheading = styled(Heading)`
...
-  } = useLottery()
+  const { data } = useSWR<LotteryResponse>(['currentLottery'])
   const cakePriceBusd = usePriceCakeBusd()
-  const prizeInBusd = amountCollectedInCake.times(cakePriceBusd)
+  const prizeInBusd = new BigNumber(data.amountCollectedInCake).times(cakePriceBusd)
   const prizeTotal = getBalanceNumber(prizeInBusd)
   const { t } = useTranslation()
 
-  if (isLotteryLive(status)) {
+  if (isLotteryLive(data.status)) {
```
chefilip, line 71: is it possible if data is `null` ?

memoyil: No, we are checking in here https://github.com/pancakeswap/pancake-frontend/blob/d495334dca7990af15d2b321cde0da35e4eb9532/src/views/Home/components/Banners/hooks/useIsRenderLotteryBanner.tsx we only render Lottery Banner when it is fetched. And fetchLottery returns object when there is an error too. https://github.com/pancakeswap/pancake-frontend/blob/01063288e7ca48f3ae1037e4d23b1cc765220404/src/state/lottery/helpers.ts#L76

chefilip (approved):

PR #3595, opened 2022-03-28
perf: (translation): Avoid regex call on each translate call

memoyil:


src/contexts/Localization/Provider.tsx
```diff
@@ -10,9 +11,37 @@ const initialState: ProviderState = {
   currentLanguage: EN,
 }
 
+const includesVariableRegex = new RegExp(/%\S+?%/, 'gm')
+
+const translatedTextIncludesVariable = (translatedText: string): boolean => {
```
chefilip, line 16: In terms of performance, I think we could add lodash memoize for this function

memoyil: This func will run per distinct key only when it is loaded to language map, so it will not run with same key multiple times. imho we don't need to memoize this

chefilip (approved):

src/contexts/Localization/Provider.tsx
memoyil: But if we keep the old approach and  extract that include variable check into an memoize function only would be a cleaner solution. Wdyt?

PR #3600, opened 2022-03-29
fix: (liquidity): Farm pairs that doesn't contain tracked base is not visible when removing liquidity

memoyil:
The pairs that doesn't contain tracked base such as Duet-Cake or Gmt-usdc not visible by default in the list when removing liquidity therefore user has to import it.

### DIFF
--- a/src/state/user/hooks/index.tsx
+++ b/src/state/user/hooks/index.tsx
@@ -1,6 +1,7 @@
 import { ChainId, Pair, Token } from '@pancakeswap/sdk'
 import { differenceInDays } from 'date-fns'
 import flatMap from 'lodash/flatMap'
+import farms from 'config/constants/farms'
 import { useCallback, useMemo } from 'react'
 import { useDispatch, useSelector } from 'react-redux'
 import { CHAIN_ID } from 'config/constants/networks'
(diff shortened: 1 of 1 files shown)


chef-jojo:
We can add cake to `BASES_TO_TRACK_LIQUIDITY_FOR`

memoyil:
> We can add cake to `BASES_TO_TRACK_LIQUIDITY_FOR`

Yes but we have another base tokens in farm pairs such as (eth and usdc) that we currently dont check and if we add all of them we might end up with very big list for multicall request. 

I dont know the limits or the server side consequences of big multicall request thats i followed this approach. If you want i can add them to the list. What do you prefer?

memoyil:
@chef-jojo Imo it is a good idea to add farm pairs regardless of this fix to ensure these liquidities will be shown by default on remove liquidity page. Currently if a LP contains a the token is not listed in default lists yet, it is not shown in remove liquditiy page.

chef-jojo:
Yea I think it's good for now. In the future, we should consider add cake as base pair to fetch though

chef-jojo (approved):

chef-jojo (approved):

PR #3605, opened 2022-03-30
fix: Register token with token list logo uri

chef-jojo:


src/components/TransactionConfirmationModal/index.tsx
```diff
@@ -90,7 +91,14 @@ export function TransactionSubmittedContent({
               variant="tertiary"
               mt="12px"
               width="fit-content"
-              onClick={() => registerToken(token.address, token.symbol, token.decimals)}
+              onClick={() =>
+                registerToken(
+                  token.address,
+                  token.symbol,
+                  token.decimals,
+                  token instanceof WrappedTokenInfo && token.logoURI,
```
memoyil, line 99: This will return false if token is not WrappedTokenInfo type where method expects string only. I think it is better to pass null or undefined instead

chef-jojo: check here
https://github.com/pancakeswap/pancake-frontend/pull/3605/files#diff-65f9606ab5af155b8d90242f605a3fe1fa053d6fcf49183e5ee0ec1014ebf11fR85

memoyil: Yes, currently there is no issue. Might be problem when we want to switch strict mode (although it is very unlikely in near future)

chef-jojo: I don't think it's a big deal but yea why not

memoyil:
LGTM

PR #3618, opened 2022-04-01
feat: List YEL farms

ChefMomota:
YEL-BNB LP address: `0x8290d3CA64f712de9FB7220353dAa55bf388F3A3`


src/config/constants/farms.ts
```diff
@@ -41,6 +41,38 @@ const farms: SerializedFarmConfig[] = [
...
+  {
+    pid: 386,
+    lpSymbol: 'HOTCROSS-BNB LP',
+    lpAddresses: {
+      97: '',
+      56: '0xf23bad605e94de0e3b60c9718a43a94a5af43915',
+    },
+    token: serializedTokens.hotcross,
+    quoteToken: serializedTokens.wbnb,
+  },
```
ChefMomota, line 75: Do we need add `isCommunity: true` ?

chefilip: Yeah. Then we need to turn it back to Core when the auction farm is done.

ChefMomota: @chefilip updated

chefilip (approved):

PR #3622, opened 2022-04-02
fix: User added tokens logos is not visible on token search modal list

memoyil:
To reproduce: 

1. Go to swap
2. Open currency search
3. Add one token from lists (coingecko, cmc) that is not currently on the list
4. See logo is not visible question mark shown instead


src/components/CurrencyInputPanel/index.tsx
```diff
@@ -4,7 +4,7 @@ import styled from 'styled-components'
 import { registerToken } from 'utils/wallet'
 import { isAddress } from 'utils'
 import { useTranslation } from 'contexts/Localization'
-import { WrappedTokenInfo } from 'state/lists/hooks'
```
memoyil, line 7: These are changes are to avoid circular dependency

chef-jojo (approved):

PR #3629, opened 2022-04-04
fix: (lottery banner): Missing current lottery fetch causes banner always shows preparing

memoyil:
To fix the blinking issue swiper observer is used.


chef-jojo (approved):

memoyil:
@chef-jojo I used swiper observer, so we don't need to re-mount when banner list changes, and we can use is render lottery banner hook again.

Chef-Yogi (approved):

PR #3634, opened 2022-04-05
refactor: Bump up graphql and graphql-request

memoyil:


Chef-Cheems:
I think `graphql` is peer dependency of `graphql-request`, you can't remove it - https://github.com/prisma-labs/graphql-request/blob/master/package.json#L52-L54 🤔

memoyil:
> I think `graphql` is peer dependency of `graphql-request`, you can't remove it - https://github.com/prisma-labs/graphql-request/blob/master/package.json#L52-L54 🤔

Yes, I added it back again with the updated version. I figured out that it was working because gelato limit order library has it. Should we stay in version 15 or update to 16 ?

PR #3615, opened 2022-03-31
fix: (nft): Token detail and user profile pages not show updated market data

memoyil:
Because of the delays on graph api, users will not able to see actual market data on token detail and user profile page for long times.

To reproduce on token detail page:

1. Go to any collection 
2. Buy one item 
3. After successful transaction it still shows the old price and it will be still on sale

To reproduce issues on profile page:

1. Change sell price of nft that is on market
2. See market data is not updated if graph data is delayed

------------------------

1. Remove nft from the market
2. See it shows the same nft twice on the list until graph api is updated

------------------------

1. List nft to the market
2. User has to wait until graph api is updated (and until that time that nft is not shown in the user nft list)

PR #3635, opened 2022-04-05
fix: (nft): Pancake Bunny Nft cheapest price and for sale card shows obsolete data which causes error when buy

memoyil:
To reproduce:

1. Go pancake bunny token page
2. Buy one nft from the lowest price card at the top
3. See nft is still there after purchase if graphql is delayed
4. If other user tries to buy it it will get an error.

With this and https://github.com/pancakeswap/pancake-frontend/pull/3615 changes, there will be less dependency to the data from graphql


chefilip (approved):

chefilip:
Screen Shot 2022-04-11 at 08 52 21

Hi !

@memoyil  Could you help me check the mobile version ? It seems there is a mismatch in navigation

chefilip:
Screen Shot 2022-04-11 at 09 09 13
Hi @memoyil  !

I think the navigation table is mismatched. for 100 items, it should be `1 of 20` ?

memoyil:
> 

No, We fetch the itemsPerPage multiplied by 10 thus on mobile it is 50.

memoyil:
> Screen Shot 2022-04-11 at 09 09 13
> Hi @memoyil !
> 
> I think the navigation table is mismatched. for 100 items, it should be `1 of 20` ?

No, We fetch the itemsPerPage multiplied by 10 thus on mobile it is 50.

chefilip (changes requested):
There is a weird behavior in navigation

src/views/Nft/market/hooks/usePancakeBunnyCheapestNfts.tsx
```diff
@@ -7,23 +7,37 @@ import {
...
+  const nftsMarketTokenIds = nftsMarket.map((marketData) => marketData.tokenId)
+  const lowestPriceUpdatedBunny = await getLowestUpdatedToken(pancakeBunniesAddress.toLowerCase(), nftsMarketTokenIds)
+
+  const cheapestBunnyOfAccount = nftsMarket
+    .filter((marketData) => marketData.tokenId === lowestPriceUpdatedBunny?.tokenId)
+    .map((marketData) => {
+      const apiMetadata = getMetadataWithFallback(nftMetadata.data, marketData.otherId)
+      const attributes = getPancakeBunniesAttributesField(marketData.otherId)
+      const bunnyToken = combineApiAndSgResponseToNftToken(apiMetadata, marketData, attributes)
+      const updatedPrice = formatBigNumber(lowestPriceUpdatedBunny.currentAskPrice)
```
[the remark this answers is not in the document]
memoyil: We are handling in filter stage 

`.filter((marketData) => marketData.tokenId === lowestPriceUpdatedBunny?.tokenId)`

chefilip:
CleanShot 2022-04-11 at 17 46 34

Hi @memoyil  !

The navigtion seems recursively added.

chefilip:
Screen Shot 2022-04-11 at 18 01 41

Could we take the supply count to get the total page ? @memoyil

src/views/Nft/market/Collection/IndividualNFTPage/PancakeBunnyPage/ForSaleTableCard/index.tsx
```diff
@@ -114,27 +153,23 @@ const ForSaleTableCard: React.FC<ForSaleTableCardProps> = ({ bunnyId, nftMetadat
...
-                    setPage(page - 1)
-                  }
+                  switchPage(internalPage === 1 ? internalPage : internalPage - 1)
                 }}
               >
-                <ArrowBackIcon color={page === 1 ? 'textDisabled' : 'primary'} />
+                <ArrowBackIcon color={internalPage === 1 ? 'textDisabled' : 'primary'} />
               </Arrow>
-              <Text>{t('Page %page%', { page })}</Text>
+              <Text>{t('Page %page% of %maxPage%', { page: internalPage, maxPage: maxInternalPage })}</Text>
```
chefilip, line 161: I think `maxPage` shouldn't be displayed because the number isn't the maximum page which might make users confused.

src/views/Nft/market/Collection/IndividualNFTPage/PancakeBunnyPage/ForSaleTableCard/index.tsx
```diff
@@ -53,26 +57,61 @@ const ForSaleTableCard: React.FC<ForSaleTableCardProps> = ({ bunnyId, nftMetadat
...
     setDirection,
+    isFetchingNfts,
     isLastPage,
     isValidating,
-  } = usePancakeBunnyOnSaleNfts(bunnyId, nftMetadata, itemsPerPage)
+  } = usePancakeBunnyOnSaleNfts(bunnyId, nftMetadata, itemsPerPage * 10)
 
-  const { t } = useTranslation()
-  const { theme } = useTheme()
+  const [internalPage, setInternalPage] = useState(1)
```
chefilip, line 65: could you help me understand why we need `internalPage` while we already have `page` from `usePancakeBunnyOnSaleNfts`

chefilip: I assume it would be used to maintain the updated list ( from SC ) and the list ( from Subgraph ). 

Instead of trying the maintain the table state, why don't we maintain the row state only. For example, when we load the list from Subgraph, we check that list status from SC. If items in the Subgraph are no longer tradable, we just disable the `Buy` button. 

In this way, we don't have to maintain another table

memoyil: We need to fetch in batches then chunk it on the view side otherwise if nfts sold or price changed individual pages will show less items then intended. With current behaviour only end of the batch we will show less items to the user.

src/views/Nft/market/Collection/IndividualNFTPage/PancakeBunnyPage/ForSaleTableCard/index.tsx
memoyil: Done

src/views/Nft/market/Collection/IndividualNFTPage/PancakeBunnyPage/ForSaleTableCard/index.tsx
memoyil: > I assume it would be used to maintain the updated list ( from SC ) and the list ( from Subgraph ).
> 
> Instead of trying the maintain the table state, why don't we maintain the row state only. For example, when we load the list from Subgraph, we check that list status from SC. If items in the Subgraph are no longer tradable, we just disable the `Buy` button.

This would not cover the price changes which user would have to wait delayed subgraph api after making a change in listing price. Thats the other reason we need to fetch in batch then order by price in view side

memoyil: Done

chefilip (approved):

PR #3646, opened 2022-04-07
chore: Bump up ajv and tokenlist

memoyil:


memoyil:
No need for now, and it will increase the size of bundle

PR #3648, opened 2022-04-07
fix: Manage nft cards still shows user nfts after logout

memoyil:
Root cause is laggy middleware shows the last data after key changes. We should keep this behaviour and resolve the account changes in the fetcher

To reproduce:

1. Go to pb nft token detail page that account has bunny
2. See nfts on manage yours
3. Logout
4. See it is still visible

### DIFF
--- a/src/state/nftMarket/helpers.ts
+++ b/src/state/nftMarket/helpers.ts
@@ -6,7 +6,6 @@ import erc721Abi from 'config/abi/erc721.json'
 import range from 'lodash/range'
 import uniq from 'lodash/uniq'
 import { pancakeBunniesAddress } from 'views/Nft/market/constants'
-import { isAddress } from 'utils'
 import {
   ApiCollection,
   ApiCollections,
--- a/src/state/nftMarket/helpers.ts
+++ b/src/state/nftMarket/helpers.ts
@@ -987,8 +987,6 @@ export const getCompleteAccountNftData = async (
   collections: ApiCollections,
   profileNftWithCollectionAddress?: TokenIdWithCollectionAddress,
 ): Promise<NftToken[]> => {
-  if (!isAddress(account)) return []
-
   const walletNftIdsWithCollectionAddress = await fetchWalletTokenIdsForCollections(account, collections)
   if (profileNftWithCollectionAddress?.tokenId) {
     walletNftIdsWithCollectionAddress.unshift(profileNftWithCollectionAddress)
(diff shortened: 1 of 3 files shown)


chefilip (approved):

src/state/nftMarket/helpers.ts
```diff
@@ -986,6 +987,8 @@ export const getCompleteAccountNftData = async (
   collections: ApiCollections,
   profileNftWithCollectionAddress?: TokenIdWithCollectionAddress,
 ): Promise<NftToken[]> => {
+  if (!isAddress(account)) return []
```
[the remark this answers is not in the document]
memoyil: I don't want to make requests with undefined/invalid account address so i added that shortcut

src/views/Nft/market/hooks/useNftsForAddress.tsx
```diff
@@ -28,7 +27,7 @@ const useNftsForAddress = (account: string, profile: Profile, isProfileFetching:
   }, [profileNftTokenId, profileNftCollectionAddress, hasProfileNft])
 
   const { status, data, mutate } = useSWR(
-    !isProfileFetching && !isEmpty(collections) && isAddress(account) ? [account, 'userNfts'] : null,
+    !isProfileFetching && !isEmpty(collections) ? [account, 'userNfts'] : null,
```
chefilip, line 30: I think we should keep the check `isAddress(account)` here to avoid extra call.

src/views/Nft/market/hooks/useNftsForAddress.tsx
```diff
@@ -1,16 +1,23 @@
...
-import { isAddress } from 'utils'
 import useSWR from 'swr'
 import { FetchStatus } from 'config/constants/types'
 import { laggyMiddleware } from 'hooks/useSWRContract'
+import usePrevious from 'hooks/usePreviousValue'
+import { isAddress } from 'utils'
 
 const useNftsForAddress = (account: string, profile: Profile, isProfileFetching: boolean) => {
   const { data: collections } = useGetCollections()
+  const resetLaggyRef = useRef(null)
```
memoyil, line 15: This should be in order, not done by useEffect

chefilip:
I like this solution. It's much easier to understand. Awesome work ! Could you help me address eslint ? Then I could merge it. Thanks

memoyil:
> I like this solution. It's much easier to understand. Awesome work ! Could you help me address eslint ? Then I could merge it. Thanks @memoyil

Done 👍

PR #3663, opened 2022-04-11
fix: Previous account data shown when change account

memoyil:
To reproduce:

1. Go to swap
2. Do some transactions
3. Click recent transactions
4. See the transaction
5. Change account from wallet
6. See the same transaction still there


chef-jojo:
Can we merge it into `useInactiveListener`?

memoyil:
> Can we merge it into `useInactiveListener`?

Done

chef-jojo (approved):
👍

PR #3675, opened 2022-04-12
chore: Use Deutsch as a language label

ChefNina:
"Datch" seems to be typo of "Dach".
"Dach" means the areas where people speak German (Deutschland+Austria+Confoederatio Helvetica).
As a language label, "Deutsch"(German) is more clear.


chef-jojo (approved):
🐥

PR #3711, opened 2022-04-19
chore: Avoid evaluation error in name prop

chefilip:


memoyil:
https://github.com/chefilip/pancake-frontend/blob/b1e20433ca494d77f8b322b387013c0c37e79423/src/views/Nft/market/components/Activity/ActivityRow.tsx#L69

we can remove onclick if nft is null

memoyil:
LGTM

chef-jojo (approved):

PR #3723, opened 2022-04-20
fix: Farm table history integration test

memoyil:


src/views/Farms/Farms.tsx
```diff
@@ -312,7 +312,7 @@ const Farms: React.FC = ({ children }) => {
   })
 
   const renderContent = (): JSX.Element => {
-    if (viewMode === ViewMode.TABLE && rowData.length) {
```
chefilip, line 315: we should change the test to make the code passed instead of changing the code to make it passed.

memoyil: Done

chefilip (approved):

PR #3739, opened 2022-04-21
fix: Default Hot order in farms

memoyil:


src/views/Farms/Farms.tsx
```diff
@@ -232,6 +235,10 @@ const Farms: React.FC = ({ children }) => {
           return orderBy(farms, (farm: FarmWithStakedValue) => Number(farm.liquidity), 'desc')
         case 'latest':
           return orderBy(farms, (farm: FarmWithStakedValue) => Number(farm.pid), 'desc')
+        case 'hot': {
```
chefilip, line 238: could we do this logic in the `contants/farms.ts` file ?

chef-jojo: assign to @ChefMomota. Let's make it static

memoyil: 👍 I will close this pr

PR #3740, opened 2022-04-21
feat: skeleton with transition

Chef-Yogi:
demo:
skeletonAfter

### DIFF
--- a/packages/uikit/src/components/Skeleton/Skeleton.tsx
+++ b/packages/uikit/src/components/Skeleton/Skeleton.tsx
@@ -1,7 +1,15 @@
-import React from "react";
+import { AnimatePresence, domAnimation, LazyMotion, m as Motion } from "framer-motion";
+import React, { useRef } from "react";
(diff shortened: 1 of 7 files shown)


chefilip (approved):

chef-jojo:
since it's called SkeletonV2, can you give us more suggestions on when we are going to use it?
What will happen if we replace all skeleton to v2?

Chef-Yogi:
@chef-jojo 
the original props have not changed, add some new props to make the transition between the Skeleton and Content easier, considering the mechanism, I think we should partially switch to Skeletonv2 and check every place we switched

Chef-Yogi:
original way
`<h1 style={{ width: 200 }}>{loading ? <Skeleton {...args} /> : "H1"}</h1>;`

V2
`    <SkeletonV2 {...args} isDataReady={!loading} width={200}>
      <h1>H1</h1>
    </SkeletonV2>`

chef-jojo (approved):

PR #3754, opened 2022-04-22
fix(pool): Avoid NAN usd

chef-jojo:


memoyil:
@chef-jojo We should fix the card as well imo

https://github.com/pancakeswap/pancake-frontend/blob/f926fdf08c339c184c39328ec71afb3f6ecb9a8d/src/views/Pools/components/PoolCard/CardActions/HarvestActions.tsx#L58

@chefilip @chef-jojo Maybe we should use userdataloaded and isvaultloading flags together to determine all related data is available. 

The second refactor could be remove the seperate fetching of cake vault and put it inside the pool fetchings

wdyt?

chef-jojo:
Thanks for flagging. I will add card later.

I think we already use userdataloaded there, it's just preventing USD price in case not displaying properly.

chefilip (approved):

PR #3757, opened 2022-04-22
fix: Excessive space in pool table stake component

memoyil:
Before:
image

After:
image


chefilip:
After the lock pool release, I think this fix is no longer needed.

memoyil:
@chefilip We still have theat excessive space on other pools stake component

chef-jojo (approved):

PR #3759, opened 2022-04-22
fix: ANKR APR

Chef-Cheems:


src/config/constants/priceHelperLps.ts
```diff
@@ -19,6 +19,16 @@ const priceHelperLps: SerializedFarmConfig[] = [
     token: tokens.qsd,
```
memoyil, line 19: qsd is no longer needed we can remove it from here

memoyil:
@Chef-Cheems I moved price helper fetch to pools so we don't need to fetch if user outside of the pools page. 

https://github.com/pancakeswap/pancake-frontend/pull/3760

If you like this solution we can apply

PR #3760, opened 2022-04-22
refactor: Move price helper fetch to pools

memoyil:


Chef-Cheems (approved):
Thanks! We've been thinking what to do with changing 200+ lines 😅

PR #3763, opened 2022-04-23
fix: Language selector dropdown appears when scrolling in mobile

memoyil:
In mobile especially smaller screen size phones when scrolling down language dropdown appears. 

https://user-images.githubusercontent.com/2213635/164975427-f7c3f3ee-6838-4786-ad93-ffcfbfd7029b.mp4


chefilip:
Since the issue only occurs on small screen, could you add a condition that the fix is only applied for small screen. Thanks.

memoyil:
> Since the issue only occurs on small screen, could you add a condition that the fix is only applied for small screen. Thanks.

Done 👍

chefilip (approved):

PR #3782, opened 2022-04-26
fix(locked-pool): Remove user locked UI when logging out

chefilip:


memoyil:
Imo we can remove the state via resetUserState action of poolReducer

chefilip:
> Imo we can remove the state via resetUserState action of poolReducer. The UI here is just an extra guard.

We already have this logic in the reducer.
Screen Shot 2022-04-26 at 18 47 09

PR #3823, opened 2022-05-03
perf: Use promise all to reduce execution time of pools user data fetch

memoyil:
It reduces user data execution time from ~3s to ~1s


chefilip:
Great PR! Thanks for your contribution.

chefilip (approved):

PR #3830, opened 2022-05-04
refactor(prediction): Convert Global State to Local State and add Config Provider

chefilip:
Purpose: Enable to have multiple prediction token version in the future.
Tasks: 
- Convert Prediction Global State into Local State
- Generalize the Prediction Config


src/pages/prediction/leaderboard.tsx
```diff
@@ -1,3 +1,18 @@
...
+
 import PredictionsLeaderboard from '../../views/Predictions/Leaderboard'
 
-export default PredictionsLeaderboard
+const store = makeStore(reducers, initialState, config)
+
+export default () => {
+  return (
+    <ConfigProvider>
+      <LocalReduxProvider store={store}>
```
chef-jojo, line 13: we don't need to make different store for prediction and leaderboard page, right?

src/state/predictions/hooks.ts
```diff
@@ -136,6 +138,31 @@ export const useGetOrFetchLeaderboardAddressResult = (account: string) => {
...
 
-export const useGetSelectedAddress = () => {
-  return useSelector((state: State) => state.predictions.leaderboard.selectedAddress)
+export const useGetSelectedAddress = (): string => {
+  return useSelector((state: PredictionsState) => state.leaderboard.selectedAddress)
+}
+
+// Because Modal Component is rendered outside the Prediction Page contexts
+// We have to pass local state as props instead of retrieving directly in component
+export const useStatModalProps = (account?: string) => {
```
chef-jojo, line 147: umm can we inject the provider in _app.tsx?

src/pages/prediction/leaderboard.tsx
chefilip: Good catch.

src/state/predictions/hooks.ts
chefilip: we couldn't achieve various prediction config if Injecting Config Provider in `_app.tsx`. The flaw is in the Modal component design. It should use Portal to keep the Context.

chef-jojo (approved):

PR #3833, opened 2022-05-04
fix: Contract hooks cause unnecessary rerenders and requests even signer not used

memoyil:
https://github.com/pancakeswap/pancake-frontend/blob/2d1c22c99035d319cf118019d0968776e7295b0c/src/components/GlobalCheckClaimStatus/index.tsx#L73

When loading the page for the first time, it renders twice since reference from contract hook changes (because of wallet login) even the signer is not used. In here it makes requests twice unnecessarily.

### DIFF
--- a/src/hooks/useContract.ts
+++ b/src/hooks/useContract.ts
@@ -292,7 +292,7 @@ function useContract<T extends Contract = Contract>(
   )
 
   const canReturnContract = useMemo(
-    () => !(!address || !ABI || withSignerIfPossible ? !library : false),
+    () => (address && ABI && withSignerIfPossible ? library : true),
     [address, ABI, library, withSignerIfPossible],
   )
 
--- a/src/hooks/useContract.ts
+++ b/src/hooks/useContract.ts
@@ -50,6 +50,8 @@ import { getContract, getProviderOrSigner } from '../utils'
 
 import { IPancakePair } from '../config/abi/types/IPancakePair'
 
+const EMPTY_ARRAY = Object.freeze([])
+
 /**
  * Helper hooks to get specific contracts (by ABI)
  */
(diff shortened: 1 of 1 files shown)


src/hooks/useContract.ts
```diff
@@ -68,18 +70,20 @@ export const useERC20 = (address: string, withSignerIfPossible = true) => {
   const { library, account } = useActiveWeb3React()
   return useMemo(
     () => getBep20Contract(address, withSignerIfPossible ? getProviderOrSigner(library, account) : null),
-    [account, address, library, withSignerIfPossible],
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
chefilip, line 73: Could we do this?
```
const signer = useMemo(() => withSignerIfPossible ? getProviderOrSigner(library, account) : null, [withSignerIfPossible, library, account])

return useMemo(() => getContract(address, signer), [address, signer])
```

memoyil: Done 👍

src/hooks/useContract.ts
```diff
@@ -277,16 +286,20 @@ function useContract<T extends Contract = Contract>(
...
     if (!address || !ABI || !library) return null
     try {
-      return getContract(address, ABI, withSignerIfPossible ? getProviderOrSigner(library, account) : null)
+      return getContract(address, ABI, signer)
     } catch (error) {
       console.error('Failed to get contract', error)
       return null
     }
-  }, [address, ABI, library, withSignerIfPossible, account]) as T
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
chefilip, line 301: do we still need this comment?

memoyil: Removed 👍

chefilip (approved):

PR #3848, opened 2022-05-05
refactor: Use promise all in nft refetch

memoyil:


memoyil:
It should be ordered thus we should not use promise all here

PR #3859, opened 2022-05-08
fix: Modal bottom cannot be tapped on safari mobile due to safari bottom bar expanding

memoyil:


packages/uikit/src/widgets/Modal/styles.tsx
```diff
@@ -49,6 +50,7 @@ export const ModalContainer = styled(Box)<{ minWidth: string }>`
   border-radius: 32px;
   width: 100%;
   max-height: 100vh;
```
memoyil, line 52: This will stay as fallback if calc is not supported

chef-jojo (approved):

PR #3875, opened 2022-05-09
feat: (limit-orders): Add import token warning modal

memoyil:
To review:

1. Go to http://pancakeswap.finance/limit-orders?outputCurrency=0x8f7ff372869a25eded0c094e854e66c3a471de3d
(which is a token that is not in the default list)
3. See import token doesn't appear


src/views/LimitOrders/index.tsx
```diff
@@ -44,8 +48,38 @@ const LimitOrders = () => {
...
+    return (
+      urlLoadedTokens &&
+      urlLoadedTokens.filter((token: Token) => {
+        return !(token.address in defaultTokens)
+      })
+    )
+  }, [defaultTokens, urlLoadedTokens])
+
+  const [onPresentImportTokenWarningModal] = useModal(
+    <ImportTokenWarningModal tokens={importTokensNotInDefault} onCancel={() => router.push('/limit-orders')} />,
```
chef-jojo, line 74: let's add a modal key for that

memoyil: Done

chef-jojo (approved):

PR #3893, opened 2022-05-11
feat: Make isCommunity data fetched from farm auction on chain contract

memoyil:


src/state/farms/fetchFarmsWithAuctions.ts
```diff
@@ -0,0 +1,59 @@
...
+      },
+    ],
+    { requireSuccess: false },
+  )
+  const blocksUntilBlock = auctionData.endBlock.toNumber() - currentBlock
+  if (blocksUntilBlock < 0) {
+    const secondsUntilStart = blocksUntilBlock * BSC_BLOCK_TIME
+    const currentAuctionEndDate = add(new Date(), { seconds: secondsUntilStart })
+    if (
+      Math.abs(now - add(currentAuctionEndDate, { days: FARM_AUCTION_HOSTING_IN_DAYS }).getTime()) /
```
chefilip, line 37: `now` here will no longer be `the now` if `await` functions above are executed slowly.

src/state/farms/fetchFarmsWithAuctions.ts
```diff
@@ -0,0 +1,59 @@
...
+        (24 * 3600 * 1000) >
+      FARM_AUCTION_HOSTING_IN_DAYS
```
chefilip, line 39: I think using `seconds` format would make it more readable.

src/state/farms/fetchFarmsWithAuctions.ts
```diff
@@ -0,0 +1,59 @@
...
+        params: [currentAuctionId, 0, 500],
+      },
+    ],
+    { requireSuccess: false },
+  )
+  const blocksUntilBlock = auctionData.endBlock.toNumber() - currentBlock
+  if (blocksUntilBlock < 0) {
+    const secondsUntilStart = blocksUntilBlock * BSC_BLOCK_TIME
+    const currentAuctionEndDate = add(new Date(), { seconds: secondsUntilStart })
+    if (
```
chefilip, line 36: Could we write like this:
```
secondsUntilStart > FARM_AUCTION_HOSTING_IN_SECONDS
```

src/state/farms/fetchFarmsWithAuctions.ts
memoyil: Done

src/state/farms/fetchFarmsWithAuctions.ts
memoyil: Done

src/state/farms/fetchFarmsWithAuctions.ts
memoyil: Done

chefilip (approved):

PR #3894, opened 2022-05-11
fix: wrong query condition in token transaction

vunhutien:
fix: wrong query condition for token transaction info page.


chef-jojo (approved):
Nice catch!

PR #3896, opened 2022-05-11
feat(cake-prediction): Add Cake Prediction

chefilip:


src/views/Predictions/components/RoundCard/SetPositionCard.tsx
```diff
@@ -73,19 +76,38 @@ const getValueAsEthersBn = (value: string) => {
...
+
+  const { isVaultApproved, setLastUpdated } = useCakeApprovalStatus(token.symbol === 'CAKE' ? predictionsAddress : null)
+  const { handleApprove, pendingTx } = useCakeApprove(
+    setLastUpdated,
+    predictionsAddress,
+    t('You can now start prediction'),
+  )
+
+  // BNB prediction doesn't need approval
+  const doesCakeApprovePrediction = token.symbol === 'BNB' || isVaultApproved
```
memoyil, line 108: Why don't we use useApproveConfirmTransaction hook?

chefilip: AFAIK, `useApproveConfirmTransaction` is low-level hook which is used for more generic cases. If we use `useApproveConfirmTransaction` for such specific cases, the duplication would increase.

src/views/Predictions/context/config.ts
```diff
@@ -0,0 +1,20 @@
+import addresses from 'config/constants/contracts'
+
+import { GRAPH_API_PREDICTION_CAKE, GRAPH_API_PREDICTION } from 'config/constants/endpoints'
+import { getAddress } from 'utils/addressHelpers'
+import tokens from 'config/constants/tokens'
+
+export default {
+  BNB: {
+    address: getAddress(addresses.predictions),
+    api: GRAPH_API_PREDICTION,
```
memoyil, line 10: name can have bnb suffix

chefilip: Good idea. I will do it after release because it requires a env config on Vercel.

chef-jojo (dismissed):

PR #3923, opened 2022-05-14
test: added header related cypress tests

tamasmagyar:
- Added header related tests and the related data-cy attributes


tamasmagyar:
@chef-jojo can you review please?

chefilip:
Thanks for your contribution. I have some feedback:
1. I'd prefer `data-test` name because we might replace cypress in the furture.
2. I think the test case is basic and there is no need to add automation test.

PR #3933, opened 2022-05-16
fix: (cake pool): Expected roi calculation in other languages

memoyil:
Root cause of the issue is, the return value is converted back to number which causes the issue if user device language is other than english

To reproduce

1. Go to locked pool with a device language german
2. Click add cake
3. Enter value to make expected roi higher than 1k
4. See it shows wrong value

### DIFF
--- a/src/utils/formatBalance.ts
+++ b/src/utils/formatBalance.ts
@@ -26,6 +26,9 @@ export const getFullDisplayBalance = (balance: BigNumber, decimals = 18, display
   return getBalanceAmount(balance, decimals).toFixed(displayDecimals)
 }
 
+/**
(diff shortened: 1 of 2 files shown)


src/utils/formatBalance.ts
```diff
@@ -26,6 +26,9 @@ export const getFullDisplayBalance = (balance: BigNumber, decimals = 18, display
   return getBalanceAmount(balance, decimals).toFixed(displayDecimals)
 }
 
+/**
+ * Don't use this string value to convert back to number.
```
chefilip, line 30: Could you add a reason why here?

chefilip (approved):

PR #3934, opened 2022-05-16
refactor(mono): Move doc folder

chefilip:


chef-jojo:
Don’t think it needs to be a app

chefilip:
> Don’t think it needs to be a app

Yeah I agreed. Just follow a standard. What is your suggestion?

chef-jojo:
Let's keep it as it is, it's not a doc site with package.json, just some markdown.

PR #3945, opened 2022-05-17
fix: Cake Pool slider reset after ROI

ChefMomota:
https://user-images.githubusercontent.com/98292246/168784066-518878e9-e80a-48fc-a487-34d8a5c8a554.mov

### DIFF
--- a/src/views/Pools/components/LockedPool/Common/BalanceField.tsx
+++ b/src/views/Pools/components/LockedPool/Common/BalanceField.tsx
@@ -37,25 +37,21 @@ const BalanceField: React.FC<PropsType> = ({
   const { userNotEnoughCake, notEnoughErrorMessage } = useUserEnoughCakeValidator(lockedAmount, stakingTokenBalance)
 
   useEffect(() => {
-    const lockedAmountBigNumber = new BigNumber(lockedAmount)
-    if (lockedAmountBigNumber.gt(0)) {
-      handleStakeInputChange(lockedAmount)
+    const amount = new BigNumber(lockedAmount)
+    if (amount.gt(0)) {
+      const convertedInput = amount.multipliedBy(BIG_TEN.pow(stakingDecimals))
+      const percentage = Math.floor(convertedInput.dividedBy(stakingMax).multipliedBy(100).toNumber())
+      setPercent(percentage > 100 ? 100 : percentage)
+    } else {
+      setPercent(0)
     }
-  }, [])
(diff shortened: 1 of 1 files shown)


src/views/Pools/components/LockedPool/Common/BalanceField.tsx
```diff
@@ -36,6 +36,13 @@ const BalanceField: React.FC<PropsType> = ({
   const [percent, setPercent] = useState(0)
   const { userNotEnoughCake, notEnoughErrorMessage } = useUserEnoughCakeValidator(lockedAmount, stakingTokenBalance)
 
+  useEffect(() => {
+    const lockedAmountBigNumber = new BigNumber(lockedAmount)
+    if (lockedAmountBigNumber.gt(0)) {
+      handleStakeInputChange(lockedAmount)
+    }
+  }, [])
```
chefilip, line 44: missing deps here?

chef-jojo:
I wonder can we just treat the `lockedAmount` as single source, and compute the percentage on the fly.

src/views/Pools/components/LockedPool/Common/BalanceField.tsx
ChefMomota: Update watch the `lockedAmount`

src/views/Pools/components/LockedPool/Common/BalanceField.tsx
```diff
@@ -36,19 +36,22 @@ const BalanceField: React.FC<PropsType> = ({
   const [percent, setPercent] = useState(0)
   const { userNotEnoughCake, notEnoughErrorMessage } = useUserEnoughCakeValidator(lockedAmount, stakingTokenBalance)
 
+  useEffect(() => {
+    const amount = new BigNumber(lockedAmount)
+    if (amount.gt(0)) {
+      const convertedInput = amount.multipliedBy(BIG_TEN.pow(stakingDecimals))
```
memoyil, line 42: BIG_TEN.pow(stakingDecimals) can be extracted to a state by using initial state as function. Therefore we will not calculate it each render

memoyil: Ignore my comment above, the correct approach should be a create a state with initalstate as function  and a useEffect that sets that state when stakingDecimals change.

chef-jojo (approved):

PR #3958, opened 2022-05-18
fix(farms): Auction loading and auction only tag

chef-jojo:


src/views/Farms/components/FarmTable/Row.tsx
```diff
@@ -151,11 +153,12 @@ const Row: React.FunctionComponent<RowPropsWithLoading> = (props) => {
           <FarmMobileCell colSpan={3}>
             <Flex justifyContent="space-between" alignItems="center">
               <Farm {...props.farm} />
-              {props.type === 'community' ? (
-                <FarmAuctionTag marginRight="16px" scale="sm" />
-              ) : (
-                <CoreTag marginRight="16px" scale="sm" />
-              )}
+              {farmAuctionLoaded &&
```
Chef-Yogi, line 156: I can add skeleton here while loading after you merge this

memoyil:
Currently fetchFarmAuction only sets isCommunity to true (not to false if not), that means after the auction period ends farm will not be visible as core instead of farmauction since you set the data statically. Is this intended or should we fix it?

chef-jojo:
> Currently fetchFarmAuction only sets isCommunity to true (not to false if not), that means after the auction period ends farm will not be visible as core instead of farmauction since you set the data statically. Is this intended or should we fix it?

@memoyil farms eligible with farm auction only should stay as farm auction even after ends

chefilip:
Screen Shot 2022-05-18 at 05 09 13

I think we need to `isCommunity` in this file as well @chef-jojo

chef-jojo:
> Screen Shot 2022-05-18 at 05 09 13
> 
> I think we need to `isCommunity` in this file as well @chef-jojo

yeah we need this, it's not removed

PR #3963, opened 2022-05-18
feat: Don't hide phishing warning banner when scrolling down

memoyil:
To reproduce: 

1. Go to website after cleaning cache
2. See phishing warning at the top
3. See it disappears (which it shouldn't) when user scrolls down

### DIFF
--- a/packages/uikit/src/widgets/Menu/Menu.tsx
+++ b/packages/uikit/src/widgets/Menu/Menu.tsx
@@ -20,12 +20,14 @@ const Wrapper = styled.div`
   width: 100%;
 `;
 
-const StyledNav = styled.nav`
+const StyledNav = styled.nav<{ showMenu: boolean }>`
   display: flex;
   justify-content: space-between;
   align-items: center;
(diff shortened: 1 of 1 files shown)


chefilip:
CleanShot 2022-05-20 at 15 52 37

I think the current behavior is already visible enough for users.

memoyil:
> CleanShot 2022-05-20 at 15 52 37 [ ](https://user-images.githubusercontent.com/97418926/169593745-0c1bb59b-2f58-4074-b54b-f3eed9c44048.gif)
> 
> I think the current behavior is already visible enough for users.

Yes but when user scrolls down it disappears without clicking close button, btw which browser you see this flickering issue

memoyil:
@chefilip Safari transition fixed

chef-jojo:
I don't think it's necessary to keep the banner appearing on top.

memoyil:
> I don't think it's necessary to keep the banner appearing on top.

I will close the pr then

PR #3992, opened 2022-05-21
perf: Run lists updater only when the page has currency search

memoyil:
This will save 2-3 mb of data when visiting other pages that does not contain currency search thus token list is not needed

### DIFF
--- a/src/hooks/useInterval.ts
+++ b/src/hooks/useInterval.ts
@@ -1,5 +1,6 @@
 import { useEffect, useRef, useState, useCallback } from 'react'
 import isUndefinedOrNull from 'utils/isUndefinedOrNull'
+import useLastUpdated from './useLastUpdated'
 
 export default function useInterval(
   callback: () => void,
@@ -7,8 +8,17 @@ export default function useInterval(
   leading = true,
   initiateUpdate = true,
 ) {
+  const [runImmediate, setRunImmediate] = useState(leading)
+  const [runAfter, setRunAfter] = useState(delay)
   const savedCallback = useRef<() => void>(callback)
   const [isReadyForUpdate, setIsReadyForUpdate] = useState(false)
+  const { lastUpdated, setLastUpdated: refresh } = useLastUpdated()
+
+  // Use props indirectly to not render set up timeout effect twice if both props changed.
+  useEffect(() => {
+    setRunImmediate(leading)
+    setRunAfter(delay)
+  }, [leading, delay])
 
   const tick = useCallback(() => {
     setIsReadyForUpdate(true)
@@ -23,19 +33,23 @@ export default function useInterval(
     if (initiateUpdate && isReadyForUpdate) {
       savedCallback.current?.()
       setIsReadyForUpdate(false)
+      setRunImmediate(false)
+      refresh()
     }
-  }, [initiateUpdate, isReadyForUpdate])
+  }, [initiateUpdate, isReadyForUpdate, refresh])
 
-  // Set up the interval.
+  // Set up the timeout.
   useEffect(() => {
-    if (!isUndefinedOrNull(delay)) {
-      if (leading) tick()
-      const id = setInterval(tick, delay)
-      return () => {
-        setIsReadyForUpdate(false)
-        clearInterval(id)
+    if (!isUndefinedOrNull(runAfter)) {
+      if (runImmediate) tick()
+      else {
+        const id = setTimeout(tick, runAfter)
+        return () => {
+          setIsReadyForUpdate(false)
+          clearTimeout(id)
+        }
       }
     }
     return () => setIsReadyForUpdate(false)
-  }, [delay, leading, tick])
+  }, [runAfter, runImmediate, lastUpdated, tick])
 }
--- a/src/hooks/useInterval.ts
+++ b/src/hooks/useInterval.ts
@@ -1,4 +1,4 @@
-import { useEffect, useRef, useState } from 'react'
+import { useEffect, useRef, useState, useCallback } from 'react'
 import isUndefinedOrNull from 'utils/isUndefinedOrNull'
 
 export default function useInterval(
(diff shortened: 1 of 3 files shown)


src/index.tsx
```diff
@@ -6,9 +7,13 @@ import MulticallUpdater from './state/multicall/updater'
 import TransactionUpdater from './state/transactions/updater'
 
 export function Updaters() {
+  const router = useRouter()
+  const includeListUpdater = ['/swap', '/limit-orders', '/add', '/find', '/remove'].some((item) => {
```
chefilip, line 11: when user go to `swap`, go to `farm` and then go back to `swap`, the whole thing in `ListsUpdater` will be executed again, right?

memoyil: You are right, I need to change the logic to not make request every time page with currency search visited

memoyil: Done 👍

src/hooks/useInterval.ts
```diff
@@ -1,27 +1,55 @@
...
+import isUndefinedOrNull from 'utils/isUndefinedOrNull'
+import useLastUpdated from './useLastUpdated'
 
-export default function useInterval(callback: () => void, delay: null | number, leading = true) {
-  const savedCallback = useRef<() => void>()
+export default function useInterval(
+  callback: () => void,
+  delay: undefined | null | number,
+  leading = true,
+  initiateUpdate = true,
```
chef-jojo, line 9: what's the `initiateUpdate` here for?

memoyil: This is for to run interval/timeout on the background but not calling callback function (which is list updater in our example) when time has come but user is not in swap / liquidity pages. Reason for doing that to avoid calling list updater everytime user access those pages.

src/state/lists/updater.ts
```diff
@@ -27,8 +34,8 @@ export default function Updater(): null {
     )
   }, [fetchList, isWindowVisible, lists])
 
-  // fetch all lists every 10 minutes, but only after we initialize library
-  useInterval(fetchAllListsCallback, library ? 1000 * 60 * 10 : null)
+  // fetch all lists every 10 minutes, but only after we initialize library and page has currency input
+  useInterval(fetchAllListsCallback, library ? 1000 * 60 * 10 : null, true, includeListUpdater)
```
chef-jojo, line 38: can we simplify this as
```js
useInterval(fetchAllListsCallback, library && includeListUpdater ? 1000 * 60 * 10 : null, includeListUpdater)
```

memoyil: No, we need to run interval on the background even includeListUpdater is false to not run every time user visits swap pages.

chefilip: Could we do this?

```
useInterval(() => includeListUpdater && fetchAllListsCallback(), ...)
```

chefilip:
```
const fetchAllListsCallback = useCallback(() => {
  if (!isWindowVisible || includeListUpdater) return
  ....
})
```

memoyil: Done

src/hooks/useInterval.ts
```diff
@@ -1,27 +1,50 @@
-import { useEffect, useRef } from 'react'
+import { useEffect, useRef, useState, useCallback } from 'react'
+import isUndefinedOrNull from 'utils/isUndefinedOrNull'
+import useLastUpdated from './useLastUpdated'
 
-export default function useInterval(callback: () => void, delay: null | number, leading = true) {
-  const savedCallback = useRef<() => void>()
+export default function useInterval(callback: () => void, delay: undefined | null | number, leading = true) {
```
chefilip, line 5: for readability, I suggest the option API 
```
useInterval(callback, options)
const { delay, leading } = options

```

src/hooks/useInterval.ts
```diff
@@ -1,27 +1,50 @@
...
+  const [runAfter, setRunAfter] = useState(delay)
+  const savedCallback = useRef<() => void>(callback)
+  const [isReadyForUpdate, setIsReadyForUpdate] = useState(false)
+  const { lastUpdated, setLastUpdated: refresh } = useLastUpdated()
+
+  // Use props indirectly to not render set up timeout effect twice if both props changed.
+  useEffect(() => {
+    setRunImmediate(leading)
+    setRunAfter(delay)
+  }, [leading, delay])
```
chefilip, line 16: Screen Shot 2022-07-01 at 08 45 04


Could you check if this pattern is similar to this one? 

Reference: https://beta-reactjs-org-git-you-might-not-fbopensource.vercel.app/learn/you-might-not-need-an-effect

memoyil: In our case delay and leading props not used directly, so we can use it in effect without side effects

src/hooks/useInterval.ts
```diff
@@ -1,27 +1,50 @@
...
-    if (delay !== null) {
-      if (leading) tick()
-      const id = setInterval(tick, delay)
-      return () => clearInterval(id)
+  // Set up the timeout.
+  useEffect(() => {
+    if (!isUndefinedOrNull(runAfter)) {
+      if (runImmediate) tick()
+      else {
+        const id = setTimeout(tick, runAfter)
```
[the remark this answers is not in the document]
memoyil: To reduce calls by setting new timeout after calling a callback otherwise there will be more calls if user enters the exchange pages which triggers callback immediately and interval kicking in very close results unnecessary requests.

memoyil:
@chefilip I had to revert my last commit, we have to use initiateUpdate prop separately to trigger callback immediately when user enters relevant pages

src/hooks/useInterval.ts
```diff
@@ -1,27 +1,55 @@
-import { useEffect, useRef } from 'react'
+import { useEffect, useRef, useState, useCallback } from 'react'
+import isUndefinedOrNull from 'utils/isUndefinedOrNull'
+import useLastUpdated from './useLastUpdated'
 
-export default function useInterval(callback: () => void, delay: null | number, leading = true) {
-  const savedCallback = useRef<() => void>()
+export default function useInterval(
```
chefilip, line 5: Could you help me double check other components using `useInterval`?

We should keep the same behaviour for other places. To be safe, I think we could create another hook with better name.

memoyil: The only other place that useInterval used is in locked pools BurningCountdown. I tested it, it is working fine 👍

chefilip (approved):

PR #4025, opened 2022-05-27
fix: avoid flash of content

bertho-zero:
When I enter the address of a token that is part of a list that I have not yet activated I see this for a few ms:

and that right after:

The first capture should not display


memoyil:
> Bad fix

Check this one https://github.com/pancakeswap/pancake-frontend/pull/4041

PR #4027, opened 2022-05-28
chore(limit-order): Add more order in format

chefilip:


src/views/LimitOrders/hooks/useFormattedOrderData.ts
```diff
@@ -34,7 +34,7 @@ const formatForDisplay = (amount: CurrencyAmount | Price) => {
   }
   return parseFloat(amount.toSignificant(18)).toLocaleString(undefined, {
     minimumFractionDigits: 0,
-    maximumFractionDigits: 8,
```
chef-jojo, line 37: is amount.toExact() better?

PR #4032, opened 2022-05-30
feat: Zap

chef-jojo:


src/config/constants/index.ts
```diff
@@ -72,6 +72,8 @@ export const DEFAULT_DEADLINE_FROM_NOW = 60 * 20
 
 export const BIG_INT_ZERO = JSBI.BigInt(0)
 
+export const LIQUIDLY_MINIMUM_AMOUNT = JSBI.BigInt(1000)
```
memoyil, line 75: typo

MirekBrzez (dismissed):

MirekBrzez:
Is everything Good, I'll be back with new phone.
Sorry

memoyil:
It crashes when BNB to Bunny selected

memoyil:
Imo if user has already lp's and clicks 'Add liquidity instead' button, it should bypass the choose valid pair page and directly points to the add liquidity page. wdyt @chef-jojo ?

chef-jojo:
> Imo if user has already lp's and clicks 'Add liquidity instead' button, it should bypass the choose valid pair page and directly points to the add liquidity page. wdyt @chef-jojo ?

Added querystring to bypass choose pair 💪

memoyil:
> > Imo if user has already lp's and clicks 'Add liquidity instead' button, it should bypass the choose valid pair page and directly points to the add liquidity page. wdyt @chef-jojo ?
> 
> Added querystring to bypass choose pair 💪

I think it should be vice versa, if user accesses via direct link or clicks get lp links in farms, it should be pointed to the add liquidity page. Imho valid pair page only needed when user wants to add liquidity from scratch in liquidity page

PR #4036, opened 2022-05-30
perf: (auction-voting): Make visible prefetch false for links

memoyil:
Imo those links should not be prefetched (i think they are not frequently visited) when they are visible anyways it will still be prefetched when they are hovered. @chef-jojo Maybe all next links visible prefetch should be false wdyt?

### DIFF
--- a/src/views/Farms/Farms.tsx
+++ b/src/views/Farms/Farms.tsx
@@ -361,7 +361,7 @@ const Farms: React.FC = ({ children }) => {
         <Heading scale="lg" color="text">
           {t('Stake LP tokens to earn.')}
         </Heading>
-        <NextLinkFromReactRouter to="/farms/auction" id="lottery-pot-banner">
+        <NextLinkFromReactRouter to="/farms/auction" prefetch={false}>
           <Button p="0" variant="text">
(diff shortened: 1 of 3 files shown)


chef-jojo (approved):

chef-jojo:
Good idea on all prefetch false.
Let's have another pr to test it. I feel like now mobile navigation is relatively fast because of it though

memoyil:
> Good idea on all prefetch false. Let's have another pr to test it. I feel like now mobile navigation is relatively fast because of it though

This I agree, maybe we should keep prefetch true for mobile and make it false for desktop?

chef-jojo:
> > Good idea on all prefetch false. Let's have another pr to test it. I feel like now mobile navigation is relatively fast because of it though
> 
> 
> 
> This I agree, maybe we should keep prefetch true for mobile and make it false for desktop?


Not sure how next link prefetch works on mobile, maybe need some deep dive into the code.
I would suggest avoid using useMatchBreakpoint or at least refactor it to not adding new event listeners on each link instance

chef-jojo (approved):

PR #4037, opened 2022-05-31
perf: Make next link prefect default to false

memoyil:
Test it especially on mobile for checking degradation on responsiveness because even prefetch is false it will still be enabled in desktop when the link is hovered on see https://github.com/vercel/next.js/blob/15354e9b42201df8c2102f2ba2f3cfcca1e561c0/docs/api-reference/next/link.md?plain=1#L60


chef-jojo:
feel a bit slower on mobile

PR #4086, opened 2022-06-15
perf: Memoize footer links

memoyil:


src/components/Menu/index.tsx
```diff
@@ -46,7 +46,7 @@ const Menu = (props) => {
       cakePriceUsd={cakePriceUsd.toNumber()}
       links={menuItems}
       subLinks={activeMenuItem?.hideSubNav || activeSubMenuItem?.hideSubNav ? [] : activeMenuItem?.items}
-      footerLinks={footerLinks(t)}
+      footerLinks={footerLinks(t, currentLanguage.locale)}
```
chef-jojo, line 49: useMemo should be good enough

memoyil: Done

PR #4109, opened 2022-06-20
fix: (pool): Extend duration check based on current duration causes locked vault to not extend

memoyil:
Extend duration check currently being done based on current duration (instead of left current duration) which causes issue when extending.

To reproduce

1. Go to locked pool
2. Click extend 
3. Put value that should not extend 52 weeks (if you have 10 weeks of locked pool and only 2 weeks left to put 50 weeks should not give an error)
4. See error and confirm button is disabled

### DIFF
--- a/src/views/Pools/components/LockedPool/Modals/ExtendDurationModal.tsx
+++ b/src/views/Pools/components/LockedPool/Modals/ExtendDurationModal.tsx
@@ -57,9 +57,12 @@ const ExtendDurationModal: React.FC<ExtendDurationModal> = ({
   const prepConfirmArg = useCallback(
     ({ duration }) => ({
       finalDuration: duration,
-      finalLockedAmount: extendLockedPosition ? getBalanceAmount(MIN_LOCK_AMOUNT, stakingToken.decimals).toNumber() : 0,
+      finalLockedAmount:
+        extendLockedPosition && currentDuration + duration > MAX_LOCK_DURATION
+          ? getBalanceAmount(MIN_LOCK_AMOUNT, stakingToken.decimals).toNumber()
+          : 0,
     }),
-    [stakingToken.decimals, extendLockedPosition],
+    [stakingToken.decimals, extendLockedPosition, currentDuration],
   )
 
   const customOverview = useCallback(
--- a/src/views/Pools/components/LockedPool/Modals/ExtendDurationModal.tsx
+++ b/src/views/Pools/components/LockedPool/Modals/ExtendDurationModal.tsx
@@ -7,11 +7,16 @@ import { MAX_LOCK_DURATION } from 'config/constants/pools'
 import { useTranslation } from 'contexts/Localization'
 import BigNumber from 'bignumber.js'
 import { useIfoCeiling } from 'state/pools/hooks'
+
+import { getBalanceAmount } from 'utils/formatBalance'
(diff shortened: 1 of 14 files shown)


src/views/Pools/components/LockedPool/Modals/ExtendDurationModal.tsx
```diff
@@ -39,10 +47,16 @@ const ExtendDurationModal: React.FC<ExtendDurationModal> = ({
...
     },
-    [currentLockedAmount, currentDuration],
+    [currentLockedAmount, lockEndTime],
   )
 
-  const prepConfirmArg = useCallback(({ duration }) => ({ finalDuration: duration, finalLockedAmount: 0 }), [])
+  const prepConfirmArg = useCallback(
+    ({ duration }) => ({
+      finalDuration: duration,
+      finalLockedAmount: currentBalance ? getBalanceAmount(MIN_AMOUNT, stakingToken.decimals).toNumber() : 0,
```
chefilip, line 56: should it be `currentBalance.gt(MIN_AMOUNT) ? ... : ...`  ?

src/views/Pools/components/LockedPool/Modals/ExtendDurationModal.tsx
```diff
@@ -77,6 +91,7 @@ const ExtendDurationModal: React.FC<ExtendDurationModal> = ({
         </Box>
         <LockedBodyModal
           stakingToken={stakingToken}
+          hasEnoughBalanceToExtend={currentBalance?.gt(MIN_AMOUNT)}
           onDismiss={onDismiss}
           lockedAmount={new BigNumber(currentLockedAmount)}
```
chefilip, line 96: could you help me add `useMemo` for this prop to avoid re-render?

chefilip:
Good idea ! Love it.

src/views/Pools/components/LockedPool/Modals/ExtendDurationModal.tsx
memoyil: Done

src/views/Pools/components/LockedPool/Modals/ExtendDurationModal.tsx
memoyil: Done

chefilip (dismissed):

src/views/Pools/components/LockedPool/Modals/ExtendDurationModal.tsx
```diff
@@ -40,10 +48,18 @@ const ExtendDurationModal: React.FC<ExtendDurationModal> = ({
...
     },
-    [currentLockedAmount, currentDuration],
+    [currentLockedAmount, lockEndTime],
   )
 
-  const prepConfirmArg = useCallback(({ duration }) => ({ finalDuration: duration, finalLockedAmount: 0 }), [])
+  const prepConfirmArg = useCallback(
+    ({ duration }) => ({
+      finalDuration: duration,
+      finalLockedAmount: checkEnoughBalanceToExtend
```
chefilip, line 57: My concern is that we quietly block users' CAKE. Even though the amount is small, I think users should be aware of this before taking the action. At the end, it's still their CAKE, their decision.

chefilip (approved):

PR #4123, opened 2022-06-23
Add PancakeRouter into contracts.ts

AndyJiangPro:
The code is self-explainary. Simply add missing pancake-related smart contract addresses into constants.


chefilip:
The `contracts.ts` is served within the FE scope only. It's not a document for all contracts. When FE uses it, we will add those contracts in. Therefore, I will close this PR for now.

Thanks for your contribution.

PR #4133, opened 2022-06-24
fix: Menu language

ChefMomota:


src/components/Menu/hooks/useMenuItems.ts
```diff
@@ -11,9 +11,7 @@ export const useMenuItems = (): ConfigMenuItemsType[] => {
   } = useTranslation()
   const menuItemsStatus = useMenuItemsStatus()
 
-  const menuItems = useMemo(() => {
-    return config(t, languageCode)
-  }, [t, languageCode])
+  const menuItems = config(t, languageCode)
```
chef-jojo, line 14: can you try to move this line under the useMemo below?

ChefMomota: Will have the same issue with the current one.

chef-jojo: so weird... the menuItems will not be memo at current, i'll handle it later

chef-jojo (approved):

src/components/Menu/hooks/useMenuItems.ts
memoyil: What was the issue ?

chef-jojo: @memoyil 
first load with other lang doesn't work on menu items. maybe we need to use lodash memoize
yet it's weird that the useMemo doesn't trigger when locale changed

memoyil: I think the root cause of the issue is related to translation function. Can you review this ? https://github.com/pancakeswap/pancake-frontend/pull/4135

PR #4135, opened 2022-06-24
fix: Refresh translation function on first load when other language selected

memoyil:


src/contexts/Localization/Provider.tsx
```diff
@@ -53,7 +56,7 @@ export const LanguageProvider: React.FC = ({ children }) => {
     }
 
     fetchInitialLocales()
-  }, [setState])
+  }, [setState, refresh])
```
chef-jojo, line 59: can remove `setState`

memoyil: Done

chef-jojo (approved):

PR #4201, opened 2022-07-05
feat: Perpetual link add light or dark mode

ChefMomota:


src/hooks/useTheme.ts
```diff
@@ -1,11 +1,26 @@
...
 import { ThemeContext as StyledThemeContext } from 'styled-components'
 import { useTheme as useNextTheme } from 'next-themes'
 
+const COOKIE_THEME_KEY = 'theme'
+
 const useTheme = () => {
   const { resolvedTheme, setTheme } = useNextTheme()
   const theme = useContext(StyledThemeContext)
-  return { isDark: resolvedTheme === 'dark', theme, setTheme }
+  const getThemeCookie = Cookie.get(COOKIE_THEME_KEY)
```
memoyil, line 11: why do we need cookie? afaik we have theme property in local storage

ChefMomota: Because when click perpetual button, the theme will not same as our side.

memoyil: 👍

src/hooks/useTheme.ts
```diff
@@ -1,11 +1,26 @@
...
+
+  if (!getThemeCookie) {
+    const themeValue = theme.isDark ? 'dark' : 'light'
+    Cookie.set(COOKIE_THEME_KEY, themeValue)
+  }
+
+  const handleSwitchTheme = (themeValue: 'light' | 'dark') => {
+    setTheme(themeValue)
+    Cookie.set(COOKIE_THEME_KEY, themeValue)
```
chefilip, line 20: Could you add error handle here in case cookie is full?

ChefMomota: Done

chefilip (approved):

PR #4209, opened 2022-07-05
fix: Bet results history displayed decimal amount and make it configurable

memoyil:
Bet result history shows 3 digit decimals instead of 4


chefilip:
Hmm. I think we could centralize decimal format logic of the `useConfig`  in a UI component or a hook. The current approach is hard to maintain.

chef-jojo (approved):

PR #4219, opened 2022-07-06
refactor: Fix potential null profile error and use memo

memoyil:


src/views/Pools/hooks/useProfileRequirement.ts
```diff
@@ -1,14 +1,21 @@
 import { useProfile } from 'state/profile/hooks'
 import { DeserializedPool } from 'state/types'
+import { useMemo } from 'react'
 
 export function useProfileRequirement(profileRequirement: DeserializedPool['profileRequirement']) {
-  const { profile, hasProfile } = useProfile()
+  const { profile, hasActiveProfile } = useProfile()
 
-  const notMeetRequired = profileRequirement && profileRequirement.required && (!hasProfile || !profile.isActive)
```
memoyil, line 7: Although it is not possible now, if hasProfile is true and profile is null, it will crash. It is better to use hasActiveProfile property which does this check internally

chefilip (approved):

PR #4232, opened 2022-07-07
fix: (prediction): Open round card shows wrong entered position after refresh

memoyil:
To reproduce:

1. Go to prediction
2. Bet down
3. Refresh the page
4. See it displays 'Up entered'

### DIFF
--- a/src/views/Predictions/components/RoundCard/OpenRoundCard.tsx
+++ b/src/views/Predictions/components/RoundCard/OpenRoundCard.tsx
@@ -175,7 +175,7 @@ const OpenRoundCard: React.FC<OpenRoundCardProps> = ({
                   {t('Enter DOWN')}
                 </Button>
               </>
-            ) : (
+            ) : enteredPosition ? (
               <>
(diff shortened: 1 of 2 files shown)


src/views/Predictions/components/RoundCard/OpenRoundCard.tsx
```diff
@@ -172,7 +179,7 @@ const OpenRoundCard: React.FC<OpenRoundCardProps> = ({
               <>
                 <div ref={targetRef}>
                   <Button disabled startIcon={getPositionEnteredIcon()} width="100%" mb="8px">
-                    {t('%position% Entered', { position: positionDisplay })}
+                    {t('%position% Entered', { position: enteredPosition })}
```
chefilip, line 182: For null value, what should we show here?

memoyil: Fixed

chefilip (dismissed):

PR #4233, opened 2022-07-07
fix: (prediction): Collect winnings button not displayed in bet results

memoyil:
To reproduce:

1. Go prediction history
2. Go to non claimed winning bets 
3. Expand the row
4. See colelct winnings button not displayed

### DIFF
--- a/src/views/Predictions/components/History/BetResult.tsx
+++ b/src/views/Predictions/components/History/BetResult.tsx
@@ -129,7 +129,7 @@ const BetResult: React.FC<BetResultProps> = ({ bet, result }) => {
         </Flex>
       </Flex>
       <StyledBetResult>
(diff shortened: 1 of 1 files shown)


src/views/Predictions/components/History/BetResult.tsx
```diff
@@ -128,7 +128,7 @@ const BetResult: React.FC<BetResultProps> = ({ bet, result }) => {
         </Flex>
       </Flex>
       <StyledBetResult>
-        {result === Result.WIN && !canClaim && (
+        {result === Result.WIN && canClaim && (
```
chefilip, line 131: CollecWinButton needs to pass canClaim. Are you sure this is a root cause?

chefilip (approved):

PR #4263, opened 2022-07-12
feat: Make disabling zap from prompt easier

ChefMomota:


src/views/AddLiquidity/index.tsx
```diff
@@ -84,6 +85,12 @@ export default function AddLiquidity() {
   const currencyA = useCurrency(currencyIdA)
   const currencyB = useCurrency(currencyIdB)
 
+  useEffect(() => {
+    if (zapMode !== temporarilyZapMode) {
+      setTemporarilyZapMode(zapMode)
+    }
+  }, [zapMode, temporarilyZapMode, setTemporarilyZapMode])
```
chef-jojo, line 92: when zapMode is true and temporarilyZapMode is false
it will set temporarilyZapMode back to true?

src/views/AddLiquidity/index.tsx
```diff
@@ -73,6 +73,7 @@ export default function AddLiquidity() {
   const tokens = serializeTokens()
 
   const [zapMode] = useZapModeManager()
+  const [temporarilyZapMode, setTemporarilyZapMode] = useState(zapMode)
```
chef-jojo, line 76: please avoid a useState default value is another state.
how about this
`const [temporarilyDisableZapMode, setTemporarilyDisableZapMode] = useState(false)`

to know if zapMode is on
check `!!zapModa || !temporarilyDisableZapMode`

chef-jojo (dismissed):

chef-jojo (approved):

PR #4270, opened 2022-07-13
feat: Prompt Zap is available

ChefMomota:


src/views/AddLiquidity/index.tsx
```diff
@@ -472,6 +474,13 @@ export default function AddLiquidity() {
     'zapInModal',
   )
 
+  const handleEnableZap = () => {
+    if (!zapMode) {
+      toggleZapMode(!zapMode)
+      toastInfo(t('Info'), t('Zap has been enabled temporarily.'))
```
chef-jojo, line 480: should have another state if it is temporarily enabled

chef-jojo (approved):

PR #4289, opened 2022-07-18
fix: Risk modal mobile bottom style

ChefMomota:
Prefer  `RiskDisclaimer` change using `DisclaimerModal `

Before:
Screenshot 2022-07-18 at 9 39 45 AM
Screenshot 2022-07-18 at 9 40 11 AM

After
Screenshot 2022-07-18 at 10 26 21 AM
Screenshot 2022-07-18 at 10 26 43 AM
:


cypress/integration/predictions/predictions.test.ts
```diff
@@ -1,12 +0,0 @@
...
-    cy.visit('/prediction')
-    cy.get('#predictions-risk-disclaimer').should('be.visible')
-    cy.get('#prediction-disclaimer-continue').should('be.disabled')
-    cy.get('#responsibility-checkbox').click()
-    cy.get('#beta-checkbox').click()
-    cy.get('#prediction-disclaimer-continue').should('not.be.disabled')
-    cy.get('#prediction-disclaimer-continue').click()
-    cy.get('#predictions-risk-disclaimer').should('not.exist')
-  })
-})
```
Chef-Yogi, line 12: is this test case not suited anymore?
why should we remove this

ChefMomota: Yup. Because we use `DisclaimerModal`, so no have so much or id can be check.

Chef-Yogi: we could remove the test case or we can pass the id to props and make the id dynamic, so we can still have the test case to protect.

what do you guys think @ChefMomota @chef-jojo @chefilip

ChefMomota: @Chef-Yogi  fixed

Chef-Yogi (approved):

PR #4292, opened 2022-07-18
feat: add wagmi package

chef-jojo:


src/utils/wallet.ts
```diff
@@ -99,3 +99,5 @@ export const registerToken = async (
 
   return tokenAdded
 }
+
+export const canRegisterToken = () => typeof window !== 'undefined' && window?.ethereum?.isMetaMask
```
chefilip, line 103: does it need to be a function?

chef-jojo: let's keep it to make sure it's re-evaluated on client side

PR #4337, opened 2022-07-26
perf(IFO): improve token vesting panel viable

woodywu810:
This panel should not be displayed when no tokens vesting, it will confusion users.


chefilip:
Thanks for your contribution. But I think for users who are new to IFO, it would be better to display it so they could learn more about IFO.

woodywu810:
Got it. 

Wouldn't it be better to put it here on mobile?
39C451CF-70C7-4391-BB49-8B174628230D

PR #4355, opened 2022-07-29
fix(Profile): remove subMenu

woodywu810:
Profile's subMenu is the same as NFT, but they are not directly related, so it is recommended to remove it.


src/components/Menu/utils.ts
```diff
@@ -12,6 +12,13 @@ export const getActiveSubMenuItem = ({ pathname, menuItem }: { pathname: string;
     return undefined
   }
 
+  // When Pathname includes "/nfts/profile", then hide subMenu.
+  if (pathname.includes('/nfts/profile')) {
```
chefilip, line 16: Good idea. But adding the `ntfs` logic here is not appropriate, the Menu component handles generic cases. Probably implement it in somewhere else ( might be in config file )

woodywu810: I tried to do it in Config, but the Profile is not included in the Menu list, maybe the Profile should appear in the subNav?

e.g. Overview, Collections, Activity, Profile

woodywu810: Move the Profile to the root directory, I think that solves all problems 👀

ChefMomota (approved):

chefilip (approved):

PR #4357, opened 2022-07-29
perf(ROI Calculator): add LP rewards desc.

woodywu810:
add LP rewards desc.


woodywu810:
> This is not a performance improvement imho, can you change the title of the pr?

I'm not sure it is feat or chore... Which one do you think you should choose?

woodywu810:
> This is not a performance improvement imho, can you change the title of the pr?

Thanks for reminding.

PR #4358, opened 2022-07-29
perf: Use matchbreakpoints context in uikit components if available

memoyil:
This reduces to matchbreakpoint hook to 1 when it is used in pcs frontend


packages/uikit/src/contexts/MatchBreakpoints/useMatchBreakpointsContext.ts
```diff
@@ -1,14 +1,13 @@
...
   const matchBreakpointContext = useContext(MatchBreakpointsContext);
+  const contextAvailable = matchBreakpointContext !== undefined;
+  const hookState = useMatchBreakpoints(contextAvailable);
 
-  if (matchBreakpointContext === undefined) {
-    throw new Error("Match Breakpoint context is undefined");
-  }
-
-  return matchBreakpointContext;
+  return contextAvailable ? matchBreakpointContext : hookState;
```
chefilip, line 10: Could I ask: in which case, the component doesn't have context?

chefilip: Hmm. Basically, we add a production code for storybook. I think we could remove the hookState and add context in storybook. It would be more consistent.

chef-jojo (changes requested):
Let's consolidate these two useMatchBreakpoints into one. 
We can try add our matchbreakpoint provider into globalDecorator if storybook needed

chef-jojo (approved):
Good work

PR #4381, opened 2022-08-02
feat(GlobalSettings): Setting button optimization

woodywu810:
1) The corresponding options appear in GlobalSettings according to the function.
2) Remove LimitOrder's setting button

image
image
image


ChefMomota:
1) Title font size 22px is too big. Maybe 18px is much better.
2)  same as @memoyil

woodywu810:
> I would keep that button, it gives user a quick access to change swap based settings especially in mobile devices

There is currently no option for LimitOrder, so placing the settings button will confuse users.So I prefer to remove the button.

ChefMomota:
Please fix eslint

ChefMomota (approved):

PR #4394, opened 2022-08-03
fix(SwapSettings): Add transationSpeed option

woodywu810:
TransationSpeed move to swapSettings

### DIFF
--- a/src/components/Menu/GlobalSettings/SettingsModal.tsx
+++ b/src/components/Menu/GlobalSettings/SettingsModal.tsx
@@ -103,7 +103,6 @@ const SettingsModal: React.FC<InjectedModalProps> = ({ onDismiss, mode }) => {
                   }}
                 />
               </Flex>
-              <GasSettings />
             </Flex>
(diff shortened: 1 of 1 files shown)


src/components/Menu/GlobalSettings/SettingsModal.tsx
```diff
@@ -103,7 +103,6 @@ const SettingsModal: React.FC<InjectedModalProps> = ({ onDismiss, mode }) => {
                   }}
                 />
               </Flex>
-              <GasSettings />
```
chef-jojo, line 106: Let's keep it in global

chef-jojo:
> Imho seeing all the settings in one place was more convenient to the user, currently if user wants to change a setting it has to go to the corresponding page first which can be annoying and there will be a maintenance problem if we want to introduce same setting in different pages

yeah agree. the icon is the same, so it's quite confusing at this moment

chef-jojo (approved):

PR #4403, opened 2022-08-03
perf: Use img onerror for fallback instead of picture tag to not load fallback always

memoyil:
Some browsers (safari, firefox i.e) still has a issue to load fallback even the original image is already available if picture tag is used, this causes to increase bandwidth.

image

It saves ~400kb data if page loaded without cache

image

image


chefilip:
https://nextjs.org/docs/api-reference/next/image

Is it possible to use `next/image` ?

chefilip (approved):

PR #4426, opened 2022-08-06
feat(Swap&LimitOrder): Add pin token bar

Woody810:


src/components/CurrencyInputPanel/index.tsx
```diff
@@ -86,6 +86,7 @@ interface CurrencyInputPanelProps {
   otherCurrency?: Currency | null
   id: string
   showCommonBases?: boolean
+  commonBasesType?: string
```
chef-jojo, line 89: can use enum, example:
```ts
enum CommonBasesType {
  SWAP_LIMIT = 'SWAP_LIMIT'
}
```

src/components/SearchModal/sorting.ts
```diff
@@ -49,4 +52,41 @@ function useTokenComparator(inverted: boolean): (tokenA: Token, tokenB: Token) =
...
+  // Tokens with balances will be placed first
+  tokens.forEach((token) => {
+    const balance = Number(balances[token.address]?.toSignificant(4))
+    if (Number.isNaN(balance) || balance === 0) {
+      return
+    }
+
+    result.push(token)
+
+    const tokenIndex = tokensTemp.findIndex((_token) => _token.symbol.toLowerCase() === token.symbol.toLowerCase())
```
chef-jojo, line 70: `token1.equal(token2)`

chef-jojo:
> Imho tbh this pr makes already complicated the token search & sort logic even more complicated, i would just add common tokens title&tooltip and rest can be done when the whole logic is refactored

agreed, I think we already have pinned token on top, we might not need to sort the pinned token, or we need to hide it when searching 🤔

CleanShot 2022-08-08 at 13 39 45@2x

chef-jojo (dismissed):

PR #4428, opened 2022-08-06
fix: Transaction description in recent trxs can be off if slippage is too high and add translation support

memoyil:
This creates confusion to the users where there is discrepancy between onchain data and pcs website

### DIFF
--- a/src/hooks/useSwapCallback.ts
+++ b/src/hooks/useSwapCallback.ts
@@ -1,6 +1,6 @@
 import { BigNumber } from '@ethersproject/bignumber'
 import { Contract } from '@ethersproject/contracts'
-import { SwapParameters, Trade } from '@pancakeswap/sdk'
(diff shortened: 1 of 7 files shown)


src/hooks/useSwapCallback.ts
```diff
@@ -128,10 +129,21 @@ export function useSwapCallback(
...
+            const inputAmount =
+              trade.tradeType === TradeType.EXACT_INPUT
+                ? trade.inputAmount.toSignificant(3)
+                : trade.maximumAmountIn(pct).toSignificant(3)
+            const outputAmount =
+              trade.tradeType === TradeType.EXACT_OUTPUT
+                ? trade.outputAmount.toSignificant(3)
+                : trade.minimumAmountOut(pct).toSignificant(3)
+
+            const base = `Swap ${
```
chef-jojo, line 142: Let's translate it

chef-jojo (approved):

PR #4454, opened 2022-08-10
perf: Reduce gas estimate calls to on chain in remove liquidity

memoyil:
Instead of making calls to all the possible methods, just return safe gas estimate when the we have it. Current behavior is similar to this but it chooses the first safe gas estimate after making calls for all the methods to the on-chain


chef-jojo:
How about promise.any?

chef-jojo (approved):

PR #4473, opened 2022-08-12
perf: (nft): Remove nftstate from redux

memoyil:
Filtering nfts that is stored in redux was a leftover from old implementation, it currently not used in anywhere. When any filter selected it requests to nft api and graphql unnecessarily in async thunk which is already done in swr hooks in useCollectionNfts hook.

To review: 

1. Go to nft items of any collection
2. See filtering by traits working properly


chef-jojo:
@memoyil let's move all nft out of redux. We can use session storage, or jotai to do the simple global state management

chef-jojo (approved):

PR #4481, opened 2022-08-15
fix(IFO): Fix token vesting ui

Woody810:


src/views/Ifos/components/IfoVesting/index.tsx
```diff
@@ -24,7 +24,6 @@ const StyleVestingCard = styled(Card)`
 
 const VestingCardBody = styled(CardBody)`
   overflow-y: auto;
-  max-height: 570px;
```
ChefMomota, line 27: why delete this?

PR #4500, opened 2022-08-16
fix: Link inside Booster tooltip cannot be clicked in mobile

memoyil:
To reproduce: 

1. Go to mobile
2. Go to boosted farm
3. Tap yield booster tooltip
4. Link cannot be tapped because it disappears so quickly


Chef-Yogi (dismissed):

Chef-Yogi:
Shall we apply this for mobile only?
Wdyt 🙏

Chef-Yogi (approved):

PR #4511, opened 2022-08-17
fix: Make Boosted Apr calculation based on cake rewards apr

memoyil:
I am not sure on this but boosted apr should be calculated based on cake rewards apr not the total apr which includes lp rewards apr.


Chef-Yogi:
let me check this with PM

Chef-Yogi (approved):

PR #4533, opened 2022-08-19
fix: Fix Show Trade BottomNav's DropdownMenu on Mobile

stvegerrad:
All menu items have submenu and all available on mobile except **Trade** menu Item.
I think we will consistency on the project and that's why I send you pull request.
image


chef-jojo:
Hey thanks for the PR.
I think it's intentional on mobile

stvegerrad:
Yes, but I think it should be changed to be consistent between mobile menu items.

PR #4538, opened 2022-08-21
feat(Pottery): Remove duplicate subgraphHealth buttom

Woody810:


chef-jojo:
It's by design we keep both? @ChefMomota

ChefMomota:
Should keep both

PR #4566, opened 2022-08-24
fix: modal context initial value

izayl:
## Steps To Reproduce

1. use latest version of `imToken` app.
2. open browser with https://pancakeswap.finance
3. click `trade` tab
4. click `token` 

**expected**

the token list modal opened

**now**

the token list not shown

https://user-images.githubusercontent.com/10740043/186350940-1361022b-f4bc-43dd-bf6e-e304f36a7c19.mov

## Problem

the modal context will initialize the `--vh` css variable, but the handle function under `useEffect` scope which will not triggered after repaint. So the result of `window.innerHeight` will be `zero` or other some value in some device as the figure shows below.

image

## Resolve

for DOM measurement reason, should use `useLayoutEffect` to replace `useEffect`.

### DIFF
--- a/packages/uikit/src/widgets/Modal/ModalContext.tsx
+++ b/packages/uikit/src/widgets/Modal/ModalContext.tsx
@@ -1,5 +1,5 @@
 import { AnimatePresence, domMax, LazyMotion, m } from "framer-motion";
-import React, { createContext, useEffect, useRef, useState } from "react";
+import React, { createContext, useEffect, useLayoutEffect, useRef, useState } from "react";
 import styled from "styled-components";
 import { mountAnimation, unmountAnimation } from "../../components/BottomDrawer/styles";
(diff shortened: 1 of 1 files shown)


chef-jojo (approved):
Nice finding! Yeah this should be layout effect

izayl:
@chef-jojo  this solution not resolve the problem, what do you think about it? Maybe we should assign a default value to the variable. If you agreed, I can create a new PR to handle it.

chef-jojo:
> @chef-jojo this solution not resolve the problem, what do you think about it? Maybe we should assign a default value to the variable. If you agreed, I can create a new PR to handle it.

Hi @izayl, sorry message got buried.

I'm not sure what's the root case there, it seems the `innerHeight` is `8` at that moment, resulting the `0.08vh`
Yeah feel free to create another PR to try, we have preview link on each PR to let you test it

PR #4603, opened 2022-08-28
fix: Sort field not updated when using table sort function

### LINKED ISSUE #4604: [BUG] Sort field not updated when using table sort function

### Product

NFT

### Current Behavior

No updating on "SORT BY" field.
See details #4603 

### Expected Behavior

"SORT BY" field should be changed.

### Steps To Reproduce

1. Go to [Collections ](https://pancakeswap.finance/nfts/collections)Page
2. Sort table by selecting the table header

### Environment

```markdown
- Device:
- OS:
- Browser:
- Wallet:
```

### Anything else?

_No response_

stvegerrad:
image
Table was sorted by Volume but "SORT BY" field still shows Collection.
Fixes #4604


src/components/Select/Select.tsx
```diff
@@ -141,6 +141,10 @@ const Select: React.FunctionComponent<React.PropsWithChildren<SelectProps>> = ({
     }
   }, [])
 
+  useEffect(() => {
+    setSelectedOptionIndex(defaultOptionIndex)
+  }, [defaultOptionIndex])
+
```
chef-jojo, line 147: not a fan of using `defaultOptionIndex` as `value`. but it's a flawed design on this Select component.
I will let this pass this time

chef-jojo:
LGTM, let's add the changes as @memoyil suggested

stvegerrad:
> LGTM, let's add the changes as @memoyil suggested

Okay, I'll do it

stvegerrad:
@memoyil @chef-jojo I just made changes, please check them and let me know

src/views/Nft/market/Collections/index.tsx
```diff
@@ -79,6 +79,44 @@ const Collectible = () => {
...
+      {
+        label: t('Lowest Price'),
+        value: SORT_FIELD.lowestPrice,
+      },
+      {
+        label: t('Highest Price'),
+        value: SORT_FIELD.highestPrice,
+      },
+    ]
+  }, []);
```
chef-jojo, line 109: Missing deps

stvegerrad: I just added it, please check again

src/components/Select/Select.tsx
```diff
@@ -141,6 +141,13 @@ const Select: React.FunctionComponent<React.PropsWithChildren<SelectProps>> = ({
     }
   }, [])
 
+  useEffect(() => {
+    if(defaultOptionIndex) {
+      setSelectedOptionIndex(defaultOptionIndex - 1)
```
chef-jojo, line 146: start index by 0

src/components/Select/Select.tsx
```diff
@@ -141,6 +141,13 @@ const Select: React.FunctionComponent<React.PropsWithChildren<SelectProps>> = ({
     }
   }, [])
 
+  useEffect(() => {
+    if(defaultOptionIndex) {
```
chef-jojo, line 145:
```suggestion
    if (defaultOptionIndex) {
```

do you have prettier or any editor extension installed?
recommended to have a proper setup for formatting

src/components/Select/Select.tsx
stvegerrad: If the index starts by 0, we can't detect first changes when the user sort the table using table header.
That's why I starts the index by 1.

stvegerrad:
```
<Text color={!optionSelected && placeHolderText ? 'text' : undefined}>
  {!optionSelected && placeHolderText ? placeHolderText : options[selectedOptionIndex].label}
</Text>
```
As you can see here, **select** component only shows the current sort field when optionSelected is false.
In case the user sort the table using table header, that value doesn't change.
I solved this problem by starting the index from 1.

src/components/Select/Select.tsx
stvegerrad: fixed

stvegerrad:
I think all issues are solved
Do you think so? @chef-jojo @memoyil

chef-jojo (approved):
Should refactor the select into accept value prop later.
Other than that should be good to go
Thanks for the pr

PR #4624, opened 2022-08-31
fix: (prediction): Markascollected action makes epoch claimable again

memoyil:
To reproduce: 

1. Go to prediction
2. Click collect winnings
3. After the collection collect winnings appears again until the next fetch


chefilip:
Do you know why it marks `true` in the first place?

PR #4636, opened 2022-09-01
perf: (info): Eliminate intermediate arrays when getting ids

memoyil:
To review

1. Go to info page
2. Search some pools for example using 'wbnb'
3. See pools listed properly on search component

### DIFF
--- a/src/state/info/queries/search/index.ts
+++ b/src/state/info/queries/search/index.ts
@@ -50,7 +50,10 @@ interface PoolSearchResponse {
 
 const getIds = (entityArrays: SingleQueryResponse[][]) => {
   const ids = entityArrays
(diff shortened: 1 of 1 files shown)


chefilip (dismissed):

src/state/info/queries/search/index.ts
```diff
@@ -50,7 +50,10 @@ interface PoolSearchResponse {
 
 const getIds = (entityArrays: SingleQueryResponse[][]) => {
   const ids = entityArrays
-    .reduce((entities, currentTokenArray) => [...entities, ...currentTokenArray], [])
+    .reduce((entities, currentTokenArray) => {
```
chefilip, line 53: can we use flatten in this case?

chefilip (approved):

PR #4640, opened 2022-09-01
perf: (lottery): Use promise all when fetching user tickets multiple rounds

memoyil:


src/state/lottery/getUserTicketsData.ts
```diff
@@ -58,14 +58,18 @@ export const fetchUserTicketsForMultipleRounds = async (
...
-  for (let i = 0; i < idsToCheck.length; i += 1) {
-    const roundId = idsToCheck[i]
-    // eslint-disable-next-line no-await-in-loop
-    const ticketsForRound = await fetchUserTicketsForOneRound(account, roundId)
+
+  const results = await Promise.all(
+    idsToCheck.map((roundId) => Promise.all([Promise.resolve(roundId), fetchUserTicketsForOneRound(account, roundId)])),
+  )
+
+  results.forEach((result) => {
```
chefilip, line 66: I think we could use `map` method here, right?

chefilip (dismissed):

chef-jojo (approved):

PR #4709, opened 2022-09-11
fix(farm): Display wrong usdc value

chefilip:


src/state/farms/selectors.ts
```diff
@@ -117,10 +117,10 @@ export const makeLpTokenPriceFromLpSymbolSelector = (lpSymbol: string) =>
       const lpTotalSupply = farm.lpTotalSupply ? new BigNumber(farm.lpTotalSupply) : BIG_ZERO
 
       if (lpTotalSupply.gt(0) && lpTotalInQuoteToken.gt(0)) {
-        const farmTokenPriceInUsd = new BigNumber(farm.tokenPriceBusd)
-        const tokenAmountTotal = farm.tokenAmountTotal ? new BigNumber(farm.tokenAmountTotal) : BIG_ZERO
+        const farmQuoteTokenPriceInUsd = new BigNumber(farm.quoteTokenPriceBusd)
+        const quoteTokenAmountTotal = farm.quoteTokenAmountTotal ? new BigNumber(farm.quoteTokenAmountTotal) : BIG_ZERO
         // Total value of base token in LP
-        const valueOfBaseTokenInFarm = farmTokenPriceInUsd.times(tokenAmountTotal)
+        const valueOfBaseTokenInFarm = farmQuoteTokenPriceInUsd.times(quoteTokenAmountTotal)
```
Woody810, line 123: What is the difference between the attributes with quote?

chefilip: Oh. I found out that there is miscalulating in tokenPrice. Will create another PR

PR #4711, opened 2022-09-11
perf: (farms): Don't fetch farm user data until proxy address fetched and dont fetch proxy farm data when proxy is not created

memoyil:
User data loaded twice when first access to the page

### DIFF
--- a/src/state/farms/hooks.ts
+++ b/src/state/farms/hooks.ts
@@ -34,7 +34,7 @@ export function useFarmsLength() {
 export const usePollFarmsWithUserData = () => {
   const dispatch = useAppDispatch()
   const { account, chainId } = useActiveWeb3React()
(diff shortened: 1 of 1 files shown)


src/state/farms/hooks.ts
```diff
@@ -54,7 +54,7 @@ export const usePollFarmsWithUserData = () => {
     : ['farmsWithUserData', account, chainId]
 
   useSWRImmutable(
-    account && chainId ? name : null,
+    account && chainId && !isProxyContractLoading ? name : null,
```
chefilip, line 57: For not boosted farms, does this logic still fetch `fetchFarmUserDataAsync`, right?

chefilip: Got it. Thanks for your contribution.

chefilip (approved):

PR #4723, opened 2022-09-12
perf: (token-search): Remove tag info from wrappedtokeninfo

memoyil:
It is not used anywhere and degrades the performance


chefilip:
I wonder what is the purpose of tags when developer created it? @memoyil

chefilip (approved):

chefilip (approved):

PR #4726, opened 2022-09-13
fix: PollFarmsWithUserData not fetch data when chainId is not BSC

ChefMomota:


ChefMomota:
> imo it is better to do this logic inside useBCakeProxyContractAddress wdyt? it will return 0x000... or null if the chain is not bsc

@memoyil  done

Chef-Yogi (approved):

Chef-Yogi:
SGTM

PR #4729, opened 2022-09-13
perf: (currency-search): Don't convert all tokens in lists to token instance when in currency search

memoyil:
It improves performance significantly when searching tokens in low-end mobile devices

### DIFF
--- a/apps/web/src/components/SearchModal/ImportToken.tsx
+++ b/apps/web/src/components/SearchModal/ImportToken.tsx
@@ -151,13 +151,24 @@ function ImportToken({ tokens, handleCurrencySelect }: ImportProps) {
           disabled={!confirmed}
           onClick={() => {
             tokens.forEach((token) => {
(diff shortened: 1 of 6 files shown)


apps/web/src/components/SearchModal/ImportToken.tsx
```diff
@@ -106,13 +116,24 @@ function ImportToken({ tokens, handleCurrencySelect }: ImportProps) {
...
+                const list = lists[url].current
+                tokenInList = list?.tokens.find(
+                  (tokenInfo) => tokenInfo.address === token.address && tokenInfo.chainId === token.chainId,
+                )
+                if (tokenInList) {
+                  break
+                }
+              }
+
+              const inactiveToken = chainId && tokenInList
```
chef-jojo, line 130: It's easy to have bug if someone forget to use this snippet.

chef-jojo: putting this into an atom getter might help

PR #4742, opened 2022-09-14
fix: useHover hook

memoyil:


chef-jojo:
What's wrong with current ref approach?

PR #4773, opened 2022-09-19
feat: IFO card title

ChefMomota:


src/views/Ifos/components/IfoFoldableCard/IfoPoolCard/index.tsx
```diff
@@ -134,14 +134,16 @@ const SmallCard: React.FC<React.PropsWithChildren<IfoCardProps>> = ({
     )
   }, [account, ifo, poolId, publicIfoData, vestingInformation, walletIfoData])
 
+  const cardTitle = ifo.isCommunity ? `${config.title} (cIFO)` : config.title
```
chefilip, line 137: Hmm. Currently there is a logic if point is required and squad isn't, it's cIFO. We should either choose the current one or replace with the flag here for other place to avoid multiple logic for the same state.

chef-jojo:
How about we just add a flag on ifo.ts?

chefilip:
> How about we just add a flag on ifo.ts?

By using SC attribute, FE doesn't need to maintain the type on FE side. Just delegate the IFO type maintenance to BD.

chefilip (approved):

PR #4782, opened 2022-09-19
feat: (swap): Add scanning risks for input token

memoyil:
It is easier for user to know risks of input token (especially for air-drops) without switching to output first

### DIFF
--- a/src/views/Swap/components/AccessRisk/index.tsx
+++ b/src/views/Swap/components/AccessRisk/index.tsx
@@ -1,13 +1,14 @@
-import { useState, useEffect, useMemo, useCallback } from 'react'
-import { Currency, ChainId } from '@pancakeswap/sdk'
+import { useState, useMemo, useCallback } from 'react'
+import { Currency } from '@pancakeswap/sdk'
 import { useTranslation } from '@pancakeswap/localization'
 import { Flex, Button, HelpIcon, useTooltip, Text, Link, useToast } from '@pancakeswap/uikit'
 import { fetchRiskToken, RiskTokenInfo } from 'views/Swap/hooks/fetchTokenRisk'
 import RiskMessage from 'views/Swap/components/AccessRisk/RiskMessage'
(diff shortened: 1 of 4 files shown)


src/views/Swap/components/AccessRisk/index.tsx
```diff
@@ -1,87 +1,208 @@
...
-  currency: Currency
+  inputCurrency: Currency
+  outputCurrency: Currency
 }
 
-const AccessRisk: React.FC<AccessRiskProps> = ({ currency }) => {
+const AccessRisk: React.FC<AccessRiskProps> = ({ inputCurrency, outputCurrency }) => {
   const { t } = useTranslation()
   const { toastInfo } = useToast()
+  const tokenMap = useAtomValue(tokenListFromOfficialsUrlsAtom)
```
chefilip, line 20: At this moment, Scanning risk only supports BSC mainnet. I think we don't have to include tokens of other networks

src/views/Swap/components/AccessRisk/index.tsx
```diff
@@ -1,87 +1,210 @@
...
-  const [riskTokenInfo, setRiskTokenInfo] = useState<RiskTokenInfo>({
-    isSuccess: false,
-    address: '',
-    chainId: ChainId.BSC,
-    riskLevel: TokenRiskPhases[0],
-    riskResult: '',
-    scannedTs: 0,
-    riskLevelDescription: '',
-  })
+  const [currentRiskTokensInfo, setCurrentRiskTokensInfo] = useState<{
```
chefilip, line 25: I think we could simplify the state structure here by using an array variable.
```
const [{ results, loading }, scanResults] = useState({ results: [], loading: false })

const tokensForScan = useMemo(() => { 
  1. check if input not in cache or not in token list --> add [].push(input)
  2. check if output not in cache --> add [].push(output)
}, [])

const results = await Promise.all(tokensForScan.map(scanningCall))

setResults(results)
```

chefilip (approved):

chefilip (approved):

PR #4834, opened 2022-09-26
feat: Support multichain in transaction updater

memoyil:


chef-jojo:
how about we just loop the updater with chains for simplicity 
`chains.map(chain => <TxUpdater chainid={chain.id} />)`

ChefMomota:
Tested. LGTM!

PR #4870, opened 2022-09-30
fix: (swap): Slippage error in confirm modal doesn't have settings and docs link in other languages

memoyil:
This pr separates useSwapCallback from translation function

Issue: 

image

image

### DIFF
--- a/apps/web/src/hooks/useSwapCallback.ts
+++ b/apps/web/src/hooks/useSwapCallback.ts
@@ -1,6 +1,5 @@
 import { BigNumber } from '@ethersproject/bignumber'
 import { Contract } from '@ethersproject/contracts'
-import { useTranslation } from '@pancakeswap/localization'
 import { SwapParameters, TradeType } from '@pancakeswap/sdk'
 import isZero from '@pancakeswap/utils/isZero'
 import truncateHash from '@pancakeswap/utils/truncateHash'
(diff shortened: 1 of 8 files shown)


src/views/Swap/components/SwapCommitButton.tsx
```diff
@@ -116,7 +117,7 @@ export default function SwapCommitButton({
         setSwapState({
           attemptingTxn: false,
           tradeToConfirm,
-          swapErrorMessage: error.message,
+          swapErrorMessage: getErrorReasonFromTransactionError(error),
```
chef-jojo, line 120: can you help check if UserRejectedRequestError
I think we shouldn't reset the whole swap state there

chef-jojo: Ah right. I mean we don't need to show transaction reject?
Just revert to no tx attempting

apps/web/src/hooks/useSwapCallback.ts
```diff
@@ -188,17 +184,14 @@ export function useSwapCallback(
             return response.hash
           })
           .catch((error: any) => {
-            // if the user rejected the tx, pass this along
-            if (error?.code === 4001) {
-              throw new Error('Transaction rejected.')
```
chefilip, line 193: should we keep this?

chefilip: Got it.

chefilip (dismissed):

apps/web/src/hooks/useSwapCallback.ts
chef-jojo: @memoyil Let's do the same thing for farms. I think it's still popup the error when user reject

PR #4872, opened 2022-10-01
feat(ToolTip): Add border color

Woody810:
image
image


ChefMomota:
For the design team.
The color for the tooltip is switched between bright/dark mode. No border.

PR #4874, opened 2022-10-01
perf: Farms test

ChefMomota:
Duplicates should only check in each chain


src/config/__tests__/farms.test.ts
```diff
@@ -12,14 +14,31 @@ const farmsToTest: [number, SerializedFarm, number][] = farms56
 
 const farms1ToTest: [number, SerializedFarm, number][] = farms1.slice(0, 10).map((farm) => [farm.pid, farm, 1])
 
+let logged = false
+export const getFarmConfig = async (chainId: ChainId) => {
```
chef-jojo, line 18: we don't need to lazy load in test

src/config/__tests__/farms.test.ts
```diff
@@ -12,14 +14,31 @@ const farmsToTest: [number, SerializedFarm, number][] = farms56
...
+    return []
+  }
+}
+
 describe('Config farms', () => {
-  it.each([...farmsToTest, ...farms1ToTest])('Farm #%d has an unique pid', (pid) => {
-    const duplicates = farms56.filter((f) => pid === f.pid)
+  it.each([...farmsToTest, ...farms1ToTest])('Farm #%d has an unique pid', async (pid, farm, chainId) => {
+    const farmConfig = await getFarmConfig(chainId)
+    const duplicates = farmConfig.filter((f) => pid === f.pid)
```
chef-jojo, line 35: we should test unique pid between all chains no?

ChefMomota: yeah pid should check bsc chain.
Only lpAddress need check in each chain

chef-jojo (approved):

chefilip (approved):

PR #4877, opened 2022-10-02
feat: Flip tooltip color

ChefMomota:


ChefMomota:
> Is it possible to set text component’s color in the styled css of tooltip component?

Is difficult to set the Text component color in `StyledTooltip`. 
It will cause all the `Link` or some custom Text color be covered

chef-jojo:
Can invert the theme by setting

data-theme=isDark? light : dark

There's no ssr tooltip there
Should be safe to use theme hook

PR #4908, opened 2022-10-06
fix: Network support modal shown in all chains supported pages when wallet network is other than bsc and eth

memoyil:
To reproduce:

Visit the homepage when wallet network is other than bsc and eth, It shows a modal without showing which networks are available

image

### DIFF
--- a/src/components/NetworkModal/NetworkModal.tsx
+++ b/src/components/NetworkModal/NetworkModal.tsx
@@ -1,5 +1,6 @@
 import { ModalV2 } from '@pancakeswap/uikit'
 import useActiveWeb3React from 'hooks/useActiveWeb3React'
+import { CHAIN_IDS } from 'utils/wagmi'
 import { ChainId } from '@pancakeswap/sdk'
 import { useMemo } from 'react'
 import { useNetwork } from 'wagmi'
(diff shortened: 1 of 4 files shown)


src/components/NetworkModal/NetworkModal.tsx
```diff
@@ -25,6 +25,12 @@ export const NetworkModal = ({ pageSupportedChains = SUPPORT_ONLY_BSC }: { pageS
     [chainId, pageSupportedChains],
   )
 
+  const isPageAvailableForAllChains = useMemo(() => {
+    return !pageSupportedChains?.length
+  }, [pageSupportedChains])
```
chefilip, line 30: Since length is number, there is no need to use memo here

chefilip:
IMO, showing modal is more obvious than the display in the wallet at the top corner.

src/components/NetworkModal/NetworkModal.tsx
```diff
@@ -36,7 +39,9 @@ export const NetworkModal = ({ pageSupportedChains = SUPPORT_ONLY_BSC }: { pageS
   if ((chain?.unsupported ?? false) || isPageNotSupported) {
     return (
       <ModalV2 isOpen closeOnOverlayClick={false}>
-        <UnsupportedNetworkModal pageSupportedChains={pageSupportedChains} />
+        <UnsupportedNetworkModal
+          pageSupportedChains={isPageAvailableForAllSupportedChains ? CHAIN_IDS : pageSupportedChains}
```
chefilip, line 43: Could we simplify it a little bit?
```
Boolean(pageSupportedChains?.length) ? pageSupportedChains : CHAIN_IDS
```

Thanks

chefilip (approved):

PR #4909, opened 2022-10-06
fix: (info): Infinite redirect loop in development mode in info page with different network than bsc

memoyil:


chefilip:
why developement mode is different from prod mode?

Chef-Yogi (approved):

PR #4910, opened 2022-10-06
fix: (swap): Import token modal appears if accessing swap page with unsupported network

memoyil:
To reproduce:

1. Set wallet network other than bsc and eth
2. Access http://pancakeswap.finance/swap directly
3. See behind the unsupported network modal there is import token modal and still appears after switching to correct network

### DIFF
--- a/apps/web/src/views/Swap/hooks/useWarningImport.tsx
+++ b/apps/web/src/views/Swap/hooks/useWarningImport.tsx
@@ -35,10 +35,11 @@ export default function useWarningImport() {
   const defaultTokens = useAllTokens()
 
   const importTokensNotInDefault =
-    urlLoadedTokens &&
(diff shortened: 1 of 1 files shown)


src/views/Swap/hooks/useWarningImport.tsx
```diff
@@ -37,7 +37,7 @@ export default function useWarningImport() {
   const importTokensNotInDefault =
     urlLoadedTokens &&
     urlLoadedTokens.filter((token: Token) => {
-      return !(token.address in defaultTokens) && token.chainId === chainId
+      return !(token.address in defaultTokens) && token.chainId === chainId && !isWrongNetwork
```
chefilip, line 40: if user is in wrong network, can se simply return empty array instead of calling filter

chef-jojo (approved):

PR #4917, opened 2022-10-07
fix: Cannot switch network when accessing page via url with query

memoyil:
To reproduce:

1. Switch network to bsc on wallet
2. Go to bsc only page with query id indicates wrong network such as https://pancakeswap.finance/nfts?chainId=1
3. See switch button in modal doesn't work

### DIFF
--- a/src/components/NetworkModal/PageNetworkSupportModal.tsx
+++ b/src/components/NetworkModal/PageNetworkSupportModal.tsx
@@ -1,7 +1,7 @@
 import { Button, Modal, Text, Grid, Box, Message, MessageText } from '@pancakeswap/uikit'
 import { ChainId } from '@pancakeswap/sdk'
 import Image from 'next/future/image'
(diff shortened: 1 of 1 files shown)


src/components/NetworkModal/PageNetworkSupportModal.tsx
```diff
@@ -56,7 +57,7 @@ export function PageNetworkSupportModal() {
           <Button
             variant={foundChain && lastValidPath ? 'secondary' : 'primary'}
             isLoading={isLoading}
-            onClick={() => switchNetworkAsync(ChainId.BSC)}
+            onClick={() => (isWrongNetwork ? switchNetworkLocal(ChainId.BSC) : switchNetworkAsync(ChainId.BSC))}
```
chefilip, line 60: it seems like `isWrongNetwork` logic is leaking in our codebase.

chefilip (approved):

PR #4930, opened 2022-10-10
feat: Remove user gas fee for bsc

memoyil:


src/state/user/hooks/index.tsx
```diff
@@ -404,15 +406,26 @@ export function useRemoveUserAddedToken(): (chainId: number, address: string) =>
...
+      return gasPrice.toString()
+    },
+    {
+      revalidateOnFocus: false,
+      revalidateOnReconnect: false,
+    },
+  )
   const { data } = useFeeData({
     chainId,
     enabled: chainId !== ChainId.BSC && chainId !== ChainId.BSC_TESTNET,
```
chef-jojo, line 424: Can just remove this line and reuse the feedata?

chef-jojo (commented):
Let's remove the user gas setting on redux

src/state/user/hooks/index.tsx
chef-jojo: Suggest keeping this hook as a simple current gas market price hook. 

It shouldn't be used to send transactions

src/hooks/useCallWithGasPrice.ts
```diff
@@ -1,13 +1,10 @@
 import { useCallback } from 'react'
 import { TransactionResponse } from '@ethersproject/providers'
 import { Contract, CallOverrides } from '@ethersproject/contracts'
-import { useGasPrice } from 'state/user/hooks'
 import get from 'lodash/get'
 import { addBreadcrumb } from '@sentry/nextjs'
 
 export function useCallWithGasPrice() {
```
chef-jojo, line 7: Rename this function to prevent confusion

chef-jojo (approved):

PR #4956, opened 2022-10-12
chore(ifo): delete 'of LP' from IFO achievements

Aosiepcs:
We use cake instead LP tokens at IFOs, need delete this. 

We can't input CAKE because it's need a lot value and valuecheck codes, no need.

### DIFF
--- a/src/utils/achievements.ts
+++ b/src/utils/achievements.ts
@@ -31,7 +31,7 @@ export const getAchievementDescription = (campaign: Campaign): TranslatableText 
   switch (campaign.type) {
     case 'ifo':
       return {
-        key: 'Committed more than $5 worth of LP in the %title% IFO',
+        key: 'Committed more than $5 worth in the %title% IFO',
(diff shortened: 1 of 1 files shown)


src/utils/achievements.ts
```diff
@@ -31,7 +31,7 @@ export const getAchievementDescription = (campaign: Campaign): TranslatableText
   switch (campaign.type) {
     case 'ifo':
       return {
-        key: 'Committed more than $5 worth of LP in the %title% IFO',
+        key: 'Committed more than $5 worth in the %title% IFO',
```
Woody810, line 34: 'Committed more than $%amount% %token% in the %title% IFO'

amount: Lowest price, IFO: $100, cIFO: $10
token: Subscription tokens used by IFO.

ChefMomota: Please change the wording to
`Committed more than $5 in the %title% IFO`

PR #4978, opened 2022-10-15
perf: Use poll core farm data only in pools and farms instead of global

memoyil:
Afaik since we don't use core farms to determine cake price we only need in pools to determine other pools earning token prices and farms


ChefMomota:
Can you clear the eslint

ChefMomota (approved):

Issue #4975, opened 2022-10-14
[BUG] Unitary tests


0xBRL:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Product

Other

### Current Behavior

Some unit tests are failing on the `develop` branch:
* [src/views/LimitOrders/utils/getRatePercentageMessage.test.ts](https://github.com/pancakeswap/pancake-frontend/blob/develop/src/views/LimitOrders/utils/getRatePercentageMessage.test.ts)
* [src/__tests__/views/predictions/components/History/helpers.test.ts](https://github.com/pancakeswap/pancake-frontend/blob/develop/src/__tests__/views/predictions/components/History/helpers.test.ts)

Running the unit tests:
```console
$ yarn test
Test Suites: 2 failed, 43 passed, 45 total
Tests:       15 failed, 2 todo, 2715 passed, 2732 total
Snapshots:   0 total
Time:        4.87 s, estimated 5 s
Ran all test suites.
```
🔎 More info: [test.log](https://github.com/pancakeswap/pancake-frontend/files/9788464/info.log)

### Expected Behavior

All unit tests are expected to pass:
```console
$ yarn test
Test Suites: 0 failed, 45 passed, 45 total
Tests:       0 failed, 2 todo, 2730 passed, 2732 total
Snapshots:   0 total
Time:        4.87 s, estimated 5 s
Ran all test suites.
```

### Steps To Reproduce

1. Clone the pancake-frontend project
2. Install dependencies
3. Run the tests locally

```console
$ git clone https://github.com/pancakeswap/pancake-frontend.git
$ yarn
$ yarn test
```

### Environment

```markdown
- Device: MacBook Pro (M1)
- OS: macOS Monterey 12.6
- Browser: N/A
- Wallet: N/A
- Node: v16.17.1
```

0xBRL:
Heey @chef-jojo, I was trying to run some local tests and I find these errors.
However these errors don't happen in **Unit tests (Jest) (16)** job in the GitHub, let me know if these tests need to be fixed 🤔
I can work on it 😀

chef-jojo:
yeah i think it's something related to `toLocaleString`. We should explicit setting the locale setting on jest test env

PR #4985, opened 2022-10-17
fix: define the locale setting in test environment

0xBRL:
More info: #4975


chef-jojo (approved):
Should be good to go. Thanks

PR #4993, opened 2022-10-17
feat: (homepage): Bring back username visibility toggle

memoyil:


Woody810:
There is no toggle in the profile page

src/views/TradingCompetition/components/RegisterModal/RegisterWithProfile.tsx
```diff
@@ -44,7 +46,7 @@ const RegisterWithProfile: React.FC<React.PropsWithChildren<CompetitionProps>> =
 
   return (
     <>
-      <Heading scale="md" mb="24px">{`@${profile.username}`}</Heading>
+      <Heading scale="md" mb="24px">{`@${userUsernameVisibility ? profile.username : '🐰🐰🐰🐰🐰🐰'}`}</Heading>
```
chef-jojo, line 49: i think we can have a hook to merging profile + visibility and the output

chef-jojo:
Should we show username by default?

chef-jojo:
> > Should we show username by default?
> 
> The default value of that parameter is false, if want to make it default true we should add migrate in redux persist or create a new parameter

yeah it's tricky. That's why we have some flags are `disabledXXXX` instead

chef-jojo (dismissed):

chef-jojo (dismissed):

chef-jojo (approved):

PR #5031, opened 2022-10-21
perf: (swap): Don't render access risk component when it is not supported

memoyil:


chefilip:
@ChefMomota  Could you recall why we applied `style` in the first place?

chef-jojo (approved):

PR #5033, opened 2022-10-21
fix: (swap): Chart button visible when chain is not supported

memoyil:


chefilip:
I think we should create `isChartSupported` as a context so we don't have to pass in multi components.

chef-jojo (approved):

PR #5116, opened 2022-10-28
fix: missing border on small devices

chefjackson:
Before: the pagination is inside the scrollable container
image
After: move the pagination outside of the scrollable container
image


chef-jojo (approved):
lgtm

PR #5119, opened 2022-10-28
feat: Aptos add useCatchTxError

ChefMomota:


chefilip (changes requested):
why is there a local file in the commit changed?

ChefMomota:
move to here
https://github.com/pancakeswap/pancake-frontend/pull/5130

PR #5128, opened 2022-10-28
feat: (liquidity): Introduce lp apr info when adding eth liquidity

memoyil:


chefilip (dismissed):

apps/web/src/state/swap/useLPApr.ts
```diff
@@ -22,26 +22,34 @@ interface PoolReserveVolumeResponse {
...
+      ? 'ETH'
+      : null
+    : null
   const { data: poolData } = useSWRImmutable(
-    pair && pair.chainId === ChainId.BSC ? ['LP7dApr', pair.liquidityToken.address] : null,
+    chainName ? ['LP7dApr', pair.liquidityToken.address, chainName] : null,
     async () => {
       const timestampsArray = getDeltaTimestamps()
-      const blocks = await getBlocksFromTimestamps(timestampsArray, 'desc', 1000)
+      const blocks = await getBlocksFromTimestamps(timestampsArray, 'desc', 1000, chainName)
```
chefilip, line 36: I think we could move the chainName logic into the `getBlocksFromTimestamps`. 
If the last parameter is string, we treat it as `chainName`
If the last parameter is number we treat it as `chainId`

What do you think?

PR #5133, opened 2022-10-30
feat(TradingCompetitions): Add subMenus

Woody810:
Before:
image

After:
image


ChefMomota:
Imo, we do no need to change the submenu style.

Woody810:
> Imo, we do no need to change the submenu style.



> Imo, we do no need to change the submenu style.

Prediction, Lottery, and Pottrty all have subMenu for switching between each other, but Trading Competion does not. I think UI operations should be consistent.

PR #5146, opened 2022-10-31
refactor: (swap): Use context for swap supported feature instead of props

memoyil:


apps/web/src/views/Swap/SwapFeaturesContext.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  const isAccessTokenSupported = useMemo(() => ACCESS_TOKEN_SUPPORT_CHAIN_IDS.includes(chainId), [chainId])
+
+  useEffect(() => {
+    setUserChartPreference(isChartDisplayed)
+  }, [isChartDisplayed, setUserChartPreference])
+
+  return (
+    <SwapFeaturesContext.Provider
+      value={{
```
chefilip, line 53: Could you use useMemo for value object? Thanks

chefilip (approved):

PR #5151, opened 2022-10-31
fix: (swap): Derived swap price chart data

memoyil:


Chef-Yogi (dismissed):

Chef-Yogi:
should we remove the same pattern from `pairHourDatasByIds.ts`

Chef-Yogi (approved):

PR #5153, opened 2022-11-01
feat(aptos-pools): Add Pools

chefilip:
- [x] polish decimal display


chef-jojo:
no need to add to swap sdk since it's unrelated

chefilip:
> no need to add to swap sdk since it's unrelated

Hmm. I'm still thinking where I could add all contract info in one place for accessibility.

PR #5189, opened 2022-11-03
feat(Limit): Add scan risk

Woody810:
image

### DIFF
--- a/apps/web/src/views/LimitOrders/index.tsx
+++ b/apps/web/src/views/LimitOrders/index.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useState, useMemo } from 'react'
+import { useCallback, useEffect, useState, useMemo, useContext } from 'react'
 import { useRouter } from 'next/router'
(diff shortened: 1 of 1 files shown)


apps/web/src/views/LimitOrders/index.tsx
```diff
@@ -408,6 +408,9 @@ const LimitOrders = () => {
                       showCommonBases
                       commonBasesType={CommonBasesType.SWAP_LIMITORDER}
                     />
+                    <Box>
```
ChefMomota, line 411: Add an array to check that is network supported?
The red alarm api only work in bsc

apps/web/src/views/LimitOrders/index.tsx
```diff
@@ -408,6 +408,9 @@ const LimitOrders = () => {
...
+                      <AccessRisk inputCurrency={currencies.input} outputCurrency={currencies.output} />
+                    </Box>
```
ChefMomota, line 413: Add an array to check that is network supported?
The red alarm api only work in bsc

Woody810: Done.

ChefMomota (approved):

PR #5215, opened 2022-11-06
refactor: Use account hook

memoyil:


chefilip:
Could I ask What the benefit of the replacement is?

chef-jojo:
> @chef-jojo This issue is fixed
> 
> https://github.com/pancakeswap/pancake-frontend/blob/5053369419a94a874029b0d82965803551355049/packages/wagmi/src/useWeb3React.ts#L9
> 
> in here
> https://github.com/pancakeswap/pancake-frontend/blob/5053369419a94a874029b0d82965803551355049/packages/awgmi/core/src/accounts/account.ts#L47
> 
> correct? therefore useAccount hook can be used without web3react hook

Need to double check the wagmi version one

chef-jojo (approved):

PR #5221, opened 2022-11-07
feat: smart router

chefjackson:


packages/smart-router/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "prepublishOnly": "yarn run build",
+    "clean": "rm -rf .turbo && rm -rf node_modules && rm -rf dist"
+  },
+  "dependencies": {
+    "@pancakeswap/farms": "*",
+    "@pancakeswap/multicall": "*",
+    "@pancakeswap/sdk": "*",
+    "@pancakeswap/tokens": "*",
+    "@pancakeswap/token-lists": "*",
+    "ethers": "^5.7.2",
```
chef-jojo, line 22: might need to fix the version, better split `ethers`

packages/smart-router/evm/getStableSwapPairs.ts
```diff
@@ -0,0 +1,23 @@
+import { ChainId, CurrencyAmount, Pair } from '@pancakeswap/sdk'
+import { getStableConfig } from '@pancakeswap/farms/constants'
```
chef-jojo, line 2: shouldn't rely on farms config 🤔 
that might be a case that stableswap pair is not in farms
let's add a constants for stableswap

packages/smart-router/evm/getStableSwapPairs.ts
```diff
@@ -0,0 +1,23 @@
...
+import { deserializeToken } from '@pancakeswap/token-lists'
+
+import { StableSwapPair } from './types'
+import { createStableSwapPair } from './stableSwap'
+
+export async function getStableSwapPairs(chainId: ChainId): Promise<StableSwapPair[]> {
+  // Stable swap is only supported on BSC chain
+  if (chainId !== ChainId.BSC) {
```
chef-jojo, line 10: can add testnet constants

packages/smart-router/package.json
chefjackson: OK.

packages/smart-router/evm/getStableSwapPairs.ts
chefjackson: 🤔 I thought we have farms for all pairs we supported. Let me add it's own config.

chefjackson: Do we have a lp config that maintains all token pairs that we support?

packages/smart-router/evm/getStableSwapPairs.ts
chefjackson: done

packages/smart-router/package.json
chefjackson: Done

chef-jojo (approved):

PR #5229, opened 2022-11-08
fix: use own stable swap pool config instead of depending on farms

chefjackson:
Extract stable swap config from farms

### DIFF
--- a/packages/smart-router/evm/constants/stableSwap/index.ts
+++ b/packages/smart-router/evm/constants/stableSwap/index.ts
@@ -1,0 +1,8 @@
+import { ChainId } from '@pancakeswap/sdk'
+
+import { StableSwapPool } from '../../types/pool'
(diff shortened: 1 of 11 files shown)


packages/smart-router/evm/constants/stableSwap/index.ts
```diff
@@ -0,0 +1,8 @@
+import { ChainId } from '@pancakeswap/sdk'
+
+import { StableSwapPool } from '../../types/pool'
+import { poolMap } from './pools'
+
+export async function getStableSwapPools(chainId: ChainId): Promise<StableSwapPool[]> {
```
chef-jojo, line 6: no need async

chefjackson: fixed

chef-jojo (approved):

PR #5261, opened 2022-11-10
feat: Add route type to identify trade with stable swap

chefjackson:
Add route type to identify trade with stable swap


chef-jojo (approved):
lgtm

PR #5295, opened 2022-11-14
test: Setup test framework & add basic test case example

chefjackson:


packages/smart-router/package.json
```diff
@@ -10,6 +10,7 @@
   "scripts": {
     "build": "tsup",
     "dev": "tsup --watch",
+    "test": "NODE_ENV=test jest --verbose --runInBand",
```
chefilip, line 13: could we use `vitest` like other packages?

chef-jojo (approved):

PR #5304, opened 2022-11-14
feat: (prediction): Use server time for prediction

memoyil:
currently if user client date/time is wrong it shows wrong info in prediction

### DIFF
--- a/apps/web/src/hooks/useServerTimestamp.ts
+++ b/apps/web/src/hooks/useServerTimestamp.ts
@@ -12,7 +12,7 @@ const useServerTimestamp = () => {
       const res = await fetch('https://worldtimeapi.org/api/timezone/Etc/UTC', { method: 'GET', cache: 'no-store' })
       if (res.ok) {
         const data = await res.json()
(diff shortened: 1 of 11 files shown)


apps/web/src/hooks/useServerTimestamp.ts
```diff
@@ -0,0 +1,34 @@
...
+import { useCallback } from 'react'
+
+const useServerTimestamp = () => {
+  const { mutate } = useSWRConfig()
+  const { data: lastCheck } = useSWRImmutable('serverTimestampLastCheck')
+
+  const { data: serverTimestamp } = useSWR(
+    'serverTimestamp',
+    async () => {
+      const res = await fetch('https://worldtimeapi.org/api/timezone/Etc/UTC', { method: 'GET', cache: 'no-store' })
```
chef-jojo, line 12: Let's hold it first. Make sure we don't depends on this api

chefilip (commented):
I think we should keep the client side as fallback if api is failed

PR #5312, opened 2022-11-15
feat: add static methods to trade & route with stable swap

chefjackson:


packages/smart-router/evm/trade.ts
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Get the maximum amount in that can be spent via this trade for the given slippage tolerance
+ */
+function maximumAmountIn<TIn extends Currency, TOut extends Currency, TTradeType extends TradeType>(
+  trade: TradeWithStableSwap<TIn, TOut, TTradeType>,
+  slippageTolerance: Percent,
+): CurrencyAmount<TIn> {
+  invariant(!slippageTolerance.lessThan(ZERO), 'SLIPPAGE_TOLERANCE')
+  if (trade.tradeType === TradeType.EXACT_INPUT) {
```
Chef-Yogi, line 32: is this condition always be true while we are using smart router?

packages/smart-router/evm/trade.ts
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Get the minimum amount that must be received from this trade for the given slippage tolerance
+ */
+function minimumAmountOut<TIn extends Currency, TOut extends Currency, TTradeType extends TradeType>(
+  trade: TradeWithStableSwap<TIn, TOut, TTradeType>,
+  slippageTolerance: Percent,
+): CurrencyAmount<TOut> {
+  invariant(!slippageTolerance.lessThan(ZERO), 'SLIPPAGE_TOLERANCE')
+  if (trade.tradeType === TradeType.EXACT_OUTPUT) {
```
Chef-Yogi, line 50: will this one always false

packages/smart-router/evm/trade.ts
chefjackson: since we now support both exact in and exact out, so we could still have different conditions met. Like when the trade is v2 not mixed

PR #5318, opened 2022-11-15
feat:smart router integration with BE API

Chef-Yogi:


chefjackson:
Can we move `useDerivedSwapInfoWithStableSwap` into the new SmartSwap directory? Maybe under `Swap/SmartSwap/hooks`

PR #5332, opened 2022-11-16
feat: get best trade from smart router

chefjackson:


Woody810:
image
I think it should be put together

chefjackson:
> image I think it should be put together

Yes it's gonna be put together. It's a temp version now

chef-jojo:
do you need changset?

Chef-Yogi (approved):

PR #5358, opened 2022-11-17
feat(Swap): Add share button

Woody810:
image


chefilip:
I think we could simply copy the URL right? For the simplicity purpose, I think we should selectively add features which users couldn't do it. Less is better.

Woody810:
> I think we could simply copy the URL right? For the simplicity purpose, I think we should selectively add features which users couldn't do it. Less is better.

I saw that there is this button on aptos, so I think bsc should be the same. Or are there different considerations on aptos?🤔

chefilip:
> > I think we could simply copy the URL right? For the simplicity purpose, I think we should selectively add features which users couldn't do it. Less is better.
> 
> I saw that there is this button on aptos, so I think bsc should be the same. Or are there different considerations on aptos?🤔

In Aptos, the token address is lengthy. That's why we put a copy there. I think in BSC, we don't need it.

PR #5359, opened 2022-11-17
feat(aptos-Settings): Distinguish between global and swap_liquidity

Woody810:
Global:
image

Swap&Liquidity:
image


chef-jojo:
I don't see the reason why splitting them especially the setting options are not many

Woody810:
> I don't see the reason why splitting them especially the setting options are not many

I think it needs to be the same as the BSC version, maybe when there are enough options, then split the settings.

Thanks for the reminder, this PR will be closed🙏

PR #5367, opened 2022-11-18
feat: integrate wih new router SC

Chef-Yogi:


apps/web/src/views/Swap/SmartSwap/components/TransactionConfirmSwapContent.tsx
```diff
@@ -0,0 +1,126 @@
...
+    !tradeA.inputAmount.currency.equals(tradeB.inputAmount.currency) ||
+    !tradeA.inputAmount.equalTo(tradeB.inputAmount) ||
+    !tradeA.outputAmount.currency.equals(tradeB.outputAmount.currency) ||
+    !tradeA.outputAmount.equalTo(tradeB.outputAmount)
+  )
+}
+
+interface TransactionConfirmSwapContentProps {
+  trade: TradeWithStableSwap<Currency, Currency, TradeType> | undefined
+  originalTrade: ITrade | undefined
```
chefjackson, line 33: I think if we have the `ConfirmSwapModal` component duplicated, we don't need `ITrade` type here anymore? This component simply works for trade with stable swap

apps/web/src/views/Swap/SmartSwap/components/TransactionConfirmSwapContent.tsx
```diff
@@ -0,0 +1,126 @@
...
+        : currencyBalances[Field.INPUT]
+      : null
+    return inputCurrencyAmount && slippageAdjustedAmounts && slippageAdjustedAmounts[Field.INPUT]
+      ? inputCurrencyAmount.greaterThan(slippageAdjustedAmounts[Field.INPUT]) ||
+          inputCurrencyAmount.equalTo(slippageAdjustedAmounts[Field.INPUT])
+      : false
+  }, [currencyBalances, trade, slippageAdjustedAmounts])
+
+  const modalHeader = useCallback(() => {
+    const SwapModalHead = isStable ? StableSwapModalHeader : SwapModalHeader
```
chefjackson, line 83: We also don't need to worry about if it's stable swap or not? Cuz if this component is used, it means that the trade must include stable swap, otherwise it would fallback to the old commit button

chefjackson: I think maybe we should duplicate the `SwapModalHead` as well and use it directly. Of course the best thing would be decouple the modal header component with the trade type, but let's just duplicate for now

apps/web/src/views/Swap/SmartSwap/components/TransactionConfirmSwapContent.tsx
```diff
@@ -0,0 +1,126 @@
...
+    showAcceptChanges,
+    trade,
+    slippageAdjustedAmounts,
+    isEnoughInputBalance,
+    isStable,
+  ])
+
+  const modalBottom = useCallback(() => {
+    return trade ? (
+      <SwapModalFooter
```
chefjackson, line 113: The footer can be duplicated as well

apps/web/src/views/Swap/SmartSwap/hooks/useSwapCallArguments.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    if (!contract) {
+      return []
+    }
+
+    const swapMethods = []
+    if (trade.tradeType === TradeType.EXACT_INPUT) {
+      swapMethods.push(
+        swapCallParameters(trade, {
+          feeOnTransfer: true,
```
chefjackson, line 57: Are we supporting `feeOnTransfer` by default now? If so we don't need this option anymore

apps/web/src/views/Swap/SmartSwap/hooks/useSwapCallArguments.ts
```diff
@@ -0,0 +1,115 @@
...
+    //     [srcToken,dstToken,amount,minReturn,flag]
+    args = [path[0], path[1], amountIn, amountOut, flag[0]]
+    value = etherIn ? amountIn : ZERO_HEX
+  }
+  // multiHop
+  else {
+    methodName = 'swapMulti'
+    //     [tokens,amount,minReturn,flag]
+    args = [path, amountIn, amountOut, flag]
+    value = etherIn ? amountIn : ZERO_HEX
```
chefjackson, line 108: Just a note we need to test the case of native currency as output

apps/web/src/hooks/useSwapCallback.ts
```diff
@@ -136,11 +135,15 @@ export function useSwapCallback(
             const inputAmount =
               trade.tradeType === TradeType.EXACT_INPUT
                 ? trade.inputAmount.toSignificant(3)
-                : trade.maximumAmountIn(pct).toSignificant(3)
+                : isV2SwapOrStableSwap(trade)
```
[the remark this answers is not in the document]
Chef-Yogi: ok I will separate it

Chef-Yogi:
will adjust the code in another PR

PR #5403, opened 2022-11-23
fix: Hide recipient on smart swap

chefjackson:


apps/web/src/views/Swap/SmartSwap/index.tsx
```diff
@@ -224,6 +224,8 @@ export function SmartSwapForm() {
     }
   }, [onUserInput, allowUseSmartRouter, independentField, tradeInfo?.inputAmount])
 
+  const allowRecipient = isExpertMode && !showWrap && !smartRouterOn
```
ChefMomota, line 227: Can we also add `recipient === null` in here?

Chef-Yogi (approved):

PR #5471, opened 2022-11-30
feat: (stableswap): Add currency select to stable swap

memoyil:


chefjackson:
Thanks @memoyil ! It's good to have currency select on the old form.
Although I'm thinking of removing the global settings so we'll switch to use the new form entirely and deprecate the old forms. 🤔

chefjackson (approved):
Let's add support to the old form. But wait until new stable pairs are launched

chefjackson (approved):

ChefMomota (approved):

PR #5472, opened 2022-11-30
fix: (smartrouter): Stable swap cannot fallback to v2 when insufficient liquidity balance

memoyil:


chefilip (dismissed):

chefjackson (approved):
LGTM. Thanks for the fix!

chefilip (approved):

PR #5498, opened 2022-12-02
refactor: Replaced let with const

thouravi:
Used the `const` keyword instead of `let` wherever possible to improve readability and avoid reassigning variables.


apis/farms/src/handler.ts
```diff
@@ -70,7 +70,7 @@ export async function saveFarms(chainId: number, event: ScheduledEvent | FetchEv
   try {
     const isTestnet = farmFetcher.isTestnet(chainId)
     const farmsConfig = await (await fetch(`${farmConfigApi}/${chainId}.json`)).json<SerializedFarmConfig[]>()
-    let lpPriceHelpers: SerializedFarmConfig[] = []
+    const lpPriceHelpers: SerializedFarmConfig[] = []
```
chefilip, line 73: it will throw eslint `Cannot assign to 'lpPriceHelpers' because it is a constant.ts`

PR #5500, opened 2022-12-03
feat(Farms): Sort filter add stable option

Woody810:


apps/aptos/components/Farms/components/Farms.tsx
```diff
@@ -348,6 +350,10 @@ const Farms: React.FC<React.PropsWithChildren> = ({ children }) => {
                     label: t('Latest'),
                     value: 'latest',
                   },
+                  {
+                    label: t('Stable'),
```
ChefMomota, line 354: Rename to `Stableswap`

apps/web/src/views/Farms/Farms.tsx
```diff
@@ -420,6 +422,10 @@ const Farms: React.FC<React.PropsWithChildren> = ({ children }) => {
                     label: t('Latest'),
                     value: 'latest',
                   },
+                  {
+                    label: t('Stable'),
```
ChefMomota, line 426: Rename to `Stableswap`

apps/aptos/components/Farms/components/Farms.tsx
```diff
@@ -348,6 +350,10 @@ const Farms: React.FC<React.PropsWithChildren> = ({ children }) => {
...
+                    value: 'stable',
+                  },
```
ChefMomota, line 356: On Aptos side, we can hide it first.

apps/web/src/views/Farms/Farms.tsx
Woody810: Done.

apps/aptos/components/Farms/components/Farms.tsx
Woody810: Done.

ChefMomota:
@Woody810  You changed wrong file

chefilip:
I think we could add a Stable toggle to show only Stable farms instead of using sort. Using sort stable, it shows no-stable farms as well.

ChefMomota:
But for toggles, it looks like we don't have enough space for another toggle UI.

Woody810:
> @Woody810 You changed wrong file

oops...

chefilip:
One idea is Moving toggles into a popup like settings so we could add more toggles. We should seek for designer'a consult about this before merging the PR

PR #5506, opened 2022-12-05
fix: info token price chart issue

Chef-Yogi:


apps/web/src/state/info/queries/tokens/priceData.ts
```diff
@@ -41,12 +41,17 @@ const fetchTokenPriceData = async (
...
-  const timestamps = []
+  let timestamps = []
   let time = startTimestamp
   while (time <= endTimestamp) {
     timestamps.push(time)
     time += interval
   }
+  if (chainName === 'BSC' && !checkIsStableSwap()) {
+    const latestBlock = await getBlocksFromTimestamps([endTimestamp], 'asc', 500, chainName)[0]
+    timestamps = timestamps.filter((t) => t < latestBlock - 32) // nodeReal will sync the the 32 block before latest
```
chefjackson, line 52: Isn't `latestBlock` type of `Block`?

Chef-Yogi: I had update the code, this version has some issue

ChefMomota (approved):

apps/web/src/state/info/queries/tokens/priceData.ts
```diff
@@ -49,7 +49,13 @@ const fetchTokenPriceData = async (
   }
   try {
     const blocks = await getBlocksFromTimestamps(timestamps, 'asc', 500, chainName)
-    if (!blocks || blocks.length === 0) {
+    const blocksLength = blocks?.length ?? 0
+    if (blocksLength > 0 && chainName === 'BSC' && !checkIsStableSwap()) {
+      const data = blocks[blocksLength - 1]
+      blocks[blocksLength - 1] = { timestamp: data.timestamp, number: data.number - 32 }
```
chefjackson, line 55: Should we filter all blocks whose height is greater than `latest - 32`? Seems like now we are just changing the block number of the last block and the corresponding timestamp might be incorrect?

PR #5511, opened 2022-12-06
fix: Strengthen the round limit of getRoundIdsArray 

TroubleSeven:
found that when currentLotteryId = 1,  getRoundIdsArray will return error round id list.
`getRoundIdsArray(1) => output:  [-1, 0, 1]`
add roundId check in the function of getRoundIdsArray to avoid this error 
Although it will not be triggered according to the current round :)


chefilip (approved):

TroubleSeven:
> Could you check prettier? Thanks for your contribution.

has updated :)

ChefMomota (approved):

PR #5533, opened 2022-12-08
feat: smart router updates

Chef-Yogi:


apps/web/src/views/Swap/index.tsx
```diff
@@ -76,15 +70,16 @@ export default function Swap() {
           <StyledSwapContainer $isChartExpanded={isChartExpanded}>
             <StyledInputCurrencyWrapper mt={isChartExpanded ? '24px' : '0'}>
               <AppBody>
-                {smartSwap ? (
```
chefjackson, line 79: I think we can just remove the old code

Chef-Yogi: ok

chefjackson (approved):

ChefMomota (approved):

PR #5612, opened 2022-12-15
style(TokenManage): SearchBox is not truncated

Woody810:
Before:
image

After:
image


chefjackson (commented):
Shall we remove the overflow restriction of the outside container?

ChefMomota:
@Woody810  Could you also fixing this?
Screenshot 2022-12-16 at 10 43 21 AM

apps/web/src/components/SearchModal/ManageTokens.tsx
```diff
@@ -94,7 +94,7 @@ export default function ManageTokens({
     <Wrapper>
       <Column style={{ width: '100%', flex: '1 1' }}>
         <AutoColumn gap="14px">
-          <Row>
+          <Flex width="96%" justifyContent="center" ml="2%">
```
chefjackson, line 97: Is it possible to remove the overflow hidden on the outside container instead of tweaking the width of the input field?

Woody810: I've tried setting Row's `overflowX: visible `, but it not working🤦‍♂️

Woody810:
> @Woody810 Could you also fixing this? Screenshot 2022-12-16 at 10 43 21 AM

done.

apps/web/src/components/SearchModal/ManageTokens.tsx
chefjackson: I think should add `overflow: visible` to the `ModalBody` component.
But the current dom structure has a problem. Need to move the scroll bar from the whole modal body to the inner list first

Woody810: @chefjackson  Fixed. Thanks for your help🙏.

Woody810: > @chefjackson Fixed. Thanks for your help.🙏

ChefMomota (approved):

PR #5682, opened 2022-12-21
chore: Update whwEth word

chefilip:


Chef-Yogi (approved):

ChefMomota:
We should also update symbol at token-list repo

PR #5714, opened 2022-12-25
feat(Farms): Menu add finished tag

Woody810:
image


Woody810:
> Imho this info should be in harvest card instead of here since it is related to user

It's great to put it on the homepage as a reminder, I'll think about how to present it.

But I think it is necessary to put it in the menu, because users may set the Swap page as a bookmark, so that they cannot be reminded.

ChefMomota:
Imo adding this for users is pretty confuse.
More prefer current state

PR #5724, opened 2022-12-27
refactor: Remove duplicate isTradeBetter util

memoyil:
Sonar fails for %8 duplication for one file but we cannot use the same util for aptos since trade types are different


chefjackson (approved):
lgtm

chefilip:
Could you check unit test? @memoyil

PR #5787, opened 2023-01-04
feat: (swap): Use lightweight charts for swap instead of recharts

memoyil:
It reduces the bundle size (~150kb), improves user experience and better performance


chef-jojo:
Need to take care of the mobile

ChefMomota (approved):

PR #5802, opened 2023-01-08
Big bug PancakeSwap Info & Analytics 2021 version

FuturIsHere:
Hi there,

I have a big problem on the display in the Info part.
I'm still working on the old version of PancakeSwap which dates from 2021 and the token and pool info no longer wants to be displayed. A few months ago it worked from time to time but today nothing is displayed. I don't know if the problem is with the subgraphs or not, but I'm completely lost.
I tried looking on the 2022 PancakeSwap (which has a slightly different architecture than the 2021) and I see that the new URL for INFO_CLIENT in the src/config/constant/endpoints.ts file is https://proxy-worker/.pancake-swap.workers.dev/bsc-exchange instead of https://bsc.streamingfast.io/subgraphs/name/pancakeswap/exchange-v2

By replacing the old link with the new one I manage to display the amount of liquidity, the volume in 24 hours (but not the graphs) and all the "Transaction" section but the "token" and "pools" section do not want show up.

How to make the 2 sections "token" and "pools" work again with the old versions of Pancakeswap 2021?

Capture d’écran 2022-12-29 à 21 38 23
Capture d’écran 2022-12-29 à 21 39 07
Capture d’écran 2022-12-29 à 21 39 20


chef-jojo:
streamingfast one is deprecated. 
[please register your own endpoint here](https://nodereal.io/meganode/api-marketplace/pancakeswap-graphql)

PR #5829, opened 2023-01-11
fix(aptos-pool): Avoid APR mismatch for pairs with different decimals

chefilip:
APR shows incorrectly for pairs with different decimals


chef-jojo (approved):

chef-jojo:
let's also double check all implementation of getPoolApr later

PR #5981, opened 2023-01-31
fix: eth farm info path when wallet unconnected

Chef-Yogi:


chefilip (approved):

roymoses07:
approve these changes

PR #5998, opened 2023-02-03
chore(Voting): Change "Vote Now" style

Woody810:
Before:
image

After:
image


chefilip:
Even though the change makes UI more consistent, I would prefer the original one. It's more highlighted.

PR #6021, opened 2023-02-06
chore(Swap): Change default price display

Woody810:
When swap from A to B, the default price display should also be A→B.

Before:
image

After:
image


chefilip:
Even though your change is easy to understand, it's duplicated with the information of input -> ouput. Also, It's not aligned with the definition of price.

roymoses07 (dismissed):
Done

chef-jojo:
@Woody810 can you revert the order B -> A?

Woody810:
> @Woody810 can you revert the order B -> A?

Done

chef-jojo (approved):

PR #6026, opened 2023-02-07
feat(Blog): Update UI

ChefMomota:


apps/blog/.env
```diff
@@ -1 +1,2 @@
-NEXT_PUBLIC_APEX_URL = "https://pancakeswap.finance"
\ No newline at end of file
+NEXT_PUBLIC_APEX_URL = "https://pancakeswap.finance"
+NEXT_PUBLIC_GTAG = "G-334KNG3DMQ"
```
chef-jojo, line 2: Let's set in vercel instead of here to avoid forks/local

ChefMomota: done

chef-jojo (approved):

PR #6058, opened 2023-02-09
fix: Aptos pool reset to 0

memoyil:
When 'watch: true' used it tries to pool onchain data very frequently which some of the calls results in error if so many users trying to fetch therefore it shows the default value for userdata which is 0


chefilip (approved):

chef-jojo:
It won't update price if you remove the watch?

PR #6073, opened 2023-02-11
feat(Tooltip): Use click to show on mobile

Woody810:


packages/uikit/src/hooks/useTooltip/useTooltip.tsx
```diff
@@ -32,7 +32,7 @@ const useTooltip = (content: React.ReactNode, options?: TooltipOptions): Tooltip
   const { isDark } = useTheme();
   const {
     placement = "auto",
-    trigger = "hover",
+    trigger = isTouchDevice() ? "click" : "hover",
```
chefilip, line 35: It seems this line would be called on server-side, right? On server-side, `isTouchDevice` doesn't work as expect.

ChefMomota (approved):

PR #6097, opened 2023-02-14
feat(web): Add MM loading status

chef-jojo:


apps/web/src/views/Swap/SmartSwap/index.tsx
```diff
@@ -183,12 +189,16 @@ export const SmartSwapForm: React.FC<{ handleOutputSelect: (newCurrencyOutput: C
         [Field.INPUT]:
           independentField === Field.INPUT
             ? parsedAmount
+            : isMMLoading
```
Chef-Yogi, line 192: Not sure if this will cause issue when MM loading, and AMM have result, but we just show undefined for a while?

chef-jojo: We should wait for MM complete

Chef-Yogi: Got it👌

Chef-Yogi (approved):

PR #6127, opened 2023-02-15
feat(Pools): When two EndsBlock are the same, only one is displayed

Woody810:
Before:
image

After:
image


chefilip (approved):

chefilip:
> I think we should keep both even if they are the same

What is your reasoning by this? To me, the info is quite duplicated. As a user, I could implicitly think "stake limit will end when the pool is end"

Chef-Yogi (approved):

PR #6136, opened 2023-02-16
fix: LP fee not accurate with axlUSDC

chefjackson:


Chef-Yogi (approved):

Chef-Yogi:
CleanShot 2023-02-16 at 14 54 12@2x
should we change this also?

chefjackson:
> CleanShot 2023-02-16 at 14 54 46@2x `packages/smart-router/evm/constants/exchange.ts` should we change this also?

Can just leave it for now since we're using the info contract configured on each pool

PR #6144, opened 2023-02-17
chore: Token highlight tracking code

Chef-Yogi:


apps/aptos/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "extends": "tsconfig/nextjs.json",
-  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx"],
+  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", "../web/src/hooks/useReactGAInitialize.ts"],
```
chef-jojo, line 3: 🤔

chef-jojo (changes requested):
Let's try not install another tracking pkg

PR #6153, opened 2023-02-18
chore(Farms): Change tag placement of frams

Woody810:
I think Boosted is the same category as StableLP, so they should be placed in the same place.

Before:
image
image

After:
image
image


ChefMomota:
Fix eslint

ChefMomota (approved):

PR #6165, opened 2023-02-19
feat(aptos-Swap&Liquidity): Add token USD price

Woody810:
image
image


Woody810:
> Input panel height when first load No need to show when no input value

Done.

chef-jojo:
input panel height can be further improved. try reload the page to see the first paint

Woody810:
> input panel height can be further improved. try reload the page to see the first paint

Done

chef-jojo (approved):

PR #6171, opened 2023-02-20
refactor: Remove row components from apps

memoyil:


chefjackson:
Close for now as it's stale

PR #6178, opened 2023-02-21
fix: Use mm quote count down only when there is a trade

memoyil:
It renders the swap page every second even if there is no trade from market maker (in bnb chain for example)

Include fixes: 

  1. Stop order book fetching when setting is off
  2. Stop fetching when rfq is connected
  3. Confirm modal shows indicator when rfq reloading


apps/web/src/views/Swap/SmartSwap/index.tsx
```diff
@@ -176,6 +176,7 @@ export const SmartSwapForm: React.FC<{ handleOutputSelect: (newCurrencyOutput: C
   })
 
   const { remainingSec: mmQuoteExpiryRemainingSec } = useMMQuoteCountDown(
+    mmRFQTrade?.trade || mmOrderBookTrade?.trade,
```
chef-jojo, line 179: Only need rfq trade to be countdown

chef-jojo:
Let's keep orderbook result when unconnected

Chef-Yogi (approved):

PR #6262, opened 2023-03-06
fix(Aptos): Swap USD value

ChefMomota:
Screenshot 2023-03-06 at 5 05 29 PM
Screenshot 2023-03-06 at 5 05 52 PM

### DIFF
--- a/apps/aptos/hooks/useStablePrice.ts
+++ b/apps/aptos/hooks/useStablePrice.ts
@@ -1,5 +1,5 @@
 import { Currency, JSBI, Price, Trade } from '@pancakeswap/aptos-swap-sdk'
-import { L0_USDC, CAKE } from 'config/coins'
+import { L0_USDC, CE_USDC, WH_USDC, CAKE } from 'config/coins'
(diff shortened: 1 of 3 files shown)


apps/aptos/hooks/useStablePrice.ts
```diff
@@ -22,7 +22,12 @@ export default function useStablePrice(currency?: Currency): Price<Currency, Cur
   const native = useNativeCurrency(chainId)
   const wrapped = currency?.wrapped
   const wnative = native.wrapped
-  const stable = L0_USDC[chainId]
+  const stableArray = [
+    { key: 'lz', coin: L0_USDC[chainId] },
+    { key: 'wh', coin: WH_USDC[chainId] },
+    { key: 'ce', coin: CE_USDC[chainId] },
+  ]
+  const stable = stableArray.find((stableCoin) => currency?.symbol.startsWith(stableCoin.key))?.coin ?? L0_USDC[chainId]
```
chef-jojo, line 30: We should probably need to find all currency-stable pairs and get the most liquidity one.

PR #6328, opened 2023-03-15
fix: LP Apr display for usdc/busd stable pair

memoyil:


chefilip:
I think it's no longer the issue for now

PR #6341, opened 2023-03-17
chore: Bump up jotai to 2.0.3

memoyil:


apps/web/src/utils/atomWithStorageWithErrorCatch.ts
```diff
@@ -1,15 +1,16 @@
 /* eslint-disable no-restricted-globals */
 /* eslint-disable @typescript-eslint/no-shadow */
-import { atom } from 'jotai'
-import type { WritableAtom } from 'jotai'
-import { RESET, unstable_NO_STORAGE_VALUE as NO_STORAGE_VALUE } from 'jotai/vanilla/utils'
+import { atom } from 'jotai/vanilla'
```
chef-jojo, line 3: No need to change in other places that import jotai?

apps/web/src/hooks/useLockedEndNotification.tsx
```diff
@@ -10,7 +10,6 @@ import atomWithStorageWithErrorCatch from 'utils/atomWithStorageWithErrorCatch'
 import { useUserCakeLockStatus } from './useUserCakeLockStatus'
 
 const storage = createJSONStorage(() => sessionStorage)
```
chef-jojo, line 12: I feel like making a copy of atom with storage try catch is a bit smelly.

Actually if we really need to try catch on localstorage we can use this pattern to assign a try catch setitem verion of localstorage in createjsonstorage

chef-jojo (approved):

PR #6377, opened 2023-03-24
chore(UserMenu): SID should not omit text

Woody810:
Before:
image

After:
image


Woody810:
> Imo we should omit it because of privacy purposes

I think the reason for the omission of the display here is because the address is too long. However, if using the SID, this problem would not exist, and other websites display the full text as well.

chefilip (approved):

PR #6383, opened 2023-03-26
perf: Make common pairs unique

memoyil:
Everytime a new update from pairs makes a new entry to swr cache which increases the memory consumption


chefilip:
Could you show the before and after impact of the change?

chefjackson:
Close for now as it's stale

PR #6387, opened 2023-03-27
chore(Profile): Modify ChangeProfilePicFlow margin

Woody810:
When there is no CAKE in the wallet, the Enable flow for ChangeProfilePic will appear, causing users to mistakenly think that they need to Enable first before being able to RemoveProfilePic. Therefore, the marginTop of RemoveProfilePic has been increased.

Before:
image

After
image

### DIFF
--- a/apps/web/src/views/Profile/components/EditProfileModal/StartView.tsx
+++ b/apps/web/src/views/Profile/components/EditProfileModal/StartView.tsx
@@ -125,7 +125,7 @@ const StartPage: React.FC<React.PropsWithChildren<StartPageProps>> = ({ goToAppr
               {t('Change Profile Pic')}
             </Button>
           )}
-          <DangerOutline width="100%" onClick={goToRemove}>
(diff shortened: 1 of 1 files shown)


apps/web/src/views/Profile/components/EditProfileModal/StartView.tsx
```diff
@@ -114,7 +118,7 @@ const StartPage: React.FC<React.PropsWithChildren<StartPageProps>> = ({ goToAppr
                 onConfirm={needsApproval === true ? goToApprove : goToChange}
                 confirmLabel={t('Change Profile Pic')}
               />
-            </Flex>
+            </ChangeProfilePicWrapper>
```
ChefMomota, line 121: Don't think that we need the border bottom line

Woody810: Fixed

chefilip (approved):

PR #6391, opened 2023-03-27
chore: Use pcs rpc for farm booster contract when reading

memoyil:


apps/web/src/views/Farms/components/YieldBooster/hooks/useBoostMultiplier.ts
```diff
@@ -90,7 +90,7 @@ async function getMultiplierFromMC({ pid, proxyAddress, masterChefContract }): P
 }
 
 export default function useBoostMultiplier({ pid, boosterState, proxyAddress }): number {
-  const farmBoosterContract = useBCakeFarmBoosterContract()
+  const farmBoosterContract = useBCakeFarmBoosterContract(false)
```
chefilip, line 93: let's make `false` as default instead of `true`

PR #6480, opened 2023-04-04
fix(blog): Build

chef-jojo:


chef-jojo:
I think the blog doesn't use `src` dir

chefilip (approved):

PR #6486, opened 2023-04-04
fix(Liquidity): Fix history's table placement on mobile

Woody810:
Before:
image

After:
image


chefilip:
For long numbers, the UI doesn't work well. That's why we use row in mobile

PR #6515, opened 2023-04-06
chore(Liquidity): Default display fee tier options

Woody810:
I think that displaying the fee tier options by default is more user-friendly for operation.


chefilip:
It would be flaky if user use URL with feeAmount selected

chefilip:
Will close it for now. If user requests more, we could apply it.

PR #6689, opened 2023-04-22
chore: CommonToken list scrollbar appear

Woody810:
copilot:all

Before:

https://user-images.githubusercontent.com/109973128/233766165-99ec9f6d-2bb9-4990-aadb-780b993701db.mov

After:
image


chef-jojo:
instead of wrap, which scarifying the height space there. You can try remove the padding left and right and make it visible on horizontal scroll 
CleanShot 2023-04-24 at 12 31 09@2x

PR #6739, opened 2023-04-26
chore: sort latest from v3 to v2 in farm

chefilip:
- Sort latest from v3 to v2 farms
- Add multi chain support in clicking CakePrice component


Chef-Yogi:
CleanShot 2023-04-26 at 14 48 24@2x
it seems we need to update the snapshot

PR #7160, opened 2023-06-18
chore: list TUSD farms

Chef-Yogi:
copilot:all


packages/tokens/src/56.ts
```diff
@@ -2618,4 +2618,12 @@ export const bscTokens = {
     'PLAY',
     'https://xcadnetwork.com/play',
   ),
+  tusd_new: new ERC20Token(
```
chefjackson, line 2621: why it's new? Is there an old one? Can we deprecate the old one?

Chef-Yogi: Yeah, BD just confirm that we can remove the old one, will update it later

Chef-Yogi: Yeah, BD just confirm that we can remove the old one, will update it later

ChefMomota: We should not remove it, because we still have old farm is using even if move to finished farm / pool.

ChefMomota (approved):

ChefMomota (approved):

ChefMomota (dismissed):

packages/tokens/src/56.ts
Chef-Yogi: @chefjackson could you help me to force merge this first? 
we can't pass the test because the tusd_new key
expected 'tusd_new' to be 'tusd' // Object.is equality

PR #7484, opened 2023-08-03
feat(TradingReward): Add farm config

ChefMomota:
copilot:all


ChefMomota:
Need add tag at `PairInfo` file before merge

Chef-Yogi (approved):

PR #7689, opened 2023-08-29
feat: integrate Wallchain's newer SDK

fish-sausage:


apps/web/src/views/Swap/V3Swap/hooks/useWallchain.ts
```diff
@@ -0,0 +1,235 @@
...
+    const includesPair = trade.routes.some(
+      (route) =>
+        (route.inputAmount.wrapped.currency.equals(WallchainPairs[0]) &&
+          route.outputAmount.wrapped.currency.equals(WallchainPairs[1])) ||
+        (route.inputAmount.wrapped.currency.equals(WallchainPairs[1]) &&
+          route.outputAmount.wrapped.currency.equals(WallchainPairs[0])),
+    )
+    if (includesPair) {
+      setStatus('pending')
+      wrappedLoadData(account, sdk, swapCalls)
```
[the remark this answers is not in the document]
fish-sausage: For data loading we want to ensure that response always matches the swapCalls. It means, it would be better to drop task with bottleneck (false negative), then end up with cached value than doesn't match current route (false positive). Though, api is called immediately during a swap call, in case of false positive we would have wrong approval address being checked.

apps/web/src/views/Swap/V3Swap/containers/SwapCommitButton.tsx
```diff
@@ -361,7 +386,13 @@ export const SwapCommitButton = memo(function SwapCommitButton({
...
         id="swap-button"
         width="100%"
-        disabled={!isValid || (priceImpactSeverity > 3 && !isExpertMode) || !!swapCallbackError || !approved}
+        disabled={
+          !isValid ||
+          (priceImpactSeverity > 3 && !isExpertMode) ||
+          !!swapCallbackError ||
+          !approved ||
+          statusWallchain === 'pending'
+        }
```
chef-jojo, line 395: seems like disable status is toggling too frequently on bsc due to the revalidate time

apps/web/src/views/Swap/V3Swap/hooks/useWallchain.ts
```diff
@@ -0,0 +1,241 @@
...
+            )
+
+            return {
+              address: data.to as `0x${string}`,
+              calldata: data.data as `0x${string}`,
+              value: data.value as `0x${string}`,
+              gas: suggestedGas,
+            }
+          } catch (e) {
+            return previousSwapCalls[0]
```
chef-jojo, line 231: set status not found on catch

PR #7925, opened 2023-09-21
feat: Third year anniversary perdiction leaderboard

ChefMomota:
copilot:all


apps/web/src/state/predictions/config.ts
```diff
@@ -10,3 +10,5 @@ export const FUTURE_ROUND_COUNT = 2
 
 export const ROUNDS_PER_PAGE = 200
 export const LEADERBOARD_MIN_ROUNDS_PLAYED = 10
+
+export const START_THIRD_YEAR_BLOCK_HEIGHT = 20991251
```
ChefMomota, line 14: Change it before merge

apps/web/src/state/predictions/index.ts
```diff
@@ -170,7 +171,7 @@ export const fetchNodeHistory = createAsyncThunk<
   const { bufferSeconds } = getState()
 
   // Turn the data from the node into a Bet object that comes from the graph
-  const bets: Bet[] = roundData.reduce((accum, round) => {
+  const bets: Bet[] = roundData.reduce((accum: any | PredictionsRoundsResponse, round: PredictionsRoundsResponse) => {
```
chefjackson, line 174: the type of `accum` should be `Bet[]`?

ChefMomota: I'm try to set it to `Bet[]` but still throw error

chefjackson: hmm what is the error exactly? Or let's simply set any to bypass

chefjackson (approved):

PR #7961, opened 2023-09-26
chore: Fix invalid import

chefjackson:
copilot:all


packages/uikit/src/shared.ts
```diff
@@ -1,4 +1,4 @@
-import { ChainId, ERC20Token } from "@pancakeswap/sdk";
+import { ChainId, ERC20Token } from "@pancakeswap/chains";
```
ChefJerry, line 1: there is no `ERC20Token` exported by chains pkg

ChefJerry (commented):
wrong import

PR #8074, opened 2023-10-09
Update Onyx Protocol (XCN) info

dissimilllate:


chefjackson:
The protocol should have rebranded

PR #8089, opened 2023-10-10
fix: Wrong token address being used for price query

simone1999:
seems like a tiny copy mistake. But can cause price charts not working when input and output tokens are only available on different subgraph instances (v2/v2/stable) for pricing.


chefjackson (approved):

chefjackson:
Thanks for the fix

PR #8115, opened 2023-10-12
fix: swap message on user reject fix

ChefBingbong:
### Summary
🛑🐛🛡️

Improved error handling for wallet connection rejection. Added a check for undefined properties in `isUserRejected` function in `apps/web/src/utils/sentry.ts`.

> _`isUserRejected`_
> _now checks for undefined_
> _a winter bug fix_

### Walkthrough
*  Add a condition to the `isUserRejected` function to handle errors caused by user rejecting wallet connection ([link](https://github.com/pancakeswap/pancake-frontend/pull/8115/files?diff=unified&w=0#diff-a3898994b9c897df1f61a5132f19305441a63c27468cf5903db687dea5cf7d70R48))

### DIFF
--- a/apps/web/src/utils/sentry.ts
+++ b/apps/web/src/utils/sentry.ts
@@ -1,5 +1,5 @@
 import { captureException } from '@sentry/nextjs'
-import { UserRejectedRequestError, UnknownRpcError } from 'viem'
+import { UnknownRpcError, UserRejectedRequestError } from 'viem'
 
 const assignError = (maybeError: any) => {
   if (typeof maybeError === 'string') {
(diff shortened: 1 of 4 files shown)


chefjackson (changes requested):
Let's switch to use `useDeprecatedSendTransaction` to prevent breaking changes. Then open a separate pr to do the migration

https://wagmi.sh/react/migration-guide#usesendtransaction-1

chefjackson (changes requested):
We should raise to wagmi to fix this issue
https://github.com/wagmi-dev/wagmi/blob/main/packages/core/src/actions/transactions/sendTransaction.ts#L101

PR #8197, opened 2023-10-24
fix: roi calc

ChefJerry:
### Summary
📈🔧🚀

Improved and optimized the APR calculator for V3 pools using `Tick` data.

> _Oh we're the savvy sailors of the V3 pools_
> _We use the `Tick` data to find the best tools_
> _We memoize our maths to make our code run fast_
> _And we calculate the APR with every contrast_

### Walkthrough
* Import `Tick` type from `@pancakeswap/v3-sdk` to create ticks array from data ([link](https://github.com/pancakeswap/pancake-frontend/pull/8197/files?diff=unified&w=0#diff-11372967482ecb262dc9c61cfcc4c28c7ac992e7a9bad352f6fce19f1940bdb1L10-R10))
* Memoize `allTicks` and `mostActiveLiquidity` variables to avoid unnecessary computations and improve performance ([link](https://github.com/pancakeswap/pancake-frontend/pull/8197/files?diff=unified&w=0#diff-11372967482ecb262dc9c61cfcc4c28c7ac992e7a9bad352f6fce19f1940bdb1R162-R172))
* Use `mostActiveLiquidity` instead of `pool?.liquidity` to calculate ROI with more accuracy based on current price and tick data ([link](https://github.com/pancakeswap/pancake-frontend/pull/8197/files?diff=unified&w=0#diff-11372967482ecb262dc9c61cfcc4c28c7ac992e7a9bad352f6fce19f1940bdb1L167-R178))

### DIFF
--- a/apps/web/src/utils/isPoolTickInRange.ts
+++ b/apps/web/src/utils/isPoolTickInRange.ts
@@ -1,6 +1,7 @@
 import { Pool } from '@pancakeswap/v3-sdk'
 
-const isPoolTickInRange = (pool: Pool, tickLower: number, tickUpper: number) => {
+const isPoolTickInRange = (pool: Pool | undefined, tickLower: number, tickUpper: number) => {
+  if (typeof pool === 'undefined') return false
   const below = pool && typeof tickLower === 'number' ? pool.tickCurrent < tickLower : undefined
   const above = pool && typeof tickUpper === 'number' ? pool.tickCurrent >= tickUpper : undefined
   return typeof below === 'boolean' && typeof above === 'boolean' ? !below && !above : false
--- a/apps/web/src/utils/isPoolTickInRange.ts
+++ b/apps/web/src/utils/isPoolTickInRange.ts
@@ -1,9 +1,9 @@
 import { Pool } from '@pancakeswap/v3-sdk'
 
-const isPoolTickInRange = (pool: Pool | undefined, tickLower: number, tickUpper: number) => {
+const isPoolTickInRange = (pool: Pool | undefined, tickLower: number | undefined, tickUpper: number | undefined) => {
   if (typeof pool === 'undefined') return false
-  const below = pool && typeof tickLower === 'number' ? pool.tickCurrent < tickLower : undefined
-  const above = pool && typeof tickUpper === 'number' ? pool.tickCurrent >= tickUpper : undefined
+  const below = typeof tickLower === 'number' ? pool.tickCurrent < tickLower : undefined
+  const above = typeof tickUpper === 'number' ? pool.tickCurrent >= tickUpper : undefined
   return typeof below === 'boolean' && typeof above === 'boolean' ? !below && !above : false
 }
 
--- a/apps/web/src/views/AddLiquidityV3/components/AprCalculator.tsx
+++ b/apps/web/src/views/AddLiquidityV3/components/AprCalculator.tsx
@@ -7,7 +7,7 @@ import {
   useAmountsByUsdValue,
 } from '@pancakeswap/widgets-internal/roi'
 import { BIG_ZERO } from '@pancakeswap/utils/bigNumber'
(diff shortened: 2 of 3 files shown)


apps/web/src/utils/isPoolTickInRange.ts
```diff
@@ -1,6 +1,7 @@
 import { Pool } from '@pancakeswap/v3-sdk'
 
-const isPoolTickInRange = (pool: Pool, tickLower: number, tickUpper: number) => {
+const isPoolTickInRange = (pool: Pool | undefined, tickLower: number, tickUpper: number) => {
+  if (typeof pool === 'undefined') return false
   const below = pool && typeof tickLower === 'number' ? pool.tickCurrent < tickLower : undefined
```
chefjackson, line 5: we can remove the pool type check if we do the check in early return
```suggestion
  const below = typeof tickLower === 'number' ? pool.tickCurrent < tickLower : undefined
```

apps/web/src/utils/isPoolTickInRange.ts
```diff
@@ -1,6 +1,7 @@
 import { Pool } from '@pancakeswap/v3-sdk'
 
-const isPoolTickInRange = (pool: Pool, tickLower: number, tickUpper: number) => {
+const isPoolTickInRange = (pool: Pool | undefined, tickLower: number, tickUpper: number) => {
```
chefjackson, line 3: Can we also add the type to `tickLower` and `tickUpper`?
```suggestion
const isPoolTickInRange = (pool: Pool | undefined, tickLower: number | undefined, tickUpper: number | undefined) => {
```

apps/web/src/views/AddLiquidityV3/components/AprCalculator.tsx
```diff
@@ -158,13 +158,24 @@ export function AprCalculator({
   const validAmountA = amountA || (inverted ? tokenAmount1 : tokenAmount0) || (inverted ? aprAmountB : aprAmountA)
   const validAmountB = amountB || (inverted ? tokenAmount0 : tokenAmount1) || (inverted ? aprAmountA : aprAmountB)
   const [amount0, amount1] = inverted ? [validAmountB, validAmountA] : [validAmountA, validAmountB]
-  const inRange = isPoolTickInRange(pool, tickLower, tickUpper)
+  const inRange = isPoolTickInRange(pool ?? undefined, tickLower ?? 0, tickUpper ?? 0)
```
chefjackson, line 161: The reason to accept `undefined` as tick is because `0` tick is valid and has meaning so it would be better to use undefined instead
```suggestion
  const inRange = isPoolTickInRange(pool ?? undefined, tickLower ?? undefined, tickUpper ?? undefined)
```

apps/web/src/views/AddLiquidityV3/components/AprCalculator.tsx
```diff
@@ -158,13 +158,24 @@ export function AprCalculator({
...
+    () => allTicks && sqrtRatioX96 && FeeCalculator.getLiquidityFromSqrtRatioX96(allTicks, sqrtRatioX96),
+    [allTicks, sqrtRatioX96],
+  )
   const { apr } = useRoi({
     tickLower,
     tickUpper,
     sqrtRatioX96,
     fee: feeAmount,
-    mostActiveLiquidity: pool?.liquidity,
+    mostActiveLiquidity,
```
chefjackson, line 178: hmm the calculated value should match with `pool.liquidity` if the data is correct. I'm thinking probably the ticks data is not up-to-date. What about we switch to use pool liquidity inside the roi calculator?

ChefJerry: use `pool?.liquidity` is better, on my test it near the same inside and outside

chefjackson (approved):

chefjackson (approved):

PR #8202, opened 2023-10-24
feat: process all pairs with wallchain

fish-sausage:
Don't check if pair is allowed.


chefjackson:
Close as handled in https://github.com/pancakeswap/pancake-frontend/pull/8335

PR #8256, opened 2023-10-30
fix(Pools): Fix tooltip placement

Woody810:
### Summary
🆕🛠️🎨

Added `placement` option to `useTooltip` hook for better tooltip positioning in `VaultStakeButtonGroup.tsx`. This improves the UI of the vault stake buttons.

> _`useTooltip` hook_
> _adds `placement` option_
> _bottom-start, better_

### Walkthrough
*  Add `placement` option to `useTooltip` hook to position the tooltip at the bottom-start of the target element ([link](https://github.com/pancakeswap/pancake-frontend/pull/8256/files?diff=unified&w=0#diff-6729929fd0f56837495b9ef70e6c95f7ebc3b9df080dcc86ce987bf6b02a5bf8L18-R18))

image

### DIFF
--- a/apps/web/src/views/Pools/components/Vault/VaultStakeButtonGroup.tsx
+++ b/apps/web/src/views/Pools/components/Vault/VaultStakeButtonGroup.tsx
@@ -15,7 +15,7 @@ export const VaultStakeButtonGroup = ({
         'Flexible staking offers flexibility for staking/unstaking whenever you want. Locked staking offers higher APY as well as other benefits.',
       )}
     </Box>,
-    {},
+    { placement: 'bottom-start' },
   )
   return (
     <Box width="100%">


ChefMomota:
> > > If we all placement set to auto, what would happen?
> > 
> > 
> > Current Prod is fine, if we no set any `placement`
> 
> @Woody810 @ChefMomota If auto is fine, i think it is better to remove all custom placements. wdyt?

For me, I don't think we should change anything, is better follow what designer need.

Woody810:
> > > > If we all placement set to auto, what would happen?
> > > 
> > > 
> > > Current Prod is fine, if we no set any `placement`
> > 
> > 
> > @Woody810 @ChefMomota If auto is fine, i think it is better to remove all custom placements. wdyt?
> 
> For me, I don't think we should change anything, is better follow what designer need.

Agree. But, I think this PR still need to adjust.

chefjackson (approved):
Updating the position of this tooltip does no harm for now IMO. We can have separate discussion on whether it's a good practice for us to set the position manually

PR #8372, opened 2023-11-15
feat: gauges pool page

Chef-Yogi:
copilot:all


packages/uikit/src/components/Message/Message.tsx
```diff
@@ -63,7 +64,7 @@ const Message: React.FC<React.PropsWithChildren<MessageProps>> = ({
     <MessageContext.Provider value={providerValue}>
       <MessageContainer variant={variant} {...props}>
         <Flex>
-          <Box mr="12px">{icon ?? <Icon color={variants[variant].borderColor} width="24px" />}</Box>
+          {!hideIcon && <Box mr="12px">{icon ?? <Icon color={variants[variant].borderColor} width="24px" />}</Box>}
```
chefjackson, line 67: why do we need this? The icon won't display if `icon` is nullish?

chefjackson:
Let's resolve those ts errors in the newly added files and merge to the feature branch first. Would be hard to review two separate on going branches

packages/uikit/src/components/Message/Message.tsx
Chef-Yogi: if we don't pass the customize icon as props, it will render <Icon color={variants[variant].borderColor} width="24px" />
still, so need this to control should hode icon or not

PR #8483, opened 2023-11-29
feat: gauges api

ChefJerry:
copilot:all


packages/gauges/src/getAllGauges.ts
```diff
@@ -0,0 +1,36 @@
...
+  testnet?: boolean
+}
+
+export const getAllGauges = async (
+  options: getAllGaugesOptions = {
+    testnet: false,
+  },
+): Promise<Gauge[]> => {
+  const { testnet } = options
+  const client = testnet ? testnetPublicClient : publicClient
```
chefjackson, line 17: I'd recommend to pass in the client as part of the function parameters. It's hard to maintain the nodes in every package

chefjackson (dismissed):

chefjackson (approved):

PR #8604, opened 2023-12-18
fix: ACE apr

ChefMomota:
copilot:all


chef-jojo:
Why not using pirce helprr?

chefjackson:
> Why not using pirce helprr?

We need to add support for v3 farm price helper. Currently the price helper only do support v2 farms

chefjackson (approved):
It's a temp fix. We need to deal with the token price issue without hardcode

PR #8783, opened 2024-01-10
feat: Early abort

chefjackson:
---

### Detailed summary
- Added support for abort control in the code
- Updated dependencies and versions of packages
- Imported `AbortControl` and `isAbortError` from `@pancakeswap/utils/abortControl`
- Added `signal` parameter to various functions for abort control

> The following files were skipped due to too many changes: `packages/smart-router/evm/v3-router/providers/onChainQuoteProvider.ts`, `apps/web/src/utils/viem.ts`, `apps/web/src/quote-worker.ts`, `packages/multicall/src/multicall.ts`, `apps/web/src/hooks/useBestAMMTrade.ts`


packages/utils/viem/isAbortError.ts
```diff
@@ -0,0 +1,3 @@
+export function isAbortError(e: any) {
+  return e instanceof Error && e.message.startsWith('The request took too long to respond.')
```
chef-jojo, line 2: e?.name === 'AbortError'

packages/utils/viem/isAbortError.ts
```diff
@@ -0,0 +1,3 @@
+export function isAbortError(e: any) {
```
chef-jojo, line 1: is not the same as the isAbortError from abortControl?

chef-jojo: ah it's viem error
how about we use `TimeoutError` from viem? It might be breaking if they change the message
but it will require utils to have viem deps

chefjackson: Yes it's from viem. At first I tried with `TimeoutError` but actually viem is overriding the error type. For example if you make the call by `contract.simulate` then it'll throw an Error with name "ContractExecutionError", instead of the `TimeoutError`

They're not handling the abort error very well

chef-jojo: I guess you need to `walk()` the error if they wrapped the error in the details.
can we change the name, seems confusing if we have 2 same function name here

chefjackson: ok interesting. Lemme check how their `walk` works

chef-jojo (approved):

PR #8799, opened 2024-01-11
fix: Add missing translations

Woody810:
---

### Detailed summary
- Updated code to use translation functions (`t()`) instead of hardcoded strings for UI text.
- Added translations for "Wrap" and "Unwrap" in the `translations.json` file.
- Replaced hardcoded strings with translations in various components (`ChartLabel.tsx`, `SwapCommitButton.tsx`, `MMCommitButton.tsx`, `VeCakeButton.tsx`).
- Added translations for various UI text in the `translations.json` file.
- Updated the `desc` property in the `BENEFITS` object to use translations instead of hardcoded strings in `BenefitCard.tsx`.


apps/web/src/__tests__/translations.test.ts
```diff
@@ -96,6 +96,18 @@ const whitelist = [
...
+  'Boost rewards on your favorite farms',
+  'Claim additional incentives from other protocols',
+  'Boost your farm earnings',
+  'Supports multi-chain boosts',
+  'Use your veCAKE to vote on important governance proposals',
+  'Participate in IFO public sales',
+  'Supports multi-chain IFOs',
+  'Boost your winning in Trading Rewards campaigns. Regardless which chain you are trading on.',
+  'Boost your earnings in fixed staking.',
+  'More to come...',
```
ChefMomota, line 110: Why add in here?

Woody810: Because it dynamically invokes the translation function, the unit test identifies it as an unused string.

image
image

chefjackson: hmm we need to figure out a better way to extract dynamic translations

ChefMomota: Change it to [<Trans>123</Trans>].
U can check the code in `IfoQuestions` config.

Exp:
`description: [<Trans>The CAKE from the participation fee will be burnt as part of the weekly token burn.</Trans>],`

chefjackson: If we use switch to use `ReactNode` for desc, just be aware that we need to use a different value for the `key` prop of li. Key only accepts `string | number`

Woody810: Fixed

apps/web/src/views/CakeStaking/components/BenefitCard.tsx
```diff
@@ -134,7 +137,7 @@ export const BenefitCard: React.FC<{
           <Text lineHeight="130%">
             <StyleUl>
               {info.desc.map((item) => (
-                <li key={item}>{t(item)}</li>
+                <li key={item?.toString()}>{item}</li>
```
chefjackson, line 140: What's the result of `toString()` call on a `ReactNode`?

Woody810: `Object`😂
Fixed.

ChefMomota (approved):

PR #8985, opened 2024-02-03
chore(PositionManager): only show data startTime > current time

Woody810:
---

### Detailed summary
- Added `usePositionInfo` hook to fetch position information
- Filtered out positions that have not started yet


apps/web/src/views/PositionManagers/containers/VaultCards.tsx
```diff
@@ -47,6 +48,15 @@ export const VaultCards = memo(function VaultCards() {
       }
       return true
     })
+    .filter((d) => {
+      // eslint-disable-next-line react-hooks/rules-of-hooks
```
Chef-Yogi, line 52: In this case, we should batch fetch the position Info and filter here.

PR #8992, opened 2024-02-05
feat: approve & swap flow refactor

ChefJerry:
---

## PR-Codex overview
This PR focuses on adding support for the Permit2 protocol in the Swap feature. 

### Detailed summary
- Added `usePermit2Allowance` hook to handle Permit2 allowances.
- Added `usePermit2Details` hook to fetch Permit2 details.
- Added `usePermit2Requires` hook to determine if Permit2 approval is required.
- Updated `useWritePermit` hook to generate Permit2 signatures.
- Updated `useSwapCallback` hook to handle Permit2 signatures and added support for the Wallchain feature.
- Updated `useApproveRequires` hook to handle Permit2 approvals.

> The following files were skipped due to too many changes: `apps/web/src/views/Swap/MMLinkPools/hooks/useApprove.ts`, `apps/web/src/hooks/usePermit2.ts`, `apps/web/src/views/Swap/MMLinkPools/hooks/useSwapCallArguments.ts`, `apps/web/src/views/Swap/V3Swap/containers/ConfirmSwapModal.tsx`, `apps/web/src/views/Swap/MMLinkPools/hooks/useMMConfirmModalState.ts`, `apps/web/src/views/Swap/MMLinkPools/components/MMCommitButton.tsx`, `apps/web/src/views/Swap/V3Swap/containers/SwapCommitButton.tsx`, `apps/web/src/views/Swap/V3Swap/hooks/useConfirmModalState.ts`


apps/web/src/hooks/usePermit2.ts
```diff
@@ -0,0 +1,121 @@
...
+  const approve = useCallback(async () => {
+    setIsApproving(true)
+    try {
+      const result = await approveCallback()
+      setIsApproving(false)
+      return result
+    } catch (error) {
+      setIsApproving(false)
+      throw error
+    }
```
chefjackson, line 87: If we're doing the same thing maybe we can move to `finally`?
```suggestion
    try {
      const result = await approveCallback()
      return result
    } catch (error) {
      throw error
    } finally {
      setIsApproving(false)
    }
```

chefjackson: One more idea is to use `useQuery` instead of writing ourselves

apps/web/src/hooks/usePermit2Requires.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    return !!allowance && allowance.greaterThan(0) && allowance.lessThan(amount)
+  }, [allowance, amount])
+
+  const requireApprove = useMemo((): boolean => {
+    return !!amount && !!allowance && allowance.lessThan(amount)
+  }, [allowance, amount])
+
+  const requirePermit = useMemo((): boolean => {
+    return (amount && permitAmount?.lessThan(amount)) || (Boolean(expiration) && now >= expiration)
```
chefjackson, line 35: I'd suggest we do this check in real time. When we need to check if the user requires permit, check with the current timestamp. There may be a mismatch between block timestamp and current timestamp, especially on chain with longer block time

chefjackson: For instance

```typescript
if (isPermitRequired()) {
  // Then start the permit flow
}
```

chefjackson: we may even add a buffer time for that, in case the permit is going to expire in like 30s, and the user probably won't have enough time to do the actual swap

apps/web/src/hooks/usePermit2.ts
ChefJerry: I'm not sure if `useQuery` is suitable for this case, as there used for callback, not use query result directly

apps/web/src/views/Swap/V3Swap/hooks/useSendSwapTransaction.ts
```diff
@@ -245,3 +245,11 @@ export default function useSendSwapTransaction(
     addTransaction,
   ])
 }
+
+export const userRejectedError = (error: unknown): boolean => {
+  return (
+    error instanceof UserRejectedRequestError ||
+    error instanceof TransactionRejectedError ||
+    (typeof error !== 'string' && isUserRejected(error))
```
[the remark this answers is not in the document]
ChefJerry: I think we should, just strange why we create a redundant `TransactionRejectedError`

apps/web/src/hooks/usePermit2Requires.ts
ChefJerry: I think use the current `requirePermit` is enough, but is a great idea to add a buffer.

because, the steps will only create on user click the swap button, thats say if will check `requireXXX` the same time
just make sure have enough time window for user to execute all actions, so make a enough `buffer` is a good solution

apps/web/src/hooks/usePermit2.ts
chefjackson: Maybe can try combining enabled with refetch?
```typescript
const { refetch, isLoading } = useQuery({
  enabled: false,
})
async function asyncCall() {
  refetch()
}
```

PR #9081, opened 2024-02-16
fix: Cake staking week placeholder

memoyil:
Instead of 0.0 which week staking not allowing decimal values, it should show integer

---

## PR-Codex overview
This PR introduces a new component `LockImageElement` to display a lock image in the `WeekInput` component in the `LockWeeksForm.tsx` file.

### Detailed summary
- Added a new `LockImageElement` component to display a lock image
- Updated the `WeekInput` component to use the `LockImageElement` component
- Replaced direct JSX element with `LockImageElement` component in `WeekInput`

### DIFF
--- a/apps/web/src/views/CakeStaking/components/LockWeeksForm.tsx
+++ b/apps/web/src/views/CakeStaking/components/LockWeeksForm.tsx
@@ -71,16 +71,20 @@ const WeekInput: React.FC<{
     [onInput],
   )
 
-  const appendComponent = (
-    <Box width={40} mr={12}>
-      <Image src="/images/cake-staking/lock.png" height={37} width={34} />
-    </Box>
+  const appendComponent = useMemo(
+    () => (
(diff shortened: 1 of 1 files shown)


apps/web/src/views/CakeStaking/components/LockWeeksForm.tsx
```diff
@@ -71,15 +71,19 @@ const WeekInput: React.FC<{
     [onInput],
   )
 
-  const appendComponent = (
-    <Box width={40} mr={12}>
-      <Image src="/images/cake-staking/lock.png" height={37} width={34} />
-    </Box>
+  const appendComponent = useMemo(
```
chefjackson, line 74: I think it would be unnecessary to add `useMemo` to jsx? If we really need to do the perf optimization I'd recommend to add `React.memo` to the underlying component

chefjackson (approved):

PR #9117, opened 2024-02-21
fetch pools prices in parallel rather than sequential

simone1999:
Making pool page initial load faster by loading all pool prices in parallel rather than sequential. Especially when the Llama pricing API is used this massively reduces pools load time for users with high latency connections like common on mobile.
There was a race condition in the past where .map with the async callback was not correctly awaited, but this was fixed by introducing a sequential for loop rather than awaiting it correctly.

### DIFF
--- a/apps/web/src/state/pools/index.ts
+++ b/apps/web/src/state/pools/index.ts
@@ -200,7 +200,7 @@ export const fetchPoolsPublicDataAsync = (chainId: number) => async (dispatch, g
 
     const liveData: any[] = []
 
-    for (const pool of poolsConfig) {
+    await Promise.all(poolsConfig.map(async pool => {
       const timeLimit = timeLimitsSousIdMap[pool.sousId]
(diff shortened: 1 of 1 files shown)


apps/web/src/state/pools/index.ts
```diff
@@ -200,7 +200,7 @@ export const fetchPoolsPublicDataAsync = (chainId: number) => async (dispatch, g
 
     const liveData: any[] = []
 
-    for (const pool of poolsConfig) {
+    await Promise.all(poolsConfig.map(async pool => {
```
chef-jojo, line 203: Thanks, can you make it return to live pool to remain the order

chef-jojo (changes requested):

ChefJerry:
hi @simone1999 , thanks for your contributing!

close this PR because have no response for a long time, feel free to reopen it if necessary

PR #9295, opened 2024-03-13
fix: TS errors under src/views/TradingReward

chefjackson:
---

## PR-Codex overview
The focus of this PR is to explicitly define `null` as `any` in multiple files and update type assertions in components for improved type safety.

### Detailed summary
- Defined `null` as `any` in `tradingReward` configs for various farms
- Updated type assertions to `any` in `MyRank.tsx` and `index.tsx`
- Added fallback value `'0x'` for `account` in `useClaimAllReward.tsx`

### DIFF
--- a/apps/web/src/views/TradingReward/components/Leaderboard/MyRank.tsx
+++ b/apps/web/src/views/TradingReward/components/Leaderboard/MyRank.tsx
@@ -53,12 +53,12 @@ const MyRank: React.FC<React.PropsWithChildren<MyRankProps>> = ({ campaignId }) 
                   </Td>
                 </tr>
               ) : (
-                <DesktopResult key={rank.rank} rank={rank} />
+                <DesktopResult key={rank.rank} rank={rank as any} />
               )}
             </tbody>
           </Table>
         ) : (
-          <MobileResult rank={rank} isMyRank />
+          <MobileResult rank={rank as any} isMyRank />
         )}
       </Card>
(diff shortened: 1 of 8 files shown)


apps/web/src/views/TradingReward/components/Leaderboard/MyRank.tsx
```diff
@@ -53,12 +53,12 @@ const MyRank: React.FC<React.PropsWithChildren<MyRankProps>> = ({ campaignId })
...
                 </tr>
               ) : (
-                <DesktopResult key={rank.rank} rank={rank} />
+                <DesktopResult key={rank.rank} rank={rank as any} />
               )}
             </tbody>
           </Table>
         ) : (
-          <MobileResult rank={rank} isMyRank />
+          <MobileResult rank={rank as any} isMyRank />
```
ChefBingbong, line 61: how come we need any here. could we not just cast the type of rabk on to the component props?

ChefBingbong (approved):

apps/web/src/views/TradingReward/components/Leaderboard/MyRank.tsx
chefjackson: if we're using type casting then it doesn't make any difference whether we cast to the correct type or not, since the type is still broken. Will just cast to any to bypass the check first. Clear path for the wagmi migration

PR #9381, opened 2024-03-20
fix: max tick check on v3 sdk

Chef-Yogi:
---

## PR-Codex overview
This PR updates the `useV3DerivedInfo` hook in the V3 app with new imports, type changes, and adjustments to tick space limits.

### Detailed summary
- Added imports for `BIG_INT_ZERO` and `Bound`
- Updated type imports for `PoolState`
- Adjusted tick space limits for lower and upper bounds
- Modified tick calculation logic for fee tiers

### DIFF
--- a/apps/web/src/hooks/v3/useV3DerivedInfo.ts
+++ b/apps/web/src/hooks/v3/useV3DerivedInfo.ts
@@ -1,28 +1,28 @@
+import { useTranslation } from '@pancakeswap/localization'
 import { Currency, CurrencyAmount, Price, Token } from '@pancakeswap/swap-sdk-core'
 import {
-  encodeSqrtRatioX96,
   FeeAmount,
-  nearestUsableTick,
   Pool,
   Position,
-  priceToClosestTick,
   TICK_SPACINGS,
   TickMath,
+  encodeSqrtRatioX96,
+  nearestUsableTick,
+  priceToClosestTick,
 } from '@pancakeswap/v3-sdk'
+import { BIG_INT_ZERO } from 'config/constants/exchange'
+import { Bound } from 'config/constants/types'
 import { ReactNode, useMemo } from 'react'
 import { Field } from 'state/mint/actions'
-import tryParseCurrencyAmount from 'utils/tryParseCurrencyAmount'
-import { BIG_INT_ZERO } from 'config/constants/exchange'
 import { useCurrencyBalances } from 'state/wallet/hooks'
-import { useTranslation } from '@pancakeswap/localization'
-import { Bound } from 'config/constants/types'
+import tryParseCurrencyAmount from 'utils/tryParseCurrencyAmount'
 import { MintState } from 'views/AddLiquidityV3/formViews/V3FormView/form/reducer'
 
 import { useAccount } from 'wagmi'
-import { tryParseTick } from './utils'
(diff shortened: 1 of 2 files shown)


packages/v3-sdk/src/utils/tickMath.ts
```diff
@@ -80,7 +80,7 @@ export abstract class TickMath {
    * @param sqrtRatioX96 the sqrt ratio as a Q64.96 for which to compute the tick
    */
   public static getTickAtSqrtRatio(sqrtRatioX96: bigint): number {
-    invariant(sqrtRatioX96 >= TickMath.MIN_SQRT_RATIO && sqrtRatioX96 < TickMath.MAX_SQRT_RATIO, 'SQRT_RATIO')
+    invariant(sqrtRatioX96 >= TickMath.MIN_SQRT_RATIO && sqrtRatioX96 <= TickMath.MAX_SQRT_RATIO, 'SQRT_RATIO')
```
chefjackson, line 83: I think we should update the price limit instead. When we set price range from 0 -> max, the sqrtRatioLimit should be [TickMath.MIN_SQRT_RATIO + 1n, TickMath.MAX_SQRT_RATIO - 1n]

chefjackson (changes requested):

packages/v3-sdk/src/utils/tickMath.ts
Chef-Yogi: Done

chef-jojo: Should only adjust when meet the max or min

apps/web/src/hooks/v3/useV3DerivedInfo.ts
```diff
@@ -176,6 +176,8 @@ export default function useV3DerivedInfo(
           : (invertPrice && typeof rightRangeTypedValue === 'boolean') ||
             (!invertPrice && typeof leftRangeTypedValue === 'boolean')
           ? tickSpaceLimits[Bound.LOWER]
+            ? tickSpaceLimits[Bound.LOWER] + 1
```
chefjackson, line 179: for lower tick should be fine I think. Let's remove plus one here

apps/web/src/hooks/v3/useV3DerivedInfo.ts
```diff
@@ -185,6 +187,8 @@ export default function useV3DerivedInfo(
           : (!invertPrice && typeof rightRangeTypedValue === 'boolean') ||
             (invertPrice && typeof leftRangeTypedValue === 'boolean')
           ? tickSpaceLimits[Bound.UPPER]
+            ? tickSpaceLimits[Bound.UPPER] - 1
```
chefjackson, line 190: We need to consider the tick spacing. For example if tick spacing is 10, then the tickSpaceLimit here won't be 887272, but 887270. In this case there's no need to minus 1

Chef-Yogi: Done

apps/web/src/hooks/v3/useV3DerivedInfo.ts
Chef-Yogi: Done

chefjackson (approved):

PR #9389, opened 2024-03-21
feat: Phishing Banner add timer check

ChefMomota:
---

## PR-Codex overview
This PR updates the logic for displaying a phishing banner based on time intervals and environment.

### Detailed summary
- Updated logic to show phishing banner based on timestamp comparison
- Changed time interval comparison to check if current time is greater than last time stored


chefjackson (commented):
is it the same as our previous implementation?

ChefMomota:
> is it the same as our previous implementation?

Yes, different is when clicked close, after one day. If current timestamp > then that time close + 1day. It will show again

chefjackson (approved):

PR #9399, opened 2024-03-22
chore: add error name and mm log

ChefJerry:
---

## PR-Codex overview
The focus of this PR is to enhance error logging and handling in the `Swap` and `MMLinkPools` views.

### Detailed summary
- Added `target` and `errorName` fields in error logging for `Swap` view
- Imported `logger` from `utils/datadog` in `MMLinkPools` view
- Enhanced error logging in `useSendMMTransaction` function

### DIFF
--- a/apps/web/src/views/Swap/V3Swap/hooks/useSendSwapTransaction.ts
+++ b/apps/web/src/views/Swap/V3Swap/hooks/useSendSwapTransaction.ts
@@ -14,14 +14,7 @@ import { basisPointsToPercent } from 'utils/exchange'
 import { logSwap, logTx } from 'utils/log'
 import { isUserRejected } from 'utils/sentry'
 import { transactionErrorToUserReadableMessage } from 'utils/transactionErrorToUserReadableMessage'
-import {
-  Address,
-  EstimateGasExecutionError,
-  Hex,
-  TransactionExecutionError,
-  UserRejectedRequestError,
-  hexToBigInt,
-} from 'viem'
+import { Address, Hex, TransactionExecutionError, UserRejectedRequestError, hexToBigInt } from 'viem'
 import { useSendTransaction } from 'wagmi'
 import { SendTransactionResult } from 'wagmi/actions'
 
@@ -221,7 +214,7 @@ export default function useSendSwapTransaction(
             } else {
               // otherwise, the error was unexpected and we need to convey that
               logger.warn(
-                error instanceof EstimateGasExecutionError ? 'Estimate failed' : 'Swap failed',
+                'Swap failed',
                 {
                   chainId,
                   input: trade.inputAmount.currency,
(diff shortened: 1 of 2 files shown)


apps/web/src/views/Swap/V3Swap/hooks/useSendSwapTransaction.ts
```diff
@@ -214,15 +221,20 @@ export default function useSendSwapTransaction(
             } else {
               // otherwise, the error was unexpected and we need to convey that
               logger.warn(
-                'Swap failed',
+                error instanceof EstimateGasExecutionError ? 'Estimate failed' : 'Swap failed',
```
chefjackson, line 224: Estimate should be done here?
https://github.com/pancakeswap/pancake-frontend/pull/9399/files#diff-83218ea7d05848674a6cd69042376956fd3dfcb9f694f65f7954867778363739R100

chefjackson: And let's add an identifier to link the estimation and execution

chefjackson: so we can locate both gas estimation and swap execution for a single tx. Let's log the estimated gas cost and the actual gas specified for the execution.

ChefJerry: > Estimate should be done here? [#9399 (files)](https://github.com/pancakeswap/pancake-frontend/pull/9399/files#diff-83218ea7d05848674a6cd69042376956fd3dfcb9f694f65f7954867778363739R100)

you are right, firstly I think `sendTransactionAsync` will do `prepareTransactionRequest` internal, which will estimate again, but I forgot we have passed `gas` parameter.

ChefJerry: So, I think we should no need to link estimate and execution logs? as only will be one of them failed.

or you want to log normal actions?

ChefJerry: > so we can locate both gas estimation and swap execution for a single tx. Let's log the estimated gas cost and the actual gas specified for the execution.

for `TransactionExecutionError`, the error log already include `gas` info

chefjackson: the error log only shows how much gas actually used right? I'm suggesting linking the two logs is because we can compare the how much gas is estimated and how much gas we actually used. To see if we need to increase the margin for some cases

ChefJerry: https://github.com/wevm/viem/blob/main/src/actions/public/multicall.ts#L190-L196

the gas in error log is request params

chefjackson: ah ok. Then should be good

chefjackson (approved):

PR #9441, opened 2024-03-28
fix: Footer config

ChefMomota:
---

## PR-Codex overview
The focus of this PR is to update footer links in the Menu component to redirect to PancakeSwap's website. 

### Detailed summary
- Updated footer links in the Menu component to PancakeSwap's website URLs
- Added new footer links for various sections like Ecosystem, Business, Developers, Support, and About

### DIFF
--- a/packages/uikit/src/widgets/Menu/components/footerConfig.ts
+++ b/packages/uikit/src/widgets/Menu/components/footerConfig.ts
@@ -7,11 +7,11 @@ export const footerLinks: (t: ContextApi["t"]) => FooterLinkType[] = (t) => [
     items: [
       {
         label: t("Trade"),
-        href: "/swap",
+        href: "https://pancakeswap.finance/swap",
       },
       {
         label: t("Earn"),
-        href: "/farms",
+        href: "https://pancakeswap.finance/farms",
       },
       {
         label: t("Game"),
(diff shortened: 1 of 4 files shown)


packages/uikit/src/widgets/Menu/components/footerConfig.ts
```diff
@@ -7,19 +7,19 @@ export const footerLinks: (t: ContextApi["t"]) => FooterLinkType[] = (t) => [
     items: [
       {
         label: t("Trade"),
-        href: "/swap",
+        href: "https://pancakeswap.finance/swap",
```
chefjackson, line 10: We should still use relative path for swap and farms?

ChefMomota: This will cause 404 at `https://pancakeswap.games/`
at Aptos `/nfts` & `/v4` will also return 404

chefjackson: Can we have different footer links configured for different apps? We can have some common links but I think essentially they should have different links. Otherwise the swap and farm link will be redirecting to the evm site for aptos

ChefMomota: Ok. Will make a config only for Aptos. Other will maintain with this

ChefMomota: done

chefjackson (approved):

PR #9495, opened 2024-04-04
chore(PositionMgr): Strategy name is not using translation

Woody810:
---

## PR-Codex overview
The focus of this PR is to remove the translation function (`t()`) calls and directly return the strings in the `switch` statement in `strategy.ts`. 

### Detailed summary
- Removed `t()` calls and directly returned strings in the `switch` statement in `strategy.ts`
- Updated translations in `translations.json` to match the direct string returns in `strategy.ts`


chefjackson:
We don't need translations for position manager strategies, as it's hard to clearly deliver the meaning after translating

Woody810:
> We don't need translations for position manager strategies, as it's hard to clearly deliver the meaning after translating

Agree, so I remove the translation of strategies on this PR🤔

PR #9599, opened 2024-04-15
fix: useActiveChainId should always return support chainId

ChefJerry:
---

## PR-Codex overview
This PR updates the `useActiveChainId` hook to handle network mismatch and unsupported chains efficiently.

### Detailed summary
- Added `isWrongNetwork` to check for unsupported or mismatched chains
- Updated `chainId` logic to fallback to BSC if not supported

### DIFF
--- a/apps/web/src/hooks/useActiveChainId.ts
+++ b/apps/web/src/hooks/useActiveChainId.ts
@@ -57,7 +57,7 @@ export const useActiveChainId = () => {
   )
 
   return {
-    chainId: isWrongNetwork ? ChainId.BSC : chainId,
+    chainId: chainId && isChainSupported(chainId) ? chainId : ChainId.BSC,
     isWrongNetwork,
     isNotMatched,
   }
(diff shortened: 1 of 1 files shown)


apps/web/src/hooks/useActiveChainId.ts
```diff
@@ -51,13 +51,14 @@ export const useActiveChainId = () => {
...
+    [wagmiChainId, isNotMatched],
+  )
 
   return {
-    chainId,
-    isWrongNetwork: useMemo(
-      () => Boolean(((wagmiChainId && !isChainSupported(wagmiChainId)) ?? false) || isNotMatched),
-      [wagmiChainId, isNotMatched],
-    ),
+    chainId: isWrongNetwork ? ChainId.BSC : chainId,
```
chefjackson, line 60: what if it's just not matching? For example the `wagmiChainId` is `eth` while the `localChainId` is `base`

chefjackson: btw in what cases this might be an issue?

ChefJerry:
```ts
  const { chainId } = useActiveChainId()
  const provider = usePublicClient({ chainId })
```

when `chainId` is not supported, `usePublicClient` will throw `ChainNotConfiguredError: Chain not configured`

ChefJerry: > what if it's just not matching? For example the `wagmiChainId` is `eth` while the `localChainId` is `base`

refactor to check if supported only

ChefMomota (approved):

PR #9655, opened 2024-04-22
feat: Run only when block number changes

memoyil:
---

## PR-Codex overview
This PR refactors multiple hooks in `wagmi` package to use a custom `useDidMountEffect` hook instead of `useEffect` for better performance.

### Detailed summary
- Refactored `useFeeData`, `useReadContracts`, `useReadContract`, and `useBalance` hooks
- Replaced `useEffect` with `useDidMountEffect` for conditional logic
- Improved performance by reducing unnecessary re-renders

### DIFF
--- a/packages/wagmi/src/hooks/useDidMountEffect.ts
+++ b/packages/wagmi/src/hooks/useDidMountEffect.ts
@@ -1,0 +1,22 @@
+import { useEffect, useRef } from 'react'
+
+const useDidMountEffect = (func: () => void, checkObject: any) => {
+  const isFirstRun = useRef(true)
+  const prevParams = useRef(checkObject)
+
+  useEffect(() => {
+    if (isFirstRun.current) {
+      isFirstRun.current = false
+      return
+    }
+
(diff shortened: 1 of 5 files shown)


packages/wagmi/src/hooks/useDidMountEffect.ts
```diff
@@ -0,0 +1,22 @@
+import { useEffect, useRef } from 'react'
+
+const useDidMountEffect = (func: () => void, checkObject: any) => {
```
chefjackson, line 3: how about we just copy the hooks from react use, and move to our hooks pkg?

https://github.com/streamich/react-use/blob/master/src/useUpdateEffect.ts

chefjackson: yup I think the link I posted is what we want. I skips the first mount and only get called on update

chefjackson: hmm ok i see your point

chefjackson (approved):

PR #9805, opened 2024-05-14
fix: remove undefined key of stablecoin

ChefJerry:
---

## PR-Codex overview
This PR updates the `MAJOR_DEFINED_STABLE_COINS` object structure and refactors the `useFeeSaved` hook to handle stable coins more efficiently.

### Detailed summary
- Restructured `MAJOR_DEFINED_STABLE_COINS` object with a chain-specific array
- Updated `useFeeSaved` to handle stable coins based on the chain ID
- Improved handling of zero fee scenarios

### DIFF
--- a/apps/web/src/views/Swap/hooks/useFeeSaved.ts
+++ b/apps/web/src/views/Swap/hooks/useFeeSaved.ts
@@ -1,11 +1,13 @@
-import { ChainId, Currency, CurrencyAmount } from '@pancakeswap/sdk'
+import { ChainId, Currency, CurrencyAmount, ERC20Token } from '@pancakeswap/sdk'
 import { DAI, USDC, USDT } from '@pancakeswap/tokens'
 import { parseNumberToFraction } from '@pancakeswap/utils/formatFractions'
 import { BigNumber } from 'bignumber.js'
 import { useCurrencyUsdPrice } from 'hooks/useCurrencyUsdPrice'
 import { useMemo } from 'react'
 
-const MAJOR_DEFINED_STABLE_COINS = {
+const MAJOR_DEFINED_STABLE_COINS: {
+  [chainId in ChainId]?: Array<ERC20Token | undefined>
+} = {
   [ChainId.ETHEREUM]: [USDC[ChainId.ETHEREUM], DAI[ChainId.ETHEREUM], USDT[ChainId.ETHEREUM]],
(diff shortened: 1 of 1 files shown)


apps/web/src/views/Swap/hooks/useFeeSaved.ts
```diff
@@ -8,20 +8,9 @@ import { useMemo } from 'react'
 const MAJOR_DEFINED_STABLE_COINS = {
```
chefjackson, line 8: Let's add explicit type definition here. So we can identify the error at build time

chefjackson: also can remove the optional mark in line 24

ChefJerry: explicit mark type as `ERC20Token | undefined` and keep optional mark, as `Token[chainId]` will not throw any warning during static check or build time.

apps/web/src/views/Swap/hooks/useFeeSaved.ts
```diff
@@ -1,11 +1,13 @@
...
+import { ChainId, Currency, CurrencyAmount, ERC20Token } from '@pancakeswap/sdk'
 import { DAI, USDC, USDT } from '@pancakeswap/tokens'
 import { parseNumberToFraction } from '@pancakeswap/utils/formatFractions'
 import { BigNumber } from 'bignumber.js'
 import { useCurrencyUsdPrice } from 'hooks/useCurrencyUsdPrice'
 import { useMemo } from 'react'
 
-const MAJOR_DEFINED_STABLE_COINS = {
+const MAJOR_DEFINED_STABLE_COINS: {
+  [chainId in ChainId]?: Array<ERC20Token | undefined>
```
chefjackson, line 9: The list is defined by us right? So we can safely assert it's not undefined here. If any of them is undefined, the error throws on build time

```suggestion
  [chainId in ChainId]?: Array<ERC20Token>
```

apps/web/src/views/Swap/hooks/useFeeSaved.ts
chefjackson: hmm ok that's weird. Let's keep it for now. Will double check later on

chefjackson (approved):

apps/web/src/views/Swap/hooks/useFeeSaved.ts
ChefJerry: can check this [commit](https://github.com/pancakeswap/pancake-frontend/commit/06276308934e120329aa98e03500d7ad42b50a1d)

build passed with no error

PR #9808, opened 2024-05-14
fix: Type

ChefMomota:


packages/widgets-internal/package.json
```diff
@@ -43,6 +43,7 @@
     "next": "14.2.3",
     "polished": "^4.2.2",
     "react": "^18.2.0",
-    "styled-components": "6.0.7"
+    "styled-components": "6.0.7",
+    "@types/d3": "^7.4.0"
```
chefjackson, line 47: did we use `d3` in internal widget?

ChefMomota: Here is the error
https://vercel.com/pancakeswap/gamification/33YiNFyZUpxzFY1K2RZUN3dHqmt1

chefjackson (approved):

PR #9818, opened 2024-05-15
fix: add support for fiat currency prefixes on NumberDisplay component

ChefBingbong:
---

## PR-Codex overview
This PR focuses on restructuring the `NumberDisplay` components and adding new features like fiat number display formatting.

### Detailed summary
- Deleted `NumberDisplay/index.tsx`
- Added `FiatNumberDisplay.tsx`
- Added `BaseNumberDisplay.tsx`
- Updated `NumberDisplay.tsx`
- Updated `formatNumber` and `formatFiatNumber` functions
- Updated `packages/widgets-internal/index.ts` to include new components


ChefBingbong:
@chefjackson @chef-jojo . could have flag fiatCurrOps wihich if defined will assume the user wants to render a fiat price value by defaulting the valueDisplay to use number.toLocalString with currency option passed by default and symbol from whatever fiatCurrencySymbol is passed. 

decided to keep in the flexibility of allowing the prefix and suffix in this case in the scenario for unique number displays. I like the modularity of the mobile dislay types, but i think something like this (together with a few suggestions ) can work for now

chefjackson (changes requested):
Let's hold until the design has unified the number display format for fiat. And we should build fiat format on top of the general format

packages/widgets-internal/components/FiatNumberDisplay/index.tsx
chefjackson, line 1: This component should be built on top of the basic NumberDisplay component

chefjackson (changes requested):
Let's build `FiatNumberDisplay` on top of `NumberDisplay`

packages/utils/formatNumber.ts
```diff
@@ -58,10 +71,57 @@ export function formatNumber(
   return `${limitIndicator}${valueToDisplay}`
 }
 
+export function formatFiatNumber({
```
chefjackson, line 74: It's better that we treat `formatFiatNumber` as formatting numbers with a specific rule which suits fiat display. And `formatNumber` is the base function that allow us to specify the rule and `formatFiatNumber` should call `formatNumber` and specify the rule it needs

ChefBingbong: resolved these. lets chat monday

PR #9845, opened 2024-05-21
feat: zyfi integration

thechefpenguin:
---

## PR-Codex overview
This PR introduces gas token selection functionality using Zyfi Paymaster, including new hooks and components.

### Detailed summary
- Added `useGasToken` hook and `GasTokenSelector` component
- Implemented gas token selection in swap details
- Created API handler for gas token transactions
- Defined default gas tokens and paymaster URLs

> The following files were skipped due to too many changes: `apps/web/src/views/Swap/V3Swap/hooks/useSendSwapTransaction.ts`, `apps/web/src/hooks/usePaymaster.ts`, `apps/web/src/views/Home/components/Banners/PaymasterBanner.tsx`, `apps/web/src/hooks/useApproveCallback.ts`, `apps/web/src/views/Swap/V3Swap/components/SwapModalFooter.tsx`, `apps/web/src/components/Paymaster/GasTokenSelector.tsx`


chefjackson:
please move static assets to our assets repo

apps/web/src/hooks/useApproveCallback.ts
```diff
@@ -8,12 +8,13 @@ import { useHasPendingApproval, useTransactionAdder } from 'state/transactions/h
 import { calculateGasMargin } from 'utils'
 import { getViemErrorMessage } from 'utils/errors'
 import { isUserRejected, logError } from 'utils/sentry'
-import { Address } from 'viem'
+import { Address, Hex, SendTransactionReturnType, encodeFunctionData, parseAbi } from 'viem'
 import { useAccount } from 'wagmi'
-import useGelatoLimitOrdersLib from './limitOrders/useGelatoLimitOrdersLib'
+import { usePaymaster } from 'views/Swap/V3Swap/components/Paymaster/hooks/usePaymaster'
```
chefjackson, line 13: the dependency relationship should be the other way around. We can import `src/hooks` in `views`, but we shouldn't import `views` in `src/hooks`.

If we're importing `views` inside a common hook, it means that the imported module should also be a top-level common module

apps/web/src/views/Swap/V3Swap/components/Paymaster/GasTokenModal.tsx
chefjackson, line 1: I think we can move this modal component to `src/components/Paymaster`, because it actually has nothing to do with swap specifically. We could also put this gas token selection modal into global settings

apps/web/src/views/Swap/V3Swap/components/Paymaster/config/config.ts
```diff
@@ -0,0 +1,30 @@
+import { zkSync } from 'viem/zksync'
```
chefjackson, line 1: If we need chain id, let's use `ChainId` from `@pancakeswap/chains` instead

apps/web/src/views/Swap/V3Swap/components/SwapModalFooter.tsx
```diff
@@ -226,6 +244,37 @@ export const SwapModalFooter = memo(function SwapModalFooter({
...
+                {(feeToken && feeToken.symbol && feeToken.symbol.length > 10
+                  ? `${feeToken.symbol.slice(0, 4)}...${feeToken.symbol.slice(
+                      feeToken.symbol.length - 5,
+                      feeToken.symbol.length,
+                    )}`
+                  : feeToken?.symbol) || 'ETH'}
+              </Text>
+
+              <div style={{ position: 'relative' }}>
+                <StyledLogo
```
chefjackson, line 264: There's a `CurrencyLogo` component in internal-widgets package

apps/web/src/views/Swap/components/AdvancedSwapDetails.tsx
```diff
@@ -178,6 +186,38 @@ export const TradeSummary = memo(function TradeSummary({
...
+                      {t('Please refer to the transaction confirmation on your wallet for the final gas fee.')}
+                    </Text>
+                  </>
+                }
+                ml="4px"
+                placement="top"
+              />
+            </RowFixed>
+
+            {gasTokenSelector}
```
chefjackson, line 217: I was thinking the whole row including the label is part of the gas token selector

apps/web/src/views/Swap/V3Swap/components/Paymaster/GasTokenModal.tsx
```diff
@@ -0,0 +1,320 @@
...
+              </Flex>
+            </StyledFooterText>
+          </StyledModalBody>
+        </StyledModalContainer>
+      </ModalV2>
+    </>
+  )
+}
+
+export default GasTokenModal
```
chefjackson, line 320: Let's use named export if possible

apps/web/src/views/Swap/V3Swap/components/Paymaster/config/config.ts
```diff
@@ -0,0 +1,30 @@
...
+  name: 'Ether',
+  symbol: 'ETH',
+  logoURI: 'https://assets.pancakeswap.finance/web/native/324.png',
+  isNative: true,
+  isToken: false,
+  chainId: zkSync.id,
+}
+
+export const paymasterTokens: PaymasterToken[] = [
+  { ...zksyncTokens.wbtc, logoURI: 'https://tokens.pancakeswap.finance/images/symbol/wbtc.png' },
```
chefjackson, line 18: We can reuse the `CurrencyLogo` component from internal widgets so I don't think we need this config anymore?

chefjackson: sorry I mean the config of the `logoURI`

apps/web/src/views/Swap/V3Swap/components/Paymaster/types/index.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  chainId: number
+  isNative: boolean
+  isToken: boolean
+
+  name?: string
+  logoURI?: string
+}
+
+export type PaymasterToken = BasePaymasterToken & Partial<Currency & ZyfiResponse>
```
chefjackson, line 17: Can we just use `Currency` type from `sdk-core`?

chefjackson: I think it's better not mix up the zyfi response with the currency/token type

apps/web/src/views/Swap/V3Swap/components/SwapModalFooter.tsx
thechefpenguin: I had tried the local CurrencyLogo component but it didn't produce most images. The one from internal-widgets works 👍🏻

chefjackson: yea we should remove the one in web app

apps/web/src/config/paymaster.ts
```diff
@@ -0,0 +1,78 @@
...
+export const DEFAULT_PAYMASTER_TOKEN: PaymasterToken = {
+  address: NATIVE_CURRENCY_ADDRESS,
+  decimals: 18,
+  name: 'Ether',
+  symbol: 'ETH',
+  logoURI: 'https://assets.pancakeswap.finance/web/native/324.png',
+  isNative: true,
+  isToken: false,
+  chainId: ChainId.ZKSYNC,
+}
```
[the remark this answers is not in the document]
thechefpenguin: Thanks

apps/web/src/config/paymaster.ts
```diff
@@ -0,0 +1,27 @@
...
+  zksyncTokens.grai,
+  zksyncTokens.tes,
+].map((token) => ({ ...token, discount: '-20%' }))
+
+export interface BasePaymasterToken {
+  address: Address
+  discount?: string
+}
+
+export type PaymasterToken = BasePaymasterToken & Partial<Currency | ERC20Token>
```
chefjackson, line 27: How about
```suggestion
export type PaymasterToken = Currency & {
  discount?: string
}
```

thechefpenguin: Causes type issues everywhere. 

The core of the issue I think is that all tokens are ERC20 except the Native one, and the types differ in that way. Like native doesn't have address and isNative & isToken are fixed boolean fields for both `ERC20Token` and `Native`.

Maybe I can use `null` value to mean Native. A getter for feeToken could return Native if null.

chefjackson: For native token you can always `wrapped` to get the wrapped Token instance

apps/web/src/hooks/useApproveCallback.ts
```diff
@@ -30,11 +30,18 @@ export function useApproveCallback(
...
+
+    /**
+     * Use paymaster if available.
+     * Enable only if Gas Token Selector is present on the interface.
+     */
+    enablePaymaster?: boolean
   } = {
     addToTransaction: true,
+    enablePaymaster: false,
   },
```
chefjackson, line 42: better set the default value one by one
```suggestion
  {
    enablePaymaster = false,
  }: {
    addToTransaction?: boolean
    targetAmount?: bigint
    enablePaymaster?: boolean
  } = {}
```

apps/web/src/hooks/usePermit2.ts
```diff
@@ -54,7 +63,10 @@ export const usePermit2 = (
   const [isApproving, setIsApproving] = useState(false)
 
   const writePermit = useWritePermit(amount?.currency, spender, permit2Details?.nonce)
-  const { approveNoCheck, revokeNoCheck } = useApproveCallback(amount, approveTarget)
+  const { approveNoCheck, revokeNoCheck } = useApproveCallback(amount, approveTarget, {
+    addToTransaction: true,
```
chefjackson, line 67: let's keep it the same as before to use the default value

apps/web/src/hooks/usePermit2.ts
```diff
@@ -36,6 +36,15 @@ type UsePermit2ReturnType = Permit2HookState & Permit2HookCallback
...
   spender: Address | undefined,
+  options: {
+    /**
+     * Use paymaster if available.
+     * Enable only if Gas Token Selector is present on the interface.
+     */
+    enablePaymaster?: boolean
+  } = {
+    enablePaymaster: false,
+  },
```
chefjackson, line 47:
```suggestion
  {
      enablePaymaster = false,
    }: {
    /**
     * Use paymaster if available.
     * Enable only if Gas Token Selector is present on the interface.
     */
    enablePaymaster?: boolean
  } = {},
```

apps/web/src/state/paymaster/atoms.ts
chefjackson, line 1: Suggest we have a `hooks/useGasToken.ts` file and move the state management there. From the design perspective, it's better to keep the usage more intuitive and simple. 

It would be easier for us to migrate from one library to another. Imagine if we don't use jotai anymore in the future, we don't need to update the code in component as long as we keep the interface of `useGasToken` hook untouched. What we need to do is simply update the code of the hook to use another storage lib

apps/web/src/components/Paymaster/GasTokenModal.tsx
```diff
@@ -0,0 +1,335 @@
...
+              </Flex>
+            </StyledFooterText>
+          </StyledModalBody>
+        </StyledModalContainer>
+      </ModalV2>
+    </>
+  )
+}
+
+export default GasTokenModal
```
chefjackson, line 335: let's use named export

apps/web/src/hooks/usePaymaster.ts
```diff
@@ -0,0 +1,126 @@
...
+          from: account,
+          to: call.address,
+          value: call.value,
+          data: call.calldata,
+        },
+      }),
+    })
+
+    if (response.ok) {
+      const txResponse: any = await response.json()
```
chefjackson, line 70: please add correct typing for the response

apps/web/src/hooks/usePaymaster.ts
```diff
@@ -0,0 +1,126 @@
...
+  }, [feeToken])
+
+  async function sendPaymasterTransaction(
+    call: SwapCall & {
+      gas?: string | bigint | undefined
+    },
+    account: Address,
+  ) {
+    if (!feeToken.isToken)
+      return Promise.reject(new Error('Selected gas token is not an ERC20 token. Unsupported by Paymaster.'))
```
chefjackson, line 50:
```suggestion
      throw new Error('Selected gas token is not an ERC20 token. Unsupported by Paymaster.')
```

apps/web/src/hooks/usePaymaster.ts
```diff
@@ -0,0 +1,126 @@
...
+          to: call.address,
+          value: call.value,
+          data: call.calldata,
+        },
+      }),
+    })
+
+    if (response.ok) {
+      const txResponse: any = await response.json()
+      console.debug('debug txResponse', txResponse)
```
chefjackson, line 71: let's remove debug logs

apps/web/src/hooks/usePaymaster.ts
```diff
@@ -0,0 +1,126 @@
...
+        txData: {
+          from: account,
+          to: call.address,
+          value: call.value,
+          data: call.calldata,
+        },
+      }),
+    })
+
+    if (response.ok) {
```
chefjackson, line 69: personally i'd prefer to deal with whatever simpler first, so there will be fewer nested brackets
```typescript
if (!response.ok) {
  throw Error
}
// then proceed with the normal flow
```

apps/web/src/state/paymaster/atoms.ts
thechefpenguin: That's true. good thing for modular design. I'll change it

apps/web/src/components/Paymaster/GasTokenModal.tsx
```diff
@@ -0,0 +1,375 @@
...
+  const inputTokenAddress = useMemo(() => trade && trade.inputAmount.currency.wrapped.address, [trade])
+
+  // Reset fee token if account changes, connects or disconnects
+  useEffect(() => {
+    return watchAccount(config, {
+      onChange() {
+        setGasToken(DEFAULT_PAYMASTER_TOKEN)
+      },
+    })
+  }, [config])
```
chefjackson, line 155:
```suggestion
  }, [config, setGasToken])
```

apps/web/src/hooks/useApproveCallback.ts
```diff
@@ -27,20 +28,29 @@ export function useApproveCallback(
...
 ) {
```
chefjackson, line 43: please set the default value of each filed separately. If we set the default options as a whole, when some of the options are omitted, they'll become undefined by default

chefjackson: consider the case where
```typescript
useApproveCallback(amountToApprove, spender)
useApproveCallback(amountToApprove, spender, {
  enablePaymaster: false,
})
```

apps/web/src/hooks/useApproveCallback.ts
```diff
@@ -140,22 +150,47 @@ export function useApproveCallback(
...
+        const calldata = encodeFunctionData({
+          abi: parseAbi(['function approve(address spender, uint256 amount) public returns (bool)']),
+          functionName: 'approve',
+          args: [spender as Address, finalAmount],
+        })
+
+        const call = {
+          address: tokenContract.address,
+          gas: estimatedGas,
+          value: '0x0' as Hex,
```
chefjackson, line 166: can we omit this field?

apps/web/src/hooks/useApproveCallback.ts
```diff
@@ -140,22 +150,47 @@ export function useApproveCallback(
...
+          calldata,
+        }
+
+        sendTxResult = sendPaymasterTransaction(call, account as Address)
```
chefjackson, line 170: Please do not use type assertion if possible. If the value is possibly undefined, then let's make sure it's not, or we need to accept undefined for `sendPaymasterTransaction`

apps/web/src/hooks/usePaymaster.ts
```diff
@@ -0,0 +1,100 @@
...
+    const txResponse: ZyfiResponse = await response.json()
+
+    const newTx = {
+      account,
+      to: txResponse.txData.to,
+      value: txResponse.txData.value && !isZero(txResponse.txData.value) ? hexToBigInt(txResponse.txData.value) : 0n,
+      chainId: ChainId.ZKSYNC,
+      gas: BigInt(txResponse.gasLimit),
+      maxFeePerGas: BigInt(txResponse.txData.maxFeePerGas),
+      maxPriorityFeePerGas: BigInt(0),
```
chefjackson, line 78: Is this field necessary?

apps/web/src/hooks/usePaymaster.ts
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    if (walletClient) {
+      // Extend Viem's zkSync utils
+      const client: any = walletClient.extend(eip712WalletActions() as any)
+      const hash = await client.sendTransaction(newTx)
+      return hash
+    }
+
+    throw new Error('Failed to execute paymaster transaction')
```
chefjackson, line 92:
```suggestion
    if (!walletClient) {
      throw new Error('Failed to execute paymaster transaction')
    }
    
    // Extend Viem's zkSync utils
    const client: any = walletClient.extend(eip712WalletActions() as any)
    return client.sendTransaction(newTx)
```

apps/web/src/views/Swap/components/AdvancedSwapDetails.tsx
```diff
@@ -178,6 +184,8 @@ export const TradeSummary = memo(function TradeSummary({
           <Text fontSize="14px">{`${formatAmount(realizedLPFee, 4)} ${inputAmount?.currency?.symbol}`}</Text>
         </RowBetween>
       )}
+
+      {isPaymasterAvailable && gasTokenSelector}
```
chefjackson, line 188: Can we do this condition check on the upper level? The `AdvancedSwapDetails` component doesn't have to know the context. It'll show the gas token selector if the props is given

apps/web/src/components/Paymaster/GasTokenModal.tsx
```diff
@@ -0,0 +1,375 @@
...
+type Trade = TradeEssentialForPriceBreakdown &
+  Pick<SmartRouterTrade<TradeType>, 'tradeType'> & {
+    routes: RouteDisplayEssentials[]
+  }
+
+interface GasTokenModalProps {
+  trade?: Trade | null
+}
+
+export const GasTokenModal = ({ trade }: GasTokenModalProps) => {
```
chefjackson, line 130: Shall we rename? This component is not only about modal but also the selector with label

thechefpenguin: `GasTokenSelector`?
Both button and modal are part of selection, so will cover both naming cases

apps/web/src/hooks/usePaymaster.ts
thechefpenguin: We should keep the fields. They are as per [Zyfi Docs](https://docs.zyfi.org/integration-guide/quick-start/erc20-paymaster#step-3-execute-the-transaction)

chefjackson: Could you point me to the place where we need `maxPriorityFeePerGas` this field?

apps/web/src/hooks/useApproveCallback.ts
```diff
@@ -26,21 +27,26 @@ export enum ApprovalState {
...
   spender?: string,
-  options: {
-    addToTransaction
-    targetAmount?: bigint
-  } = {
-    addToTransaction: true,
-  },
+  {
+    addToTransaction = true,
+    targetAmount = MaxUint256,
```
chefjackson, line 32: let's leave it default as undefined

apps/web/src/hooks/usePermit2.ts
chefjackson: please update to the same as what we did with `apps/web/src/hooks/useApproveCallback.ts`

apps/web/src/hooks/useApproveCallback.ts
```diff
@@ -26,21 +27,26 @@ export enum ApprovalState {
...
   spender?: string,
-  options: {
-    addToTransaction
-    targetAmount?: bigint
-  } = {
-    addToTransaction: true,
-  },
+  {
+    addToTransaction = true,
+    targetAmount = undefined as bigint | undefined,
```
chefjackson, line 32:
```suggestion
    targetAmount,
```

chefjackson:
```typescript
{
    addToTransaction = true,
    targetAmount,
}: {
    addToTransaction?: boolean;
    targetAmount?: bigint
} = {}
```

chefjackson (approved):

apps/web/src/hooks/usePaymaster.ts
```diff
@@ -22,14 +22,17 @@ export const usePaymaster = () => {
...
   const [gasToken] = useGasToken()
 
   /**
    * Check if the Paymaster for zkSync is available
    */
   const isPaymasterAvailable = useMemo(() => {
-    return chain && chain.chainId === ChainId.ZKSYNC
-  }, [chain])
+    // Disable connections via WalletConnect until it is fully supported
+    return chain && chain.chainId === ChainId.ZKSYNC && (!connector || connector.type !== 'walletConnect')
```
chefjackson, line 34:
```suggestion
    return chain && chain.chainId === ChainId.ZKSYNC && connector?.type !== 'walletConnect'
```

chefjackson:
@thechefpenguin please resolve the conflict

apps/web/public/images/swap/zyfi-logo.png
chefjackson, line 1: Can we move to assets repo?

thechefpenguin: Done, removed from here

chefjackson (approved):

PR #9865, opened 2024-05-23
perf: Reduce graphql requests in farms page

memoyil:
---

## PR-Codex overview
This PR focuses on restructuring import paths for helper functions in the Info section of the web app. It updates import paths to improve organization and consistency.

### Detailed summary
- Updated import paths for `infoDataHelpers` and `infoQueryHelpers` in multiple files to `utils` directory
- Added new interfaces `V3FarmWithoutStakedValue` and `V2FarmWithoutStakedValue` in `state/farms/types.ts`
- Updated import paths for `getChangeForPeriod` in various files to `utils`
- Updated import paths for `getPercentChange` in multiple files to `utils`
- Updated import paths for `useBCakeProxyContractAddress` in different components and hooks to match new directory structure

> The following files were skipped due to too many changes: `doc/Info.md`, `apps/web/src/views/Farms/components/FarmCard/V3/FarmV3ApyButton.tsx`, `apps/web/src/views/Farms/FarmsV3.tsx`, `apps/web/src/state/farms/hooks.ts`


apps/web/src/hooks/usePoolAvgInfo.ts
```diff
@@ -34,8 +35,13 @@ const defaultInfo: Info = {
 }
 
 export function usePoolAvgInfo({ address = '', numberOfDays = 7, chainId }: UsePoolAvgInfoParams) {
+  const router = useRouter()
+  const isInFarmsPage = router.pathname.includes('/farms')
```
chefjackson, line 39: we better not depend on the route path. Can we use param to explicitly control the behavior instead?

apps/web/src/state/farms/hooks.ts
```diff
@@ -63,6 +71,86 @@ export function useFarmV2PublicAPI() {
...
+    if (activeFarms) {
+      if (prevActiveFarms) {
+        if (activeFarms.length !== prevActiveFarms.length) {
+          refetch()
+        }
+      } else {
+        refetch()
+      }
+    }
+  }, [refetch, activeFarms, prevActiveFarms])
```
chefjackson, line 151: using length to compare doesn't look ideal. Can we use active farm addresses as query key?

chefjackson: is there any better way to deal with this? Tow concerns on my end
1. This approach is gonna have unexpected result if the new active farm list has the same length as the previous one

2. Not sure if separating the poll query and the consumer query is a good practice. It'll make it harder to reuse hooks because they have some dependency relationship covered inside the implementation. For example, if I don't check the code of `usePoolAvgInfo`, I won't be able to understand that `useFarmCache` is only valid when `usePollFarmsAvgInfo` is used. It creates a hidden dependency relationship between `usePoolAvgInfo` -> `usePollFarmsAvgInfo`

chefjackson: Can we have a hook called `useActiveFarms` and reuse that in the `usePoolAvgInfo` hook?

chefjackson: I'm thinking relationship like
```typescript
function useActiveFarms(chainId) {}
function useFarmsInfo(activeFarms) {}
function usePoolAvgInfo() {
  const activeFarms = useActiveFarms()
  const farmsInfo = useFarmsInfo(activeFarms)
}
```

chefjackson: by changing a lot you mean the reference of the farm list is changing? Can we just use the address of the farm list as query key instead?

apps/web/src/state/farms/hooks.ts
```diff
@@ -3,25 +3,33 @@ import { getFarmConfig } from '@pancakeswap/farms/constants'
...
 import { useBCakeProxyContractAddress } from 'views/Farms/hooks/useBCakeProxyContractAddress'
 
 import useAccountActiveChain from 'hooks/useAccountActiveChain'
-import {
-  fetchBCakeWrapperDataAsync,
-  fetchBCakeWrapperUserDataAsync,
-  fetchFarmsPublicDataAsync,
-  fetchFarmUserDataAsync,
-} from '.'
+import { V2FarmWithoutStakedValue, V3FarmWithoutStakedValue } from 'views/Farms/FarmsV3'
```
chefjackson, line 13: importing views inside global state hooks may cause circular reference. The dependency relationship should be `views/Farms/FarmsV3` -> `state/farms/hooks`, but not the other way around

chefjackson (approved):

PR #9892, opened 2024-05-30
fix: add theme customization for widget url gen

ChefBingbong:
---

## PR-Codex overview
The focus of this PR is to add a `WidgetTheme` enum and introduce a theme option in the `useOnRampSignature` hook.

### Detailed summary
- Added `WidgetTheme` enum with 'dark' and 'light' options
- Updated `useOnRampSignature` hook to include a `theme` parameter based on user's theme preference


apps/web/src/views/BuyCrypto/hooks/useOnRampSignature.ts
```diff
@@ -52,12 +63,13 @@ export const useOnRampSignature = <selectData = GetOnRampSignatureReturnType>(
...
         externalTransactionId,
         onRampUnit,
+        theme,
       },
     ]),
     enabled: Boolean(externalTransactionId && quote && walletAddress && getIsNetworkEnabled(chainId)),
     queryFn: async ({ queryKey }) => {
       // eslint-disable-next-line @typescript-eslint/no-shadow
-      const { quote, externalTransactionId, chainId, onRampUnit } = queryKey[1]
+      const { quote, externalTransactionId, chainId, onRampUnit, theme } = queryKey[1]
```
chefjackson, line 72: why do we read from queryKey instead of using these variables directly?

ChefBingbong: no i alsways just have done it this way ever since i wroekd on mobile as we do this in the mobile useQuery hooks quite often

chefjackson: is there any specific reason?

ChefBingbong: no dnt think so but i would have though since were usingt he query key to detcect change, then extracting the vars from the query key you can be certain their the most recent variables, so as to prevent an unnessecary refetch

apps/web/src/views/BuyCrypto/hooks/useOnRampSignature.ts
```diff
@@ -77,6 +86,7 @@ export const useOnRampSignature = <selectData = GetOnRampSignatureReturnType>(
           walletAddress,
           externalTransactionId,
           onRampUnit,
+          theme,
```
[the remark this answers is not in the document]
ChefBingbong: the theme is passed directly to the various post requests for generating the widget iframe url by the different providers

PR #9895, opened 2024-05-30
refactor: Use previous data when farms avg info query changes

memoyil:
---

## PR-Codex overview
The focus of this PR is to update the placeholder data function in the `useQuery` hook and optimize the manipulation of farm addresses.

### Detailed summary
- Updated `placeholderData` function to return an empty object if previous data is not available
- Optimized the manipulation of farm addresses using the `map` function
- Simplified the key manipulation process in the `mapKeys` function

### DIFF
--- a/apps/web/src/state/farms/hooks.ts
+++ b/apps/web/src/state/farms/hooks.ts
@@ -93,11 +93,7 @@ export const usePollFarmsAvgInfo = (activeFarms: (V3FarmWithoutStakedValue | V2F
         return {}
       }
 
-      const addresses = activeFarms
-        .map((farm) => farm.lpAddress)
-        .map((lpAddress) => {
-          return lpAddress.toLowerCase()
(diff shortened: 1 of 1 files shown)


apps/web/src/state/farms/hooks.ts
```diff
@@ -79,7 +79,12 @@ export const usePollFarmsAvgInfo = (activeFarms: (V3FarmWithoutStakedValue | V2F
 
   const { data } = useQuery({
     queryKey: ['farmsAvgInfo', chainId, activeFarmAddresses],
-    placeholderData: {},
+    placeholderData: (prev) => {
+      if (!prev) {
+        return {}
+      }
+      return prev
+    },
```
chefjackson, line 87: would it be the same if we set `keepPreviousData` to true?

chefjackson (approved):

PR #10129, opened 2024-07-03
fix: Unstake farms with vault

memoyil:
---

## PR-Codex overview
The focus of this PR is to update stake and unstake functions in the Farms feature to use gas prices specific to the Binance Smart Chain.

### Detailed summary
- Updated stake and unstake functions to use BNB gas prices for non-BSC vaults
- Added logic to handle BNB gas prices in stake and unstake functions

### DIFF
--- a/apps/web/src/views/Farms/hooks/getNonBscVaultFee.ts
+++ b/apps/web/src/views/Farms/hooks/getNonBscVaultFee.ts
@@ -2,6 +2,8 @@ import { BIG_ZERO } from '@pancakeswap/utils/bigNumber'
 import BigNumber from 'bignumber.js'
 import { getCrossFarmingSenderContract, getNonBscVaultContract } from 'utils/contractHelpers'
 import { Address } from 'viem'
+import { estimateFeesPerGas } from 'viem/actions'
+import { publicClient as getPublicClient } from 'utils/viem'
 
 export enum MessageTypes {
   Deposit = 0,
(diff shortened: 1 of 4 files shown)


apps/web/src/views/Farms/hooks/getNonBscVaultFee.ts
```diff
@@ -79,8 +81,14 @@ export const getNonBscVaultContractFee = async ({
         userAddress as Address,
         messageType,
       ])
+      const publicClient = getPublicClient({ chainId })
+      const { maxFeePerGas } = await estimateFeesPerGas(publicClient)
+      const evmGasPrice = maxFeePerGas?.toString()
+      if (!evmGasPrice) {
+        throw new Error('Missing evm gasPrice')
+      }
```
chefjackson, line 89: can we fix this in the `useGasPrice` hook?

chefjackson: ideally the hook should accept `chainId` as input and derive the correct gas price on target chain

chefjackson: My thought is to have a unified `useGasPrice` hook for all chains. Can we specify a list of chains that follow eip 1559? Like if the requested chain falls in this list then we'll calculate using maxFeePerGas, otherwise we'll fallback to the default way

chefjackson: ok I think we can use this approach for now. One more concern is that, it will break if we have v2 farms on other non bsc chain. Since we might not be able to get `maxFeePerGas`. 

Can we use `maxFeePerGas` as a fallback? We can use the default gasPrice if there's a valid input
```typescript
const evmGasPrice =  gasPrice ?? maxFeePerGas?.toString()
```

Secondly, just realized that the `useFeeData` hook from `@pancakeswap/wagmi` is broken. Could you help fix as well?

PR #10235, opened 2024-07-19
chore: new banner. waiting for links

ChefBingbong:

### DIFF
--- a/apps/web/src/views/Home/components/Banners/PancakeSwapAthleticsBanner.tsx
+++ b/apps/web/src/views/Home/components/Banners/PancakeSwapAthleticsBanner.tsx
@@ -1,0 +1,106 @@
+import { useTranslation } from '@pancakeswap/localization'
+import { Flex, useMatchBreakpoints } from '@pancakeswap/uikit'
+import {
+  BackgroundGraphic,
+  BannerActionContainer,
+  BannerContainer,
+  BannerDesc,
+  BannerGraphics,
+  BannerMain,
+  BannerTitle,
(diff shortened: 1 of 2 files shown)


apps/web/src/views/Home/components/Banners/PancakeSwapAthleticsBanner.tsx
```diff
@@ -0,0 +1,130 @@
...
+  lg: (t) => t('Complete tasks for special edition NFTs, merch, and USDT Prises!'),
+}
+
+export const AthleticsBanner = () => {
+  const { t } = useTranslation()
+  const { width } = useViewport()
+  const { isMobile } = useMatchBreakpoints()
+
+  const titleText = useMemo(() => {
+    if (width < 800) return Titles.sm(t)
```
chef-eric, line 83: can we use isMd/isLg from useMatchBreakpoints instead of the hard code width?

ChefBingbong: i need to use these widths because using the breakpoints from usemathc breakpoints cause title to break onto the next line too early on medium screen size and causes the button to overflow outside the bottom og the bottom of the banner

ChefBingbong: reason for this is cos the lg title for this banner is quite long compared to most others so it wraps early as quite long

chef-eric (approved):

PR #10253, opened 2024-07-23
fix: Ipfs links in voting

memoyil:
---

## PR-Codex overview
The focus of this PR is to integrate IPFS support into voting-related functionalities.

### Detailed summary
- Added `ipfs` property to `Vote` and `Proposal` types
- Updated links to use IPFS hash instead of ID
- Updated IPFS gateway URL
- Modified state management for proposal filters to include `proposalType`
- Optimized filtering of proposals by type and state

### DIFF
--- a/apps/web/src/views/Voting/config.ts
+++ b/apps/web/src/views/Voting/config.ts
@@ -7,7 +7,7 @@ export const ADMINS = [
   '0xa7551aBe0A066555cb5d859849426fB55543Ca25', // MUSTARD cakevote.eth
   '0x4a7b8400AE131feffC42E5Bf39401A094Ed497CA', // CHILI cakevote.eth
 ].map((address) => address.toLowerCase())
-export const IPFS_GATEWAY = 'https://gateway.ipfs.io/ipfs'
+export const IPFS_GATEWAY = 'https://snapshot.4everland.link/ipfs'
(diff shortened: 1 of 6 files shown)


apps/web/src/views/Voting/config.ts
```diff
@@ -7,7 +7,7 @@ export const ADMINS = [
   '0xa7551aBe0A066555cb5d859849426fB55543Ca25', // MUSTARD cakevote.eth
   '0x4a7b8400AE131feffC42E5Bf39401A094Ed497CA', // CHILI cakevote.eth
 ].map((address) => address.toLowerCase())
-export const IPFS_GATEWAY = 'https://gateway.ipfs.io/ipfs'
+export const IPFS_GATEWAY = 'https://snapshot.4everland.link/ipfs'
```
ChefJerry, line 10: why we need to update the ipfs gateway?

ChefJerry: https://github.com/snapshot-labs/snapshot/blob/4873bdf7b34608e7ab4a2fd3d8fbdaecdc4c88b4/src/helpers/utils.ts#L119-L123

I think the url change will not affect, it is IPFS after all, any gateway should works

ChefJerry (approved):

PR #10390, opened 2024-08-13
fix: universal farm issues

chef-eric:
---

## PR-Codex overview
The focus of this PR is to refactor and optimize token fetching and usage in various components and hooks.

### Detailed summary
- Removed unused token fetching hooks
- Updated token fetching logic in multiple components
- Improved token fetching efficiency and re-fetching behavior

> The following files were skipped due to too many changes: `apps/web/src/hooks/Tokens.ts`


apps/web/src/hooks/Tokens.ts
```diff
@@ -82,8 +82,12 @@ type TokenChainAddressMap<TChainId extends number = number> = {
...
  * Returns all tokens that are from active urls and user added tokens
  */
-export function useAllTokensByChainIds(chainIds: number[]): TokenChainAddressMap {
-  const tokenMap = useAtomValue(combinedTokenMapFromActiveUrlsAtom)
+export function useAllTokensByChainIds(
+  chainIds: number[],
+  allTokenMap_?: TokenAddressMap<ChainId>,
+): TokenChainAddressMap {
+  const allTokenMap = useAtomValue(combinedTokenMapFromActiveUrlsAtom)
+  const tokenMap = allTokenMap_ ?? allTokenMap
```
ChefBingbong, line 90: wouldnt there only be one `allTokenMap`, (being the current combined map of all our token lists on each chain). what would be a use case where we would need to pass an override to for this? if so wouldnt it no longer be the `allTokenMap` ?

do you think it would be better to have a seperate function specifically for the case where we want to override the combined tokenMap (since i imagine it wont be used heavily)? and use it only in the places where its needed instead of passing an override to the current `useAllTokensByChainIds` hook that isnt need too often throughout the web app like 

```ts
export function useTokensByChainIds(
  chainIds: number[],
  tokenMap: TokenAddressMap<ChainId>,
): TokenChainAddressMap {}

```
what do you think @chef-eric. Feel free to dismiss my opinion if im missing some context

chef-eric: the context is we want to reuse the logic in `useOfficialsAndUserAddedTokensByChainIds`, which like that:
```
export function useOfficialsAndUserAddedTokensByChainIds(chainIds: number[]): TokenChainAddressMap {
  const tokenMap = useAtomValue(combinedTokenMapFromOfficialsUrlsAtom)
  return useAllTokensByChainIds(chainIds, tokenMap)
}
```
yeah, your advise that to seperate a useTokensByChainIds function is better, I'll change it, thx

apps/web/src/hooks/Tokens.ts
```diff
@@ -106,6 +102,19 @@ export function useAllTokensByChainIds(chainIds: number[]): TokenChainAddressMap
   }, [userAddedTokenMap, tokenMap, chainIds])
 }
 
+/**
+ * Returns all tokens that are from active urls and user added tokens
+ */
+export function useAllTokensByChainIds(chainIds: number[]): TokenChainAddressMap {
```
chef-eric, line 108: @ChefBingbong updated as your advise

apps/web/src/hooks/Tokens.ts
ChefBingbong: ohh i see. ye cool soumds god

ChefBingbong (approved):

PR #10466, opened 2024-08-20
chore: fix decimal input for indonesian rupee

ChefBingbong:
---

## PR-Codex overview
This PR focuses on enhancing the BuyCrypto feature by implementing support for non-decimal fiat currencies.

### Detailed summary
- Updated `useDefaultAmount.ts` to handle non-decimal fiat currencies
- Added `NON_DECIMAL_FIAT_CURRENCIES` constant in `constants.ts`
- Modified `BuyCryptoForm.tsx` and `OnRampCurrencySelect/index.tsx` to incorporate `fiatCurrency` handling for non-decimal currencies

### DIFF
--- a/apps/web/src/views/BuyCrypto/components/OnRampCurrencySelect/index.tsx
+++ b/apps/web/src/views/BuyCrypto/components/OnRampCurrencySelect/index.tsx
@@ -110,7 +110,7 @@ export const BuyCryptoSelector = ({
             className="token-amount-input"
             value={inputLoading ? '' : value}
             onBlur={onInputBlur}
-            onKeyDown={(e) => blockDecimal(e)}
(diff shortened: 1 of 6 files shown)


ChefBingbong:
> @ChefBingbong Imho i think (since user still be able to paste decimal input) it is better to use/override numerical input’s pattern prop to prevent these too https://github.com/pancakeswap/pancake-frontend/blob/develop/packages/widgets-internal/swap/NumericalInput.tsx#L57
> 
> wdyt?

@memoyil . S0 the way ive written it is to not to change an amuch code as i can in the widgets package. thus for my the way i am doing the check is 

- if user is using IDR as their fiat currency enforce decimal place tying
- else continue as normal

thechefpenguin (approved):

PR #10655, opened 2024-09-10
feat: Cross-chain IFO

thechefpenguin:


apps/web/src/views/Ifos/components/IfoFoldableCard/IfoPoolCard/ClaimButton.tsx
```diff
@@ -33,7 +33,10 @@ const ClaimButton: React.FC<React.PropsWithChildren<Props>> = ({ poolId, ifoVers
       if (walletIfoData.version === 3) {
         return walletIfoData.contract.write.harvestPool([poolId === PoolIds.poolBasic ? 0 : 1], { account, chain })
       }
-      if (walletIfoData.version === 7 || walletIfoData.version === 8) {
+      if (walletIfoData.version === 7) {
```
ChefBingbong, line 36: not important but just a thought. should we make WalletInfoVersion an enum so making these equality checks is cleaner and use switch statement instead. switch statement works kinda like a map so can be better than if statement in some cases. could do something like

```ts
switch(walletIfoData.version) {
case: WalletInfoData.Version1:
   return .....
case WalletInfoData.version2:
   break;
...
...
```

apps/web/src/views/Ifos/components/IfoFoldableCard/IfoPoolCard/ContributeModal.tsx
```diff
@@ -40,9 +41,6 @@ interface Props {
 
 const multiplierValues = [0.1, 0.25, 0.5, 0.75, 1]
 
-// Default value for transaction setting, tweak based on BSC network congestion.
```
ChefBingbong, line 43: if this is needed for some config or default setting could we add to a constants file somewhere. maybe make a map of Default gas settings like 

```ts
export const DefaultGasPrices = {
   Slow: parseEther('5', 'gwei'),
   Medium: parseEther('10', 'gwei'),
   Fast: parseEther('15', 'gwei')
}


const gasPrice = DefaultGasPrices.Medium

```

apps/web/src/views/Ifos/components/IfoFoldableCard/IfoPoolCard/IfoCardTokens.tsx
```diff
@@ -228,7 +228,12 @@ const IfoCardTokens: React.FC<React.PropsWithChildren<IfoCardTokensProps>> = ({
       publicIfoData.status !== 'finished' &&
       hasProfile
     ) {
-      message = <ICakeTips ifoId={ifo.id} ifoChainId={ifo.chainId} ifoAddress={ifo.address} />
+      // If Cross-Chain IFO
+      // if (ifo.chainId !== ChainId.BSC) {
+      message = <CrossChainVeCakeTips ifoChainId={ifo.chainId} />
+      // }
+      // Phase this out later, as it applies at the same time
+      // else message = <ICakeTips ifoId={ifo.id} ifoChainId={ifo.chainId} ifoAddress={ifo.address} />
```
ChefBingbong, line 236: can we remove these comments or may they be needed in future?

packages/ifos/src/constants/ifos/arb.ts
```diff
@@ -1,3 +1,125 @@
...
+      additionalClaimingFee: false,
+    },
+    tokenOfferingPrice: 0.1,
+    vestingTitle: 'testCrossChainIfo1',
+  },
+  {
+    id: 'testCrossChainIfo0',
+    version: 8,
+
+    address: '0x6164B999597a6F30DA9aEF8A7F31D6dD7AE57e04', // Test IFO
```
ChefBingbong, line 105: is this test config going to be used often and wil the addresses of the test ifo contracts change much? if so would it be better to make a maping for these test contract addresses for each chain and use the address here, that way its easier to keep track of the contracts and change them if nessecary

packages/widgets-internal/ifo/constants.ts
```diff
@@ -0,0 +1,11 @@
+import { ChainId } from "@pancakeswap/chains";
+
+// TODO: Consider moving to IFO package
+export const ChainNameMap = {
+  [ChainId.BSC]: "BSC",
```
ChefBingbong, line 5: i think we already have a similar mapping exported from the chains package.

ChefBingbong: if not lets actually add one

apps/web/src/components/CrossChainVeCakeModal/hooks/useTxnByChain.ts
```diff
@@ -0,0 +1,13 @@
+import { ChainId } from '@pancakeswap/chains'
+import { IfoChainId } from '@pancakeswap/widgets-internal/ifo/constants'
+import { atom, useAtom } from 'jotai'
+
+const txnByChainAtom = atom<Record<Exclude<IfoChainId, ChainId.BSC>, string>>({
+  [ChainId.ARBITRUM_ONE]: '',
+  [ChainId.ETHEREUM]: '',
+  [ChainId.ZKSYNC]: '',
```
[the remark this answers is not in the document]
thechefpenguin: Yes true 👍

apps/web/src/views/Ifos/components/IfoFoldableCard/IfoPoolCard/ClaimButton.tsx
thechefpenguin: Yeah that'll be a cleaner way to do it

apps/web/src/views/Ifos/components/IfoFoldableCard/IfoPoolCard/ContributeModal.tsx
thechefpenguin: Yeah we can consider a default map but based on chain as well. 

Right now while calling we're relying on `callWithGasPrice` which should estimate gas with some buffer already. perhaps when this was coded initially this wasn't there 🤔

apps/web/src/views/Ifos/components/IfoFoldableCard/IfoPoolCard/IfoCardTokens.tsx
thechefpenguin: I would prefer to keep it for now, as might need that line of code for BNB chain IFO in future just in case

packages/ifos/src/constants/ifos/arb.ts
thechefpenguin: that's a good idea. we can consider next time, as right now not many more test IFOs

packages/widgets-internal/ifo/constants.ts
thechefpenguin: Yeah I was wondering why this didn't exist. Right now a bit tight on timeline so maybe this can be a separate PR? @ChefBingbong

apps/web/src/views/Ifos/components/IfoFoldableCard/IfoPoolCard/ContributeModal.tsx
ChefBingbong: ahh ok i see. thanks for the clarity

packages/ifos/src/constants/ifos/arb.ts
ChefBingbong: yeah i was kinda thinking that. just wanted to comment just incase. cool

packages/widgets-internal/ifo/constants.ts
ChefBingbong: yeah of course we can. do this another time for sure

packages/ifos/src/constants/ifos/arb.ts
ChefBingbong: yeah sure. i can take a note a raise pr in future

apps/web/src/views/Ifos/components/IfoFoldableCard/IfoPoolCard/IfoCardTokens.tsx
ChefBingbong: yeah i was kinda thinking that. just wanted to comment just incase. cool

ChefMomota (approved):

PR #10730, opened 2024-09-24
feat: Gauges with API

ChefMomota:
---

## PR-Codex overview
This PR focuses on refactoring the `gauges` and `smart-router` packages, particularly enhancing API calls to fetch gauge data and improving the handling of asynchronous operations. It also introduces new functions for retrieving gauges by chain.

### Detailed summary
- Deleted `prod.ts` and `index.ts` files from `packages/gauges/src/constants/config`.
- Added `GAUGES_API` constant in `endpoint.ts` for API endpoint.
- Exported `getGaugesByChain` and `safeGetGaugesByChain` functions in `getGaugesByChain.ts`.
- Updated `getAllGauges.ts` to use `getGauges` instead of `CONFIG_PROD`.
- Modified various functions to await asynchronous operations for fetching data.
- Refactored `getPairCombinations` to support async behavior.
- Updated tests to reflect changes in gauge configuration fetching.


packages/smart-router/evm/v3-router/functions/getPairCombinations.ts
```diff
@@ -29,21 +27,15 @@ const getToken = memoize(
...
   return Boolean(token && BASES_TO_CHECK_TRADES_AGAINST[token.chainId as ChainId]?.find((t) => t.equals(token)))
 }
 
 const getTokenBasesFromGauges = memoize(
-  (currency?: Currency): Token[] => {
+  async (currency?: Currency): Promise<Token[]> => {
     const chainId: ChainId | undefined = currency?.chainId
     const address = currency?.wrapped.address
-    const gauges = getGaugesByChain(currency?.chainId)
+    const gauges = await getGaugesByChain(chainId)
```
chefjackson, line 38: This is something we can cache and don't need to fetch every time.

Also we might need some retry strategy to make sure all data is fetched. Meanwhile we also need to make sure that even if the fetch failed, rest logic still works as it's not the main flow

chefjackson: we need to make sure:
1. Even if we failed to get gauges, we can still proceed with swap
2. We can't let gauge config become the bottleneck which slows down the whole swap. So it's better to add a timeout

packages/gauges/src/constants/config/prod.ts
```diff
@@ -1,5404 +1,54 @@
...
-    token0Address: baseTokens.weth.address,
-    token1Address: baseTokens.eurc.address,
-    feeTier: FeeAmount.LOWEST,
-  },
-]
+import retry from 'async-retry'
+import { GaugeConfig } from '../../types'
+
+let cache: GaugeConfig[] | undefined
+let promiseQueue: ((value: GaugeConfig[] | PromiseLike<GaugeConfig[]>) => void)[] = []
```
chefjackson, line 5: we don't really need a queue right? We can reuse the ongoing promise if there's one.

packages/gauges/src/constants/config/prod.ts
```diff
@@ -1,5404 +1,54 @@
...
+  if (inProgress) {
+    return new Promise((resolve) => promiseQueue.push(resolve))
+  }
+
+  inProgress = true
+
+  return new Promise((resolve) => {
+    retry(
+      async () => {
+        const response = await fetch('https://cms-public-api-7ys4p.ondigitalocean.app/api/data/cached/gauges', {
```
chefjackson, line 20: Let's move the endpoint to config

packages/gauges/src/constants/config/prod.ts
```diff
@@ -1,5404 +1,54 @@
...
-  },
-]
+import retry from 'async-retry'
+import { GaugeConfig } from '../../types'
+
+let cache: GaugeConfig[] | undefined
+let promiseQueue: ((value: GaugeConfig[] | PromiseLike<GaugeConfig[]>) => void)[] = []
+let inProgress = false
+
+export const CONFIG_PROD = async (): Promise<GaugeConfig[]> => {
```
chefjackson, line 8: Can we rename and move this function somewhere else? It doesn't make sense to call it `CONFIG_PROD`. Maybe `queries/getGauges.ts`

packages/gauges/src/constants/config/getGauges.ts
```diff
@@ -0,0 +1,55 @@
...
+        retries: 3,
+        minTimeout: 0,
+        onRetry: (error: Error, attempt: number) => {
+          console.warn(`Attempt ${attempt} failed: ${error.message}. Retrying...`)
+        },
+      },
+    ).catch((error: Error) => {
+      console.error('Failed to fetch gauges config after retries: ', error)
+
+      promiseQueue.forEach((queuedResolve) => queuedResolve([]))
```
chefjackson, line 48: I think it's better to throw and catch wherever we need to use this function. Because it's easier for us to do cache control. For example it would be impossible to do cache on `getGaugesByChain` right now. The gauge list is kind of large so we better apply cache to `getGaugesByChain` as well

ChefMomota: updated

packages/gauges/src/constants/config/getGauges.ts
```diff
@@ -3,53 +3,41 @@ import { GaugeConfig } from '../../types'
...
-      promiseQueue = []
-      inProgress = false
-
-      resolve([])
-    })
+    },
+  ).catch((error: Error) => {
+    console.error('Failed to fetch gauges config after retries: ', error)
+    ongoingRequest = null // Clear the ongoing request on failure
+    throw error
```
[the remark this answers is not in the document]
ChefMomota: done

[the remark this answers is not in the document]
chefjackson: we can throw here and catch the error where we use this function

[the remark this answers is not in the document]
chefjackson: the downside is we can't do caching for `getGaugesByChain` if we don't throw. Cuz the gauge list is quite large so better add memory cache to `getGaugesByChain` as well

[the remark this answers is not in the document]
chefjackson: wait does this new getflatgauges respect the gid order?

packages/gauges/src/constants/config/endpoint.ts
```diff
@@ -0,0 +1 @@
+export const GAUGES_API = 'https://cms-public-api-7ys4p.ondigitalocean.app/api/data/cached/gauges'
```
chefjackson, line 1: Please replace with the prod domain

thechefpenguin (approved):

chefjackson (approved):

PR #10754, opened 2024-09-28
fix: Add undefined check for trade order

memoyil:
---

## PR-Codex overview
This PR focuses on improving the handling of optional values in the `useAllTypeBestTrade` hook and the `V3SwapForm` component. It ensures that properties are accessed safely, preventing potential runtime errors when values are undefined.

### Detailed summary
- Updated `classicAmmOrder` to return `undefined` if `ammCurrentTrade` is falsy.
- Changed `bestOrder` type to allow `undefined`.
- Modified `validTrade` assignment to safely access `ammOrder.trade`.
- Updated `ammInputAmount` and `ammOutputAmount` to safely access values from `ammOrder`.
- Adjusted `ammOrder` structure to safely access `type`.

### DIFF
--- a/apps/web/src/views/Swap/V3Swap/hooks/useAllTypeBestTrade.ts
+++ b/apps/web/src/views/Swap/V3Swap/hooks/useAllTypeBestTrade.ts
@@ -61,12 +61,14 @@ export const useAllTypeBestTrade = () => {
   const refreshOrder = useThrottleFn(bestOrder.refresh, 3000)
 
   const classicAmmOrder = useMemo(() => {
-    return {
-      trade: ammCurrentTrade,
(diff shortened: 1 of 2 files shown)


packages/price-api-sdk/src/types/order.ts
chefjackson, line 1: Let's make sure on sdk typing end we don't allow undefined trade

chefjackson: it doesn't make sense if we have a order with type but no trade. Let's make sure if we initialize a order it has to have a valid trade or we don't initialize the order at all

chefjackson (approved):

PR #10780, opened 2024-10-03
feat: Enable zap for all bsc pools and add arb and eth whitelist

memoyil:
---

## PR-Codex overview
This PR focuses on updating the `ZapLiquidityWidget` component to change the source from `zap-widget` to `pancakeswap`, improving the logic for checking if a pool is in the `ZAP_V3_POOL_ADDRESSES`, and restructuring the pool addresses for different chains.

### Detailed summary
- Changed `source` prop in `ZapLiquidityWidget` from `zap-widget` to `pancakeswap`.
- Updated pool address validation logic in `V3FormView` to use chain-specific whitelists.
- Refactored `ZAP_V3_POOL_ADDRESSES` to support multiple chains with specific addresses.
- Added new pool addresses for `ChainId.ETHEREUM` and `ChainId.ARBITRUM_ONE`.


apps/web/src/config/constants/zapV3.ts
```diff
@@ -1,22 +1,56 @@
...
-  '0x1E4600929Edf7F36B4A7eAc4C7571057D82a246c', // SOL-WBNB 0.25%
-  '0xBe141893E4c6AD9272e8C04BAB7E6a10604501a5', // ETH-USDT 0.05%
-  '0xd881D9d0e0767719701305c614903F555d589586', // TUSD-USDT 0.01%
-  '0xf2688Fb5B81049DFB7703aDa5e770543770612C4', // USDC-WBNB 0.01%
-  '0xd15B00E81F98A7DB25f1dC1BA6E983a4316c4CaC', // XRP-WBNB 0.25%
-]
+import { ChainId } from '@pancakeswap/chains'
+
+export const ZAP_V3_POOL_ADDRESSES: Record<number, string[]> = {
+  [ChainId.BSC]: [] as string[], // Empty array to whitelist all pools
```
chefjackson, line 4: It's a bit counter intuitive. Can we use undefined check instead? Like if not specified then no whitelist restriction

chefjackson: ah can you add a comment here? It would be super hard for other devs to understand without additional comment.

Or it would be better if we can have a separate constant SUPPORTED_CHAINS

chefjackson (approved):

PR #10899, opened 2024-10-31
perf: Use query instead of queries to reduce rerenders initially

memoyil:
---

## PR-Codex overview
This PR focuses on refactoring multiple hooks in the codebase to replace `useQueries` with `useQuery`, simplifying the data fetching logic and improving error handling across various components.

### Detailed summary
- Replaced `useQueries` with `useQuery` in several hooks.
- Streamlined query functions to use `Promise.all` for concurrent fetching.
- Enhanced error handling with `console.error` in query functions.
- Updated query keys to utilize `join` for better string formatting.
- Simplified memoization of results using `useMemo`.

### DIFF
--- a/apps/web/src/hooks/v3/usePools.ts
+++ b/apps/web/src/hooks/v3/usePools.ts
@@ -215,17 +215,23 @@ export function usePoolsWithMultiChains(
             functionName: 'liquidity',
           }))
 
-          const [slot0Data, liquidityData] = await Promise.all([
-            client.multicall({
-              contracts: slot0Calls,
-              allowFailure: false,
-            }),
-            client.multicall({
-              contracts: liquidityCalls,
-              allowFailure: false,
-            }),
-          ])
-          return { [chainId]: { slot0: slot0Data, liquidity: liquidityData } }
+          try {
+            const [slot0Data, liquidityData] = await Promise.all([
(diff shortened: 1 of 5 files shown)


apps/web/src/hooks/v3/usePools.ts
```diff
@@ -190,58 +190,63 @@ export function usePoolsWithMultiChains(
...
-        ])
-      },
-      enabled: !!chainId && !!poolAddresses.length,
-      ...QUERY_SETTINGS_IMMUTABLE,
-      ...QUERY_SETTINGS_WITHOUT_INTERVAL_REFETCH,
-    }
-  })
-  const poolInfoQueries = useQueries({
-    queries,
+  const { data: poolInfoByChainId } = useQuery({
```
chef-eric, line 198: the problem is useQueries will cause re-render when every query update?

chef-eric (approved):

apps/web/src/hooks/v3/usePools.ts
chef-eric: I think useQueries can control of cache more precise, for example, if there is one poolAddress update, queries can just fetch only one chain's data, but if use single query, all chainIds need to refetch again. so it's a tradeoff

PR #11021, opened 2024-12-07
fix: will switch back to bsc after switch chain

vvvvvv1vvvvvv:
I found you will call `wallet_addEthereumChain` with invalid params for test if wallet support `wallet_addEthereumChain`
But for some multi-chain wallets like Rabby, when you call `wallet_addEthereumChain` with chainId, wallet will also switch to the target chain even you didn't set other params.
So I set up a PR, you can set params as `[]` which is invalid for all wallets and have the same effect as now


Chef-Yogi:
@vvvvvv1vvvvvv 
thanks for sharing the issue, and try to fix it
Our community chef has also opened a PR. We are now reviewing it: 
https://github.com/pancakeswap/pancake-frontend/pull/11016

PR #11394, opened 2025-03-24
feat: ido-4

ChefJerry:


apps/web/src/views/Idos/hooks/ido/useIDOContract.ts
```diff
@@ -18,6 +18,8 @@ export const useIDOContract = () => {
 }
 
 function getIdoAddressFromUrl(): `0x${string}` | null {
+  if (!(window?.location?.hostname === 'localhost' || process.env.NEXT_PUBLIC_VERCEL_ENV === 'preview')) return null
```
chefphilip, line 21: simplify by using if production, return null?

apps/web/src/views/Idos/components/IdoCards/ClaimedCard.tsx
```diff
@@ -27,6 +28,7 @@ export const ClaimedCard: React.FC<{
   const { theme, isDark } = useTheme()
   const claimed = userStatus?.claimed
   const chainId = useChainId()
+  const { address: account } = useAccount()
```
chefphilip, line 31: it seems variable is not used in the file, right?

thechefpenguin (approved):

PR #11452, opened 2025-04-08
feat: Support cross-chain swap (Bridge Only)

chefphilip:


apps/web/src/quoter/atom/bestCrossChainAtom.ts
```diff
@@ -0,0 +1,544 @@
...
+            ..._option,
+            currency: bridgeOriginCurrency,
+            // NOTE: use suffix to avoid hash collision
+            hash: _option.hash ? `${_option.hash}-swap-bridge` : '',
+          }
+
+          // Get the swap quote from base currency to bridge origin currency
+          const swapOrder = await get(bestQuoteAtom(swapOption))
+
+          if (swapOrder.data) {
```
chef-ryan, line 274: This Loadable value bring complexity when you build logic on it. 
I have a refactor in this PR to make it a monad type. 

you can review this PR for me: https://github.com/pancakeswap/pancake-frontend/pull/11632

Make it simple i want you chan do a `map` function on swapOrder, so you don't need to check it's `pending` `error` `no-value` states. 
```
return swapOrder.map(order => {
  // logic focus only on success value here 
}) // This returns another Loadable based on the map Function 
```

PR #11778, opened 2025-06-11
chore: add missing logic and test for customStatus 

chefphilip:


apps/web/src/views/Swap/Bridge/utils/__tests__/customBridgeStatus.test.ts
```diff
@@ -0,0 +1,277 @@
...
+            status: { code: BridgeStatus.PENDING },
+          },
+          {
+            command: Command.SWAP,
+            status: { code: BridgeStatus.SUCCESS },
+          },
+        ],
+      }
+      const result = customBridgeStatus(bridgeStatus)
+      expect(result).toBe(BridgeStatus.SUCCESS)
```
thechefpenguin, line 140: expect PENDING status

chefphilip: as discussed with BE, this one will be SUCCESS

